PR #8, opened 2019-10-06
Use sortedmulti() descriptor for BIP67

Sjors:
Depends on Bitcoin Core master (https://github.com/bitcoin/bitcoin/pull/17056)

ColdCard simulator no long complains about BIP67 violation (at least for the one change address that it complained about before).

### DIFF
--- a/specter.py
+++ b/specter.py
@@ -401,8 +401,8 @@ class WalletManager:
                 descs[i] = "[%s%s]%s" % (key["fingerprint"], key["derivation"][1:], key["xpub"])
         recv_descs = ["%s/0/*" % desc for desc in descs]
         change_descs = ["%s/1/*" % desc for desc in descs]
-        recv_desc = "multi({},{})".format(sigs_required, ",".join(recv_descs))
-        change_desc = "multi({},{})".format(sigs_required, ",".join(change_descs))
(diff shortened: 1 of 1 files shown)


stepansnigirev:
What worries me is that `sortedmulti` will break compatibility with current and next release of Bitcoin Core. If I understand correctly 0.19 is in RC already and `sortedmulti` is scheduled for 0.20 release.
I see two ways to fix it at the moment:
- Generate addresses in python, add them to Bitcoin Core one by one if Core version doesn't support `sortedmulti`.
- Ignore current Core release, merge this pull request and write in the `Readme.md` that we only support master branch of Core at the moment.
What way should we take?

Sjors:
`sortedmulti` is indeed scheduled for 0.20 and it's probably too late to add it to 0.19.

Given that this project is still experimental, I think the best approach is to use Bitcoin Core master, and then switch to 0.20 early next year. Hopefully native descriptor wallets also make it into that release, so you can just use `getnewaddress`.

stepansnigirev:
Fixed with https://github.com/cryptoadvance/specter-desktop/pull/22
I made a temporary workaround to import sortedmulti addresses in current BitcoinCore release. It sorts the xpubs for every index.

PR #13, opened 2019-10-12
Autodetect fix

### LINKED ISSUE #12: Manual RPC settings always defeated by autodetect
Start specter-desktop on a machine that does not have a `bitcoin.conf` file (or is in a non-standard location). Also make sure there is no ~/.spector/config.json.

The default `Specter.config` sets `autodetect: True`. But of course autodetect will fail because it can't find `bitcoin.conf`.

However when manual RPC settings are entered via the UI neither "Test" nor "Save" will be able to connect because both continue to carry the `autodetect: True` setting. The relevant `else` code in `get_cli` in specter.py will never be run.

kdmukai:
Fixes #12


stepansnigirev:
I think it got merged with https://github.com/cryptoadvance/specter-desktop/pull/23

PR #23, opened 2019-10-31
HWI support

kdmukai:
## Installation
Note the README.md question about installing libusb separately from just installing via the pip dependencies.

## Breaking changes
* Wallets will need to be manually updated to include the `device_type` attr for simple wallets and the new objs in `devices` for multisig.
* Wallet.is_multisig is changed to a property

## Known issues
* "Add more keys" to device does not bring up the correct HWI UI options.
* The server.py `_enumerate` call relies on an os call out to the hwi-cli which necessitates an earlier call to `_locate_hwi`. None of this is ideal but I could not get around Flask crashing when `enumerate` is called when no devices are plugged in. The first call would usually work but then a subsequent check would crash Flask.

## Misc
JS and new UI elements have only been tested in Chrome/macOS.


stepansnigirev:
Tested on Trezor T, Trezor One and Ledger. Works!

kdmukai:
Did you have to do the separate libusb install step or was pip installing the requirements sufficient?

stepansnigirev:
I had libusb in the system already, so I don't know :)
I will check again on a fresh computer.

stepansnigirev:
Yes, you need to install `libusb`. Otherwise `hidapi` doesn't work.

kdmukai:
Got it. Then I'll update the docs.

PR #31, opened 2019-11-22
SLIP-0132 compliant xpub extraction from HWI hardware wallets

kdmukai:
Previous implementation extracted xpubs from USB hardware wallets via HWI but regardless of derivation path the xpub was always prefixed with "xpub" for mainnet and "tpub" for testnet.

Now converting those xpubs to the expected ypub, zpub, Ypub, Zpub for mainnet and upub, vpub, Upub, Vpub for testnet as specified in SLIP-0132.
see: https://github.com/satoshilabs/slips/blob/master/slip-0132.md

Verified that the Zpubs and ypubs now match what Electrum extracts from the same hardware wallet.


stepansnigirev:
Great, thanks! I am not a big fan of xyzYZ-pubs, but as they are commonly used in the industry let's keep them in this form :)

kdmukai:
Yeah, my goal was to make sure we matched Electrum. That way even if this desktop UI is considered risky and experimental, if it's completely interoperable/interchangeable with Electrum, then there's significantly less risk.

PR #33, opened 2019-11-22
small Docu-bug an errormanagement-improvement

k9ert:
* Readme was wrong after moving server.py
* We have the details of the parsing issues, let's reveal them to the user!


kdmukai:
utACK (untested - I don't have a Coldcard yet). But changes look reasonable and more pythonic.

Caught the same README change in #32.

Issue #37, opened 2019-11-24
Standalone HWI-enabled PSBT signing mini app


kdmukai:
In conjunction with #7 I'd like to build a standalone companion app that would leverage the existing codebase to allow for remote multisig partners to sign PSBTs with usb hardware wallets.

We can already support Coldcard downloaded PSBTs and Specter QR code PSBT signing with remote multisig partners by sharing the webserver's TOR onion url with them (though with no current security model it's probably not ideal to give all signers full access to the Specter Desktop UI). But usb wallet support requires a local install, rendering remote co-signers out of luck.

Projects like [flaskwebgui](https://github.com/ClimenteA/flaskwebgui) and [PyFladesk](https://github.com/smoqadam/PyFladesk) look like pretty promising ways to package a Flask site as a standalone app that can then be packaged and installed on any OS via PyInstaller.

It should be straightforward then to re-use the existing Specter HWI UI flow (detect-unlock-sign) in a separate standalone app that is just intended for remote PSBT signers. In its simplest form the PSBT can be transmitted to the signer through some secure means and then pasted/uploaded into the standalone signing app. The signed PSBT can then be downloaded and transmitted back. This could all be accomplished with any PSBT-compliant signing app, of course. But I haven't come across any that are strictly light clients (i.e. co-signers shouldn't need a full node, Electrum, etc just to sign a PSBT).

A slightly more advanced version could take advantage of our Tor layer and have the standalone app accept a Specter Desktop onion address so it can retrieve the unsigned PSBT from the server, have the user sign it, and then automatically send it back for the Specter Desktop to combine and transmit when complete.

kdmukai:
Some early progress, some early frustrations in my dev branch:
https://github.com/kdmukai/specter-desktop/tree/psbt_signing_app/src/signer

The good news is that `flaskwebgui` is successfully launching my mini app in its own standalone browser window.

Struggling with python imports across the two apps. Using two ugly workarounds for now:
* soft link added in the `signer` app to point to the `specter` app in order to do the import to add the HWI blueprint.
* `signer` templates are currently under specter/templates/signer. Had a more isolated version working but that also relied on more soft links.

k9ert:
Cool idea! However, i'd like to definitely avoid soft-link magic like that. I'd probably prefer to have the mini-app rather more close coupled to avoid symlinks but they definitely should have different blueprints. So speaking within the current setup, we would have a server.py and a server-miniapp.py and they use different blueprints.

[...]

PR #38, opened 2019-11-24
dir reorg for app-like structure

kdmukai:
Defines a 'specter' app dir which holds all of the python files, 'static', and 'templates'. This structure is more typical for Flask and Django projects.

See: https://exploreflask.com/en/latest/organizing.html#package

This also paves the way for additional apps like #37 that would be conceptually separate but would still want to leverage parts of the 'specter' app.


k9ert:
I like that change! Maybe i should have done that already in my last commit.
However, the tests are not working. In order to work you now need imports like that:
`from specter.specter import Specter, alias ` (test/test_specter.py:4)

kdmukai:
@k9ert I'll investigate. I thought I had already confirmed the tests but maybe made further changes without looping back there.

PR #50, opened 2019-12-19
Factory pattern

k9ert:
Over the mid-term, we want to use the factory-pattern which is moving the Flask-app-generation-code into a function. This has at least the following advantages:
* testability
* modularisation (not so much global-code/-variables)
* Easier interfaces e.g. to VS-Code-flask-debugging-launch-configuration

The Pull-request is work in progress. Please don't merge yet but i want a clear path with several clear commits because otherwise all the changes as one commit would be too much.
That also means to please not squash-the-commits when merging.


k9ert:
Meeh, i mixed up the commits. I'll sort that out better and will close this PR for now.

Issue #48, opened 2019-12-12
As a Specter Dev i want to have a bitcoind at my fingertips


k9ert:
A working bitcoind (probably in regtest-mode) is essential for specter-development. However, not everyone is that familiar with getting bitcoind to run. Currently we simply assume that it needs to be there but don't give any hints on how to get it in place.
So let's change that with the help of docker.

We're currently already using docker for testing within gitlab. That image can be run like this:
`
docker run -d --publish 18544:18544 --expose 18544 --publish 18543:18543 --expose 18543 registry.gitlab.com/cryptoadvance/specter-desktop/python-bitcoind:latest `

This is far to complicated to my taste. We haven't dicussed mining yet which would make something like "docker exec ..." necessary. Also usually as a developer one wants to develop the solution and run the tests at the same time in parallel. For this to work one wants to use different ports.

For that reason, i'm planning to make that easier via a command line interface with a nice and short handle. How about "btctrl"

Maybe something like:
`
python3 btctrl --start --mine
`
This would start the bitcoind and mine a block every 30 seconds. Maybe it could also configure specter-desktop automatically so that you don't have to do that manually.

Thoughts?

k9ert:
This is already implemented. Closing.

PR #51, opened 2019-12-19
Cli

k9ert:
followup for #48 and #49 
Cli does mine now for the relevant addresses in ~/.specter/wallets/regtest
By default every 15 seconds one block.

### DIFF
--- a/src/specter/bitcoind.py
+++ b/src/specter/bitcoind.py
@@ -15,11 +15,37 @@ import docker
 from helpers import which
 
 
-@click.command()
-def main():
+@click.group()
+def cli():
+    pass
+
+@cli.command()
+@click.option('--debug/--no-debug', default=False)
+@click.option('--mining/--no-mining', default=True)
+def bitcoind(debug,mining):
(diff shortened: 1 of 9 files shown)


stepansnigirev:
Should we mine by default?
I am not sure about automatic changes in bitcoin node... Wouldn't it be better to pass `--mining` explicitly to enable mining? ... I would prefer to have `--no-mining` by default.

What I am worried about is that if you use a remote node and multiple specters on different machines, and keep them running in the background, you will end up with a huge amount of empty blocks after a while.

k9ert:
I think we should mine by default because empty blocks don't eat up much space and also they don't mess with the ~/.bitcoin/regtest directory of the user but with that directory inside the docker-instance.

If you REALLY use that with multiple machines, i'm not sure that's even possible as the ports need to be available outside the host-machine. It probably doesn't make that much sense if you don't use it on your dev-machine.

stepansnigirev:
Ok, I agree. For the docker container it's fine to mine :)

PR #54, opened 2019-12-20
Adds optional fee rate

kdmukai:
Advanced options like this can eventually be made visible or hidden via a settings menu. But for now default fee calcs were way higher than my time preference required (e.g. 7k sats vs 500).


stepansnigirev:
Looks good, thanks! Even Blockstream explorer was also notifying me that the fee is too large in the transaction. Good that we can change it now.
It would be good to have an estimate of when the tx will be included in the block for different fee rates, but I suppose there is no bitcoin-cli call like that so let's leave it for later.

PR #57, opened 2019-12-28
Factory pattern

k9ert:
We want to use the factory-pattern which is moving the Flask-app-generation-code into a function. This has at least the following advantages:

* testability
* modularisation (not so much global-code/-variables)
* Easier interfaces e.g. to VS-Code-flask-debugging-launch-configuration


kdmukai:
Wondering if the .vscode/launch.json belongs in the repo. The path setting seems potentially system dependent and, in general, I'm uncomfortable with IDE-specific files in the repo unless they're a hard requirement. I wouldn't want to imply or create any dependencies on a specific IDE.

Seems like it might make more sense as an addendum to the README. Though even then it's a judgment call if the config is something that someone accustomed to VSCode can easily build themselves (I've never used VSCode so I can't say).

_Full disclosure: Visual C++ absolutely tortured me early in my coding career. I've heard good things about VSCode but the extreme negative associations of that name persist for me even though I'm sure it's 1000x better!_

k9ert:
I can understand all of your points you're making here. I'm especially agreeing to the one with the system-dependent one. Ideally this should work out of the box but unfortunately it's not (without adding manually the path in that json) and i created a stackoverflow question about this here, yesterday:
https://stackoverflow.com/questions/59508826/pythonenv-specific-path-not-respected-when-debugging-flask-with-my-launch-json

I tend to value developer-ease over cleaniness of the repo especially if the files are not hurting the people not using that IDE. In that case, it's imho neither a runtime-dependency nor a build-/testing-dependency but an optional "development dependency". 

However i agree to the point that the launch.json isn't very portable and probably wouldn't work on ... let's say a Apple Mac (i'm using Ubuntu).

So in order to not overweight the README.md, i'll create a DEVELOPMENT.md where i describe all things specific to make development easy and will remove the launch.json.

_I felt tortured very similar about Microsoft and Visual Studio when i needed to maintain VB.Net code at some point in the very past. I would have called everyone a fool if they would have told me that i will ever use anything from Microsoft on a voluntary basis in the future. 
I can only encourage to try it, especially because of the very rich plugin-system. It's awesome indeed! _

stepansnigirev:
*I also heard good things about VSCode, it has a very nice plugin architecture. 
There is even a debug config for Specter-DIY to debug connected board:
https://github.com/diybitcoinhardware/f469-disco/blob/master/debug/vscode.md*

PR #62, opened 2020-01-10
Pyinstaller

k9ert:
Still work in progress. The first commit makes it running again on linux. One flaw was that the name of the app could not be set. See 
https://github.com/k9ert/specter-desktop/blob/pyinstaller/docs/continuous-integration.md#releasing
for details on how to build the artifact.

Please don't merge yet. Next step is windows.

### DIFF
--- a/src/specter/views/hwi.py
+++ b/src/specter/views/hwi.py
@@ -5,7 +5,7 @@ import subprocess
 import sys
 
 from flask import Flask, Blueprint, render_template, request, redirect, jsonify, current_app
-from helpers import normalize_xpubs, convert_xpub_prefix
+from helpers import normalize_xpubs, convert_xpub_prefix, which
 from hwilib import commands as hwilib_commands
 from hwilib import base58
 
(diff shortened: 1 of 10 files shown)


src/specter/views/hwi.py
```diff
@@ -18,22 +18,7 @@
...
-        returned_output = subprocess.check_output(["hwi", "--version"], cwd=virtualenv_bin_dir)
-        if "hwi" not in returned_output.decode("utf-8").lower():
-            print("Could not find 'hwi' executable for HWI hardware wallet support")
-            exit(0)
-        else:
-            HWI_DIR = virtualenv_bin_dir
-
-_locate_hwi()
-
+HWI_EXEC = which("hwi")
```
k9ert: @kdmukai i did some simplifications here. Hope that works for you as well.

kdmukai: Cool. My original workaround was so hack-y and I hated that it was necessary. I don't have the cycles to help with testing at the moment but definitely Concept ACK.

k9ert:
As my focus is going somewhere else and i won't be able for now to continue on the windows-part of this, Let's close this PR, please!

PR #63, opened 2020-01-16
Some minor things to consolidate and some guidelines for Development.md

k9ert:
* Development guidelines and hints
* Adding additional derivation-pathes to conftest in order to use them for future tests (Better seperate tests with different wallets if possible)
* Added a SpecterError class deriving from Exception to be used in proper Errormanagement (raise errors ourself which should be passed to the users)
* some minor changes


k9ert:
@kdmukai maybe you can have a look as well, especially in the changes for Development.md. Would be cool if you can commit to that as well.

kdmukai:
> @kdmukai maybe you can have a look as well, especially in the changes for Development.md. Would be cool if you can commit to that as well.

Looks good. Some minor thoughts:
* Maybe add a note about trying to roughly follow python PEP-8 but no need to be super strict.
* Vue.js is pretty cool. Was super-useful on a project that had complex nested objects 3 or 4 levels deep all on the same page. But I didn't see the point for simple objects if they didn't have much/any reuse elsewhere.
* I think any cross-browser issues are less of a big deal than getting a mobile-friendly UI.

stepansnigirev:
I was thinking that in some cases DOM updates with getElementByID().innerHTML is pretty annoying.
In these cases, vuejs would help as it can construct bindings between javascript values and DOM elements.
For example, this pull request https://github.com/cryptoadvance/specter-desktop/pull/65 could make use of vue bindings and significantly simplify the javascript part.
I still want to avoid global app-style frameworks so I would suggest using vue only for small self-contained pieces of code. I am strongly against npm in the project and a huge list of javascript dependencies. Vue allows to import it with a single file, that's a good thing about it.

If we can create vue components that can be self-contained and work without npm (ideally imported with a few lines together with vuejs) then I am all for it. Unfortunately, I am not an expert in vue or javascript, so I need an external opinion here.

@kdmukai I agree, mobile-friendly is very critical. Specter looks terrible even in a slightly reduced browser window...

stepansnigirev:
@k9ert should I merge? Looks good to me.

k9ert:
I think it's worth to also checkin a reasonable vue.js binary. So:
* Releases are tracked here: https://github.com/vuejs/vue/releases
* One month ago 2.6.11 has been released which looks like the version-numbers we're looking for (high patch number)
* this binary contains vue.js and vue.min.js which is what we're looking for:
https://github.com/vuejs/vue/archive/v2.6.11.zip
* it has this hash: 

```
➜  Downloads sha256sum vue-2.6.11.zip 
2166c0b6d3724bb9ab2aa9444849657eacae0ec5b7c12257ef83a574ec0db256  vue-2.6.11.zip
➜  Downloads unzip vue-2.6.11.zip 
Archive:  vue-2.6.11.zip
ec78fc8b6d03e59da669be1adf4b4b5abf670a34
   creating: vue-2.6.11/
  inflating: vue-2.6.11/.babelrc.js 
[...]
  inflating: vue-2.6.11/types/vue.d.ts  
  inflating: vue-2.6.11/yarn.lock    
➜  Downloads cd vue-2.6.11/dist 
➜  dist sha256sum vue.js 
352baa818da109925437a8433057ddc6f91ec48efe88bc5741b2f9e34450fdce  vue.js
➜  dist sha256sum vue.min.js
9e0156dd49c03744e79bbea60eebbbb94b5811c1b71b91f5fb38a8270dedfbaf  vue.min.js
➜  dist 
```
So i'd suggest that i, before we close this PR:
* I add these two binaries to the static-folder 
* Also i'd add this link to the dev-guidelines. https://vuejs.org/v2/guide/security.html

Agree?

stepansnigirev:
Agree! I think we can only add vui.min.js and even put it in the header of base.html such that it is always available.

requirements.txt
```diff
@@ -24,6 +24,7 @@ typing-extensions==3.7.4
 urllib3==1.25.6
 Werkzeug==0.16.0
 serial==0.0.97
```
stepansnigirev: not sure, should be only `pyserial`

k9ert:
merged!

requirements.txt
k9ert: Ups, sorry, missed your comment. I'll fix it with the next change!

PR #65, opened 2020-01-21
add dynamic fees option

### LINKED ISSUE #55: Get tx fee estimates based on bitcoin core mempool
Picking up the conversation in #54 it seems like we should be able to query bitcoin core to characterize the current fee distribution in the mempool (e.g. btc-rpc-explorer's Mempool Summary page) in order to give some rough guidance about what a reasonable fee vs time tradeoff might be at the moment.

gorazdko:
## Abstract

Close #55. Adding option to choose dynamic fees based on block confirmation time (estimatesmartfee). Open for discussion.

## Status
Ready for review or it can be put on hold. See comments below.

- [x] first draft
- [x] style fixes (python should have spaces, JS tabs?)

### DIFF
--- a/src/specter/controller.py
+++ b/src/specter/controller.py
@@ -270,6 +270,11 @@ def wallet_receive(wallet_alias):
             wallet.getnewaddress()
     return render_template("wallet_receive.html", wallet_alias=wallet_alias, wallet=wallet, specter=app.specter, rand=rand)
 
+@app.route('/get_fee/<blocks>')
+def fees(blocks):
+    res = app.specter.estimatesmartfee(int(blocks))
+    return res
+
 @app.route('/wallets/<wallet_alias>/send/', methods=['GET', 'POST'])
 def wallet_send(wallet_alias):
     app.specter.check()
@@ -289,10 +294,16 @@ def wallet_send(wallet_alias):
             address = request.form['address']
             amount = float(request.form['amount'])
             subtract = bool(request.form.get("subtract", False))
-            if request.form.get('fee_rate'):
-                fee_rate = float(request.form.get('fee_rate'))
+            fee_unit = request.form.get('fee_unit')
+
+            if 'dynamic' in request.form.get('fee_options'):
+                fee_rate = float(request.form.get('fee_rate_dynamic'))
+            else:
+                if request.form.get('fee_rate'):
+                    fee_rate = float(request.form.get('fee_rate'))
+
             # try:
-            psbt = wallet.createpsbt(address, amount, subtract=subtract, fee_rate=fee_rate)
+            psbt = wallet.createpsbt(address, amount, subtract=subtract, fee_rate=fee_rate, fee_unit=fee_unit)
             if psbt is None:
                 err = "Probably you don't have enough funds, or something else..."
             else:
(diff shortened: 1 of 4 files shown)


k9ert:
That's awesome, great work :rocket: :rocket: :rocket: ! 

I just tested it on testnet and i really like it:
image
Is the fee ("Fee: 0.00001 BTC/kB") suppose to change if i slide? I had a look into the code as well. On #63  we're currently discussing the use of Vue.js for simplifying javascript-stuff. Would you be interested in doing that with vue.js? If you've never worked with that, we can help here as i'm currently working on a document which gives a rapid start to working with specter and vue.js (accidentally exactly in that very same html-file).

Regarding testing: We have the test_WalletManager (stupid name, should rename that) which is testing the createpsbt-methos. However, i don't think we can use regtest here as we won't get numbers from regtest. So i would suggest to use mocking here. There is no example yet in the tests on how to use mocks. If you haven't done it yet, this might be helpfull:
https://www.toptal.com/python/an-introduction-to-mocking-in-python

I had issues in the past unsing these "@mock.patch('mymodule.os')" notations and so in another project i used something like
with mock.patch('mymodule.something', return_value=True) as mock_mymodule_something:
    ....
I find that even reasonable as my test-functions sometimes tend to get longer and i might not want to cut them for some reasons.

stepansnigirev:
Awesome! Fee estimation that makes sense!!!
@k9ert, I think bare javascript works fine here, but VueJS could simplify DOM update logic. Do we have an example already? I think it would make sense to start using vue only when we have a usage example and some guideline what we allow to do with vue and what not.
@gorazdko if you want to research vuejs you are welcome to do so, but bare javascript works fine as well! Great work!

gorazdko:
> Is the fee ("Fee: 0.00001 BTC/kB") suppose to change if i slide?

yes and no, it depends on the mempool conditions.

> i don't think we can use regtest here as we won't get numbers from regtest

regtest is ok for filling up the mempool, estimatesmartfee works just fine then

Thank you for the tips regarding testing, I will have to study it more when i find time. Given to the latter and the fact that this code will probably have to be rewritten in a new framework, i would defer writing test here. 

Of course this PR doesnt have to be merged at all, especially if it is conflicting with what you are doing. I can pick it up later after your work is in, and will adapt it accordingly.

stepansnigirev:
I like this PR and I am strongly against having "the framework" that is recommended to everyone.
I think that if the functionality is self-contained and works well we don't need to rewrite it in vuejs or anything else. We can just use it.
I would merge this PR and keep it this way. I like pure javascript :)

@k9ert, please take a look. If you think it's good let's merge :)

k9ert:
In general, i really think we have to invest more in testing and i was quite happy that you planned to write tests. On the one hand, most of that code is javascript and we don't have a test-concept on javascript, but on the other hand especially shared code should get tested.
I'm happy if that gets merged but would also like us all to focus a bit more on testing.

Regarding vuejs ... i definitely don't mind if stuff is implemented in pure javascript. But if anyone would rather prefer to implement it in vuejs ... just wanted to mention that there is no reason to not do it.

stepansnigirev:
I suggest merge it now, and if @gorazdko will find some time he could add a test for it.
@k9ert , ok?

k9ert:
Agreed, merged!

PR #66, opened 2020-01-27
USB support for specter

stepansnigirev:
Specter-DIY now supports USB communication.
Similar to ColdCard it can be either air-gapped or connected.
I implemented a `specter_hwi` module that has all the necessary HWI features, so most of the logic is unchanged.


stepansnigirev:
I was also thinking about having a fork of HWI as a submodule. We could tune it then if necessary for DIY devices like Specter, BitBoy or other that may appear.
Ideas?

k9ert:
I'd suggest to merge that even without @kdmukai  's review in preparation of the Advancing Bitcoin. We'll do extended testing upfront anyway and will probably roll forward with any bugs we'll find.
@kdmukai  please, would be cool if you would review nevertheless?!
@stepansnigirev  Agree?

stepansnigirev:
Sure, let's merge and bugfix later :)

PR #67, opened 2020-01-29
unique path for bitcoin core wallets

stepansnigirev:
If the same Bitcoin Core node is used by multiple people (like on the upcoming workshop) there could be name collisions for new wallets - all specters use the same wallet path `specter/walletname` in Bitcoin Core.

This pull request introduces a unique id for Specter generated when you first start the server. 
This `uid` is stored in the `config.json` file. If `config.json` already exists - empty `uid` is used (for backward compatibility).
Wallet path in Bitcoin Core becomes `specter<uid>/walletname`.

Now we should not have name collisions when using the same node with multiple specters.

Also adds a `generate_icon.svg` for mined coins on regtest.

### DIFF
--- a/src/specter/logic.py
+++ b/src/specter/logic.py
@@ -6,6 +6,7 @@ from descriptor import AddChecksum
 import base64
 from serializations import PSBT
 import helpers
+import random
 
 WALLET_CHUNK = 5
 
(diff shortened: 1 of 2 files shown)


k9ert:
Great! Looks good to me and let's test it soon!

PR #70, opened 2020-02-20
command-line options for server: daemon, ssl certificates, tor and others

stepansnigirev:
Adding one more way to run the server - with `--daemon` flag, such that it releases the terminal and runs in the background.
It saves PID file in the `.specter` folder that can be used to stop daemon with `--stop` command

Unfortunately, daemon and flask debug options don't work together so I am disabling debug mode when running as a daemon. But I think it's fine because debug option only makes sense for development and you probably want to check the output of the server when writing code...

I also added an option to pass ssl certificates so flask could run under https.
The same applies to tor - you provide your tor secret and it starts server over tor.

Here what we can do with this pull request:

```sh
python -m cryptoadvance.specter server --tor=mytorpassword --cert=./cert.pem --key=./key.pem --daemon
```

From my point of view it's way easier to set up nginx proxy-pass, I should document it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ build
 dist
 *.egg-info
 .vscode
+cert.pem
+key.pem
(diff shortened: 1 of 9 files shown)


k9ert:
Tested, works! What might be missing are some additional lines of the shortcomings running over tor: The camera is not working. Any hints on how to get that to work is highly appreaciated.
I tried to modify "permissions.default.camera" from "0" to "1" in "about:config" but that didn't work.

PR #71, opened 2020-02-21
Allow to select all devices on the same page in multisig wallet creation

stepansnigirev:
Before we had to select one device, then click next, and continue until we are done.
Now we can select all of them at once.
Also some small styles and bugfixes.


k9ert:
Tested and works! Nice :-).

PR #75, opened 2020-02-26
Auth

k9ert:
With this PR we get optional authentication.
image

If a user activates it, he needs to authenticate with the RPC-Password before accessing the application.
By default, it's set to "none".


k9ert:
Finally ... checks passed!

PR #81, opened 2020-03-01
Make RPC password input use type password

ben-kaufman:
Make RPC password input use type `password` so that users password is not displayed in clear text.


stepansnigirev:
Thanks!

Issue #86, opened 2020-03-08
Wallet Error


mcsnubbs:
I was able to use the software without issue until my node finished syncing, at which point it returned this browser error

cryptoadvance.specter.rpc.RpcError
cryptoadvance.specter.rpc.RpcError: Server responded with error code 500: {"result":null,"error":{"code":-18,"message":"Requested wallet does not exist or is not loaded"},"id":0}

Opening/closing wallets in CORE does not seem to effect it. Any ideas?

stepansnigirev:
Can you try deleting wallets folder in `~/.specter` folder?
I think somehow Bitcoin Core forgot about one of the wallets and specter can't load it now. I will try to fix this issue tomorrow.

mcsnubbs:
Thank you - ill give it a shot. Tried uninstalling specter/reinstalling but for some reason it says the package isnt available

mcsnubbs:
Removing the wallet file worked. However, after recreating the wallet (multisig), it shows 0 balance although one of the addresses contained by this multisig has valid UTXOs. I’m currently connected to a core client which is synced. Does it need to rescan for valid UTXO’s or is that happening in the background/needs time?

stepansnigirev:
Yes, you need to rescan - you can go to settings of the wallet and start rescan. You can set the start block there, so no need to rescan everything.

mcsnubbs:
I get the following error when I go to wallet settings in Specter:

AttributeError
AttributeError: 'NoneType' object has no attribute 'replace'

stepansnigirev:
Ah... Interesting... I will try to fix it.
Meanwhile you can rescan using command line.

stepansnigirev:
Should be fixed now, if it's not please reopen the issue.

PR #87, opened 2020-03-10
Wallet bugfix

stepansnigirev:
Closes https://github.com/cryptoadvance/specter-desktop/issues/86

- when switching to another node that doesn't have wallets known to specter - hide the wallets instead of crashing
- if derivation path or fingerprint are not known to specter - hide "Download ColdCard file" instead of crashing
- css improvement for key selection - selected key radio button has now opacity of 1


stepansnigirev:
@k9ert can you review this?
Also, tests are not running again...

PR #89, opened 2020-03-17
Add custom block explorer URL settings

ben-kaufman:
Support customizing the block explorer URLs for `mainnet` and `testnet` from the settings page.
This gives both a new customization option for the user and enables using a local block explorer like [btc-rpc-explorer](https://github.com/janoside/btc-rpc-explorer/) which connects directly to the user's full node (instead of relying on 3rd party data). This update keeps [blockstream.info](https://blockstream.info) as the default block explorer.

Specter Settings

### DIFF
--- a/src/cryptoadvance/specter/controller.py
+++ b/src/cryptoadvance/specter/controller.py
@@ -110,6 +110,8 @@ def settings():
     port = rpc['port']
     host = rpc['host']
     protocol = 'http'
+    mainnet_explorer = app.specter.config["mainnet_explorer"]
+    testnet_explorer = app.specter.config["testnet_explorer"]
     auth = app.specter.config["auth"]
     if "protocol" in rpc:
         protocol = rpc["protocol"]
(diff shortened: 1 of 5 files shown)


stepansnigirev:
Great! Looks good!
We currently support 4 networks: `main`, `test`, `regtest` and `signet`.
I suggest to make a section on explorers in the config file with default values:
```json
"explorers": {
    "main": "https://blockstream.info/",
    "test": "https://blockstream.info/testnet/",
    "regtest": null,
    "signet": "https://explorer.bc-2.jp/"
}
```

Then explorer for current chain can be accessed as:
```py
explorer = app.specter.config["explorers"][app.specter.chain]
```

Or better if we have a property in Specter something like this:
```py
@property
def explorer(self):
    if "explorers" in self.config and self.chain in self.config["explorers"]:
        return self.config["explorers"][self.chain]
    else:
        return None
```

ben-kaufman:
Thanks for the response @stepansnigirev. I updated the PR. The notable change besides your comment is that now the settings page would allow editing the block explorer only for the network you are currently using. I could allow for all networks but it seems to make the UI messy for no real reason. So now the Settings screen looks like this:

Specter Settings

stepansnigirev:
Works great, thanks! Merged.

Issue #91, opened 2020-03-18
Get rid of HWI as a subprocess


stepansnigirev:
We have exactly one place where we use HWI binary as a subprocess.
https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/views/hwi.py#L50

This means that if hwi is not in PATH everything fails.
We should get rid of that. We can locate hwi.py using `hwilib.__path__` and call it from there instead.

PR #92, opened 2020-03-18
hwi via suprocess and python, no PATH required

stepansnigirev:
Closes https://github.com/cryptoadvance/specter-desktop/issues/91
HWI is called as a subprocess, but using python hwilib, so doesn't require to be in PATH


k9ert:
Nice hack, just merged.

PR #94, opened 2020-03-19
Addresses tab + address labelling and avoid address reuse

ben-kaufman:
- New Addresses tab: Show all addresses of the wallet.
- Support Bitcoin Core address labeling.
- Avoid address reuse: when entering the `Receive` tab of a wallet, if the current address is used, Specter will automatically generate a new address to prevent address reuse.
- Remove address history from the `Receive` tab.
- Add empty tx view to the `Transactions` tab

### DIFF
--- a/src/cryptoadvance/specter/logic.py
+++ b/src/cryptoadvance/specter/logic.py
@@ -756,7 +756,7 @@ class Wallet(dict):
     
     def getaddressname(self, addr, addr_idx):
         address_info = self.cli.getaddressinfo(addr)
-        if "label" not in address_info or address_info["label"] == "" and addr_idx > -1:
+        if ("label" not in address_info or address_info["label"] == "") and addr_idx > -1:
             self.setlabel(addr, "Address #{}".format(addr_idx))
             address_info["label"] = "Address #{}".format(addr_idx)
         return addr if ("label" not in address_info or address_info["label"] == "") else address_info["label"]
--- a/src/cryptoadvance/specter/logic.py
+++ b/src/cryptoadvance/specter/logic.py
@@ -777,10 +777,6 @@ class Wallet(dict):
     @property
     def addresses(self):
         return [self.get_address(idx) for idx in range(0,self._dict["address_index"] + 1)]
-    
-    @property
-    def change_addresses(self):
-        return [self.get_address(idx, change=True) for idx in range(0,self._dict["change_index"] + 1)]
 
     def createpsbt(self, address:str, amount:float, subtract:bool=False, fee_rate:float=0.0, fee_unit="SAT_B"):
         """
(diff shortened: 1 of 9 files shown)


k9ert:
Works for me. I'm just wondering: Do we need to make that configurable? Address reuse is a bad idea and should not even be possible. If someone wants to reuse a specific address for a reason, he could copy the address e.g. from the Transactions-tab. 

So why not remove the configuration-part of this feature?!

ben-kaufman:
I actually left it configurable because you have a history section in the Receive which shows the past transactions of the current address, so it will never show up without address reuse and I didn't know if I should leave it or not.

If you support it then I can just remove that and the configuration-ability.

k9ert:
> I actually left it configurable because you have a history section in the Receive 
> which shows the past transactions of the current address

That's a good point. But even if we would have that feature, you can't browse other addresses then the one which is currently shown and might (like other addresses) have txs or not.
So i think showing addresses and the transactions to these addresses is an important feature which we should have.
Maybe before you remove the configuration, we can discuss here where to list addresses with txs.
1. a new tab? Currently we have:
image
Maybe add a new tab called "Addresses"?!
2. In the Receive-tab have a "older-addresses" button on the bottom of the page which expands to older addresses
3. Any other idea?

I think i tend to 1.

ben-kaufman:
Option 1 sounds good, but instead of making a new tab in that menu I think maybe we could change the `Transactions` tab to `History` and there give the option (using tabs/ dropdown menu etc.) to choose between `Transactions` and `Addresses`. What do you think?

stepansnigirev:
So I am testing the addresses tab and I see that transaction list for change addresses is always empty.
This is because `listtransactions` display only destination address, so we don't know which change address was used.
It is possible to fetch full details of all outgoing transactions, but I don't think it makes sense to flood bitcoin core with so many rpc-calls. 
I suggest to remove change addresses information from the addresses tab for now, and then come back to them when we implement caching of the history.

ben-kaufman:
Oh right, didn't notice about the change addresses. Removed it for now. I also reversed the address order and added index.

src/cryptoadvance/specter/logic.py
```diff
@@ -746,6 +746,17 @@ def keypoolrefill(self, start, end=None, change=False):
...
+    def txonaddr(self, addr):
+        txlist = [tx for tx in self.transactions if tx["address"] == addr]
+        return len(txlist)
+
+    def setlabel(self, addr, label):
+        self.cli.setlabel(addr, label)
+    
+    def getaddressname(self, addr):
+        address_info = self.cli.getaddressinfo(addr)
+        return addr if address_info["label"] == "" else address_info["label"]
```
stepansnigirev: Crashes if there is no label.

```py
return addr if ("label" not in address_info or address_info["label"] == "") else address_info["label"]
```

k9ert:
Tested on mainnet. Work great! For larger wallets probably the loading time of the addresses-tab might get an issue but let's worry about that later.
From my side, ready to merge!

Btw, i also tested the (potential) new release. For the reference:

`python3 -m pip install --upgrade --index-url https://test.pypi.org/simple/ --no-deps cryptoadvance.specter`

PR #95, opened 2020-03-22
UTXO view

ben-kaufman:
- Cache transactions details.
- Update the tx and addresses tabs to utilize the new data to better display txs data
- New UTXO view in the Addresses tab (includes change outputs).
- Added "Label view" to the Addresses tab (groups addresses with the same label, essentially allows having separate "accounts" withing the same wallet).
- Show total balance per address in the Address tab.
- Fix `immature` tx icon on the Address tab.
- Move the `Edit Label` button to the right on the Addresses tab.
- Change address auto-labeling on the Send tab.
- Add "Rebuild Cache" button to the Wallet Settings tab.

### DIFF
--- a/src/cryptoadvance/specter/logic.py
+++ b/src/cryptoadvance/specter/logic.py
@@ -765,7 +765,7 @@ class Wallet(dict):
         return len(txlist)
 
     def balanceonlabel(self, label):
-        balancelist = [utxo["amount"] for utxo in self.utxo if utxo["label"] == label or utxo["address"] == label]
+        balancelist = [utxo["amount"] for utxo in self.utxo if ("label" in utxo and utxo["label"] == label) or utxo["address"] == label]
         return sum(balancelist)
 
     def addressesonlabel(self, label):
--- a/src/cryptoadvance/specter/logic.py
+++ b/src/cryptoadvance/specter/logic.py
@@ -9,10 +9,10 @@ from collections import OrderedDict
 from . import helpers
 from .descriptor import AddChecksum
 from .helpers import deep_update, load_jsons
-from .rpc import RPC_PORTS, BitcoinCLI, autodetect_cli
+from .rpc import RPC_PORTS, autodetect_cli
+from .rpc_cache import BitcoinCLICached
 from .serializations import PSBT
 
-cache = {}
 
 # a gap of 20 addresses is what many wallets do
 WALLET_CHUNK = 20
@@ -56,7 +56,7 @@ def get_cli(conf):
         else:
             return None
     else:
-        cli = BitcoinCLI(conf["user"], conf["password"], 
+        cli = BitcoinCLICached(conf["user"], conf["password"], 
                           host=conf["host"], port=conf["port"], protocol=conf["protocol"])
     return cli
 
@@ -530,7 +530,6 @@ class Wallet(dict):
         self.cli_path = manager.cli_path
         self.cli = manager.cli.wallet(self.cli_path+self["alias"])
         self._dict = d
-        self.setup_cache()
         # check if address is known and derive if not
         # address derivation will also refill the keypool if necessary
         if self._dict["address"] is None:
@@ -538,6 +537,7 @@ class Wallet(dict):
             self.setlabel(self._dict["address"], "Address #0")
         if self._dict["change_address"] is None:
             self._dict["change_address"] = self.get_address(0, change=True)
+        self.cache_addresses()
         self.getdata()
 
     def _commit(self, update_manager=True):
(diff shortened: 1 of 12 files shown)


stepansnigirev:
Tested, works, but looks confusing.
If I look at my addresses I see that I have two incoming transactions, but the balance of the address 0 is only 0.05.
It's because I have a change address that has ~0.9 btc there, but in UTXO view I only have the final non-zero balance without any transactions.
Can we indicate outgoing transactions somehow that transfer funds to change address?

Addresses tab:
image

UTXO tab:
image

ben-kaufman:
The UTXO view should display the incoming transaction. I had there a bug where if 2 UTXOs came from the same transaction it would only display one, maybe that was it... Now fixed it, can you please test again?

ben-kaufman:
The harder thing would be to show outgoing transactions associated with an address. Bitcoin Core doesn't give this info easily (afaik). You have to loop over the `send` transactions and then get their inputs with `getrawtransaction` and loop over them to see the address it belongs to.  That's possible of course but not very clean and efficient so not sure if we should add that.

We could add the change addresses which have UTXO to the Adddresses tab so it shows up there too without showing irrelevant change addresses, that would be an easier solution I think.

The UTXO tab not showing up the tx is a bug though, hopefully fixed with the last commit (please check if it solved it for you).

src/cryptoadvance/specter/logic.py
```diff
@@ -751,8 +756,32 @@ def txonaddr(self, addr):
...
+    def balanceonaddr(self, addr):
+        balancelist = [utxo["amount"] for utxo in self.utxo if utxo["address"] == addr]
+        return sum(balancelist)
+
+    def txonlabel(self, label):
+        txlist = [tx for tx in self.transactions if (tx["label"] == label if "label" in tx else tx["address"] == label)]
+        return len(txlist)
+
+    def balanceonlabel(self, label):
+        balancelist = [utxo["amount"] for utxo in self.utxo if utxo["label"] == label or utxo["address"] == label]
```
stepansnigirev: please add a check that utxo has "label" attribute. Crashes if my change address doesn't have a label.

stepansnigirev:
> We could add the change addresses which have UTXO to the Addresses tab so it shows up there too without showing irrelevant change addresses, that would be an easier solution I think.

I think this should be enough. You expect UTXO tab to show complete UTXO set :)

> The UTXO tab not showing up the tx is a bug though, hopefully fixed with the last commit (please check if it solved it for you).

Not really, still empty tx list, I think it's because we don't show outgoing transactions that send money to change addresses.

Also breaks when change address doesn't have a label...

I think we need to start caching transactions next to the wallet json file, so we could have get more flexibility in displaying and sorting transactions of the wallet without rpc calls.

ben-kaufman:
I fixed the crash. Even for a change address the UTXO view should show all unspent transactions, I tried to make it happen myself but can't reproduce this issue, can you please give more details on how to reproduce it?

Caching would make this all much easier and faster indeed, I could try to add it too.

stepansnigirev:
Right now UTXO view looks correct. At least everything adds up together to the addresses balance.
History view is weird:
image

At the moment I only browse existing wallets, didn't create new transactions. I will try to reproduce with a new wallet.

stepansnigirev:
I created a new wallet, sent two transactions (0.1 and 0.15) to the first address.
Then I sent 0.05 to an external address.
So I would expect to see 3 transactions in the tx list, but I got this:
image

In the address tab transactions on Change 0 look weird - balance is non zero, but two transactions:
image

Looks like we are double-counting transactions somewhere...

stepansnigirev:
I think we need to show transaction once, but show both external address we are sending money to, but also internal transfer to the change address.

ben-kaufman:
So if UTXO view is working that's already good :) (let me know if it breaks again, but I think it shouldn't).

The issue with the Addresses history view is a hard one, and it's related to how Bitcoin Core saves the tx list. Right now we display it basically the way Bitcoin Core returns the list (it returns with duplicates, kind of...). We should probably first add caching before we can really solve it. So I'd suggest to keep it like that for now and I'll work on caching which will solve that (or wait with merging the PR until we have caching for that).

stepansnigirev:
Let's mark this PR as work in progress, and merge it when we have caching.
What do you think would be the best way to cache transactions?
How about a `.csv` or `.tsv` file with the same name?

k9ert:
> What do you think would be the best way to cache transactions?
> How about a .csv or .tsv file with the same name?

We could also consider to cache only in memory?  If needed, the cache could be completely rebuild from Core data? That way we could avoid complex cache-updating-logic on disk? 

I couldn't test it yet, this is mainly about the getdata method and the utxoaddresses-property, right?

ben-kaufman:
I think it's just a question of performance, so we could start with caching only in memory and if it's too slow also save it on disk (I think it should be good enough on memory, but we can see...).

For the changes in this PR, I listed the feature/ fixes it added in the top comment (I edit it when I commit something new).

ben-kaufman:
Just so that we are synced on what we want to achieve with the caching:
- Transactions Tab - List all transactions (without duplicates), for each show:
  - Tx ID
  - Sent/ Received
  - Address: If sent - to which address (not showing change address), If received - to which address (Question: what if received to multiple addresses?)
  - Amount (without change amounts)
  - Confirmations
  - Time

- Addresses (History) Tab - List all addresses, for each show transactions with:
  - Tx ID
  - Sent/ Received
  - Address: If sent - to which address (not showing change address), If received show the address itself? (Question: or maybe show where it came from? But then what if it cam from multiple addresses?)
  - Amount (Question: only amount from that address or for the entire transaction?)
  - Confirmations
  - Time

- Addresses (UTXO) Tab - List all addresses, for each show UTXO with:
  - Tx ID
  - Address: Show the address itself? (Question: or maybe show where it came from? But then what if it cam from multiple addresses?)
  - Amount - only amount of the specific output (Question: or for the entire transaction?)
  - Confirmations
  - Time

Does this make sense?

ben-kaufman:
Hi, I added a basic caching of transactions details in the app memory.
This will now allow us to show transactions data more accurately (and if we want also in more details) in the Transactions and Addresses tabs. This could be further optimized and utilized for more features, but I think this is a good start.
In the UI the most notable change is that now the Addresses tab will show more data and the balance per address will make sense, but we could use the new data available from this caching for more UI features in the future.
Also, right now the cache is saved in the app itself and will be built from scratch every time the server restarts. I think for the current level of caching this should be fine but in the future, we should consider saving this in a `json`/ `csv` file.

stepansnigirev:
I got an error:
```
{"result":null,"error":{"code":-5,"message":"No such mempool transaction. Use -txindex or provide a block hash to enable blockchain transaction queries. Use gettransaction for wallet transactions."},"id":0}
```
You are using `getrawtransaction` that requires tx index. As recommended in the error we should use `gettransaction` on wallet rpc instead.

ben-kaufman:
Thanks @stepansnigirev , I removed the need for `txindex`, should be good now.

stepansnigirev:
Works like a charm!
@k9ert will you find time to review or I can just merge?

stepansnigirev:
Another bug:
when creating a new wallet there is no cache for it yet. And... everything breaks :(
Need fixes for new wallets.

Breaks here and not only here:

```
File "/Users/ss/dev/web/forks/ben_kaufman/src/cryptoadvance/specter/logic.py", line 519, in __next__
return Wallet(v, manager=self)
File "/Users/ss/dev/web/forks/ben_kaufman/src/cryptoadvance/specter/logic.py", line 537, in __init__
self.setlabel(self._dict["address"], "Address #0")
File "/Users/ss/dev/web/forks/ben_kaufman/src/cryptoadvance/specter/logic.py", line 1015, in setlabel
Open an interactive python shell in this frameold_label = cache[self["name"]]["labels"][addr] if addr in cache[self["name"]]["labels"] else addr
```

ben-kaufman:
Fixed it, thanks!

stepansnigirev:
Checked, works now :)

src/cryptoadvance/specter/templates/wallet_addresses.html
```diff
@@ -48,45 +51,75 @@ <h1><small style="line-height:30px">Total balance:<br></small><span style="color
...
-{% for address in wallet.addresses | reverse %}
-{% set addrlabel = wallet.getaddressname(address, wallet.addresses | length - loop.index0 - 1) %}
+
+<div style="padding: 0 30px; width: 100%;">
+	<div class="row padded" style="width: intrinsic; margin: auto;">
+		<a href="./?all=True&view={{viewtype}}" class="btn radio left {% if alladdresses %}checked{% endif %}">History</a>
+		<a href="./?all=False&view={{viewtype}}" class="btn radio right {% if alladdresses != True %}checked{% endif %}">UTXO</a>
+	</div>
+	<span>View Type: </span><br>
+	<div class="dropdown" style="margin-top: 5px;">
```
k9ert: I'm not sure why it's worth the effort? The only difference seem to be that the adresses disappear?

ben-kaufman: My thought with adding that was to allow grouping multiple addresses with the same label, creating an "account"-like view within the wallet. So if you want to manage multiple "accounts" within the same wallet you give each address the name of the account as its label and then you can see all its addresses and outputs grouped together.
Another way to implement an "accounts" feature is by utilizing the account system of the BIP32 derivation path, but this would be much more different and more complicated.

k9ert: Ah, got it, neat :-). Really like that.

src/cryptoadvance/specter/logic.py
```diff
@@ -583,23 +586,42 @@ def _check_change(self):
         self._dict["change_address"] = self.get_address(self._dict["change_index"], change=True)
         self._commit()
 
+    @property
+    def txlist(self):
```
k9ert, line 590: Can we have a documentation for this property? I guess my limited understanding is somehow:
''' The last 1000 transactions for that wallet but somehow filtered because ... ?! '''

src/cryptoadvance/specter/logic.py
```diff
@@ -583,23 +586,42 @@ def _check_change(self):
...
+        txidlist = []
+        txlist = []
+        for tx in self.transactions:
+            if tx["is_change"] == False and (tx["is_self"] or tx["txid"] not in txidlist):
+                txidlist.append(tx["txid"])
+                txlist.append(tx)
+        return txlist
+
+    def cache_addresses(self):
```
k9ert, line 599: Same here

src/cryptoadvance/specter/rpc_cache.py
```diff
@@ -0,0 +1,36 @@
+from .rpc import BitcoinCLI
+from .corecache import CoreCache
+
+class BitcoinCLICached:
+    def __init__(self, user="", passwd="", host="127.0.0.1", port=8332, protocol="http", path="", timeout=30, cli=None, **kwargs):
```
k9ert, line 5: This is accepting two kind of constructors for that class. This is not very clear from the code. Can you rewrite this as described here: https://stackoverflow.com/questions/141545/how-to-overload-init-method-based-on-argument-type ?

ben-kaufman: Added, let me know if it's good now.

stepansnigirev:
Got an error:
AttributeError: 'function' object has no attribute 'cache_addresses'

```py
 def cache_addresses(self):
        ''' Caches the wallet addresses to the CachedCLI
            This is needed for caching tranactions inputs and outputs baseed on whetever the wallet owns them or if they are change addresses.
        '''
        self.cli.cache.cache_addresses(self.full_addresses, self.change_addresses)
```

I don't quite understand why. `cache` is defined as `@property`...

ben-kaufman:
Very strange, I'm unable to reproduce this error. Anyway, I have implemented a better way to cache addresses. This new way has two main improvements:
- It separates the address caching from the logic code and moves it to the `rpc_cache` where it should be. 
- It will scan for old used addresses and import them if found. This means that, for example, when importing old wallets used outside of Specter (like when a user migrates from a wallet official software to Specter) and running `rescanblockchain`, Specter will now automatically identify and import all used addresses (as compared to the present behavior where the UI have issues like not showing all addresses and txs).
This is very important (and convenient) to allow easy migration of a used wallet to Specter.

Anyway, now there is no `cache_addresses` method and everything should work well.

stepansnigirev:
So it doesn't work if cli was detected using `autodetect_cli` from `rpc.py`. To reproduce: delete rpcuser and rpcpassword from config.

I suggest to pass an optional kwarg to `autodetect_cli` - the class it should use to instantiate from.
Or, even better, refactor `autodetect_cli` to return not an instance of BitcoinCLI, but just a config that it checked to be working.

ben-kaufman:
Thanks @stepansnigirev , I changed it according to the latter suggestion. I tried it with auto_detect locally and works well now, so should be good to go I think.

stepansnigirev:
Works great, thanks! Merging :)

PR #98, opened 2020-04-08
Sorting wallet and devices and some style fixes

stepansnigirev:
Paddy reported a bug when sending transactions from the wallet - after broadcasting specter was not able to render `/tx/` page.
It was happening because of a weird bug in the `__next__` implementation of the WalletManager - order of keys in the dictionary can change during the runtime, sorting them helps to fix the issue :)

Sorting also helps to avoid reordering of the wallets in the sidebar - they stay in place now.

This pull request also fixes a download button style issue - it was huge...


stepansnigirev:
As there are no reviews I am merging

PR #99, opened 2020-04-08
fix usage with xpubs without derivation information

stepansnigirev:
Allows to use master public keys without any derivation information or fingerprint.
Checked in 1-of-2 with coldcard and proof wallet (https://github.com/hodlwave/proof-wallet)

@hodlwave your airgapped laptop wallet should work fine with Specter now.


stepansnigirev:
As there are no reviews I am merging

PR #100, opened 2020-04-08
Add Pending txs in Send tab

### LINKED ISSUE #7: Store pending transaction
If you navigate away from the send screen while signing on a device, then you can't resume the transaction.

ben-kaufman:
Resolves #7 

Changes:
- When a new transaction is created, save it to a list of pending transactions, and lock the inputs it uses.
- Add a `Pending` tab to the `Send` screen - listing all pending transactions and allows to either continue a transaction or delete it (which will unlock its inputs).
- Indicate to the user which devices already signed a transaction in a multisig wallet.
- When loading the wallet into Bitcoin Core, lock all UTXOs used by pending transactions.
- If there are locked UTXOs indicate this in the available balance in the Send screen.

### DIFF
--- a/src/cryptoadvance/specter/controller.py
+++ b/src/cryptoadvance/specter/controller.py
@@ -49,9 +49,8 @@ def combine(wallet_alias):
         if "hex" in raw:
             app.specter.broadcast(raw["hex"])
 
-        if wallet.is_multisig:
-            device_name = d['device_name']
-            wallet.update_pending_psbt(psbt, txid, device_name)
+        device_name = d['device_name']
+        wallet.update_pending_psbt(psbt, txid, device_name)
         return json.dumps(raw)
     return 'meh'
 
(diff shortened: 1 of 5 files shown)


stepansnigirev:
When I created, signed and broadcasted the transaction it was not removed from the list of pending transactions.

stepansnigirev:
Also when I tried to delete the transaction from pending that already was mined I got an RPC error:
`cryptoadvance.specter.rpc.RpcError: Server responded with error code 500: {"result":null,"error":{"code":-8,"message":"Invalid parameter, expected unspent output"},"id":0}`

ben-kaufman:
@stepansnigirev you're right, sorry, I changed a few stuff before pushing and didn't fully test it again after doing so. I fixed the issue and it should now be working well. I also found out an issue with the full and available wallet balance when locking unconfirmed utxos, so fixed that as well. This should be ready to go now.

ben-kaufman:
Closing and opening to try trigger travis.

stepansnigirev:
Tested, works! Awesome! Merging it!

PR #101, opened 2020-04-12
Support Device Passphrase

ben-kaufman:
Support using devices that have passphrase protection enabled - when passphrase protection is on, the user will be prompted to enter the passphrase for wallet interactions.
Also, notify users when the wallet they use is not initialized (instead of stuck UI as there is no).

### DIFF
--- a/src/cryptoadvance/specter/templates/hwi_new_device_xpubs.html
+++ b/src/cryptoadvance/specter/templates/hwi_new_device_xpubs.html
@@ -77,6 +77,7 @@
 		formData.append("device_name", document.getElementById('device_name_id').value);
 		formData.append("type", hwiCurrentWallet.type);
 		formData.append("path", hwiCurrentWallet.path);
+		formData.append("passphrase", hwiWalletPassphrase);
 
 		const response = await fetch(
 			'/hwi/extract_xpubs/',
--- a/src/cryptoadvance/specter/templates/includes/hwi.html
+++ b/src/cryptoadvance/specter/templates/includes/hwi.html
@@ -233,7 +233,7 @@
         var formData = new FormData();
         formData.append("type", hwiCurrentWallet.type);
         formData.append("path", hwiCurrentWallet.path);
-        formData.append("hwiWalletPassphrase", hwiWalletPassphrase);
+        formData.append("passphrase", hwiWalletPassphrase);
         formData.append("psbt", psbt);
 
         try {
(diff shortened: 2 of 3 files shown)


stepansnigirev:
Awesome! I never used passphrases on Trezor, so now it's a good chance to try :)

ben-kaufman:
If you have never used the passphrase feature of Trezor you'll probably have to enable this capability of the device. You could do this either the normal way by going to Trezor's website and enabling it there or the cool way of helping me test this PR to the HWI repo ;)
https://github.com/bitcoin-core/HWI/pull/318

If/ once that PR is merged to the HWI repo I think it will be very useful to add support for it in Specter UI as well.

stepansnigirev:
I tested with KeepKey.

HWI pull request works, kinda... 
If I want to have two wallets - one with empty passphrase and another one with non-empty passphrase, how can I use them? Do I need to toggle passphrase feature every time? 
When I tried with `--password ""` it was complaining... Otherwise works.

On Specter side - getting xpubs worked, but then I got an alert "String is expected" when I tried to sign a transaction. Can you double-check that all hwi requests get the password if it was provided? Especially sign_tx...

stepansnigirev:
More precisely, first error appeared in `hwiWalletDetectedHandler` javascript function.
Console showed error "Could not open client or get fingerprint information: Passphrase needs to be specified before the fingerprint information can be retrieved", so obviously passphrase was not provided.
Second error appeared in `hwiSendPsbt`:
"String value expected"

ben-kaufman:
@stepansnigirev regarding the first error message, I think that's normal before you passed the passphrase, just like before you passed the PIN, so it shouldn't be an issue.

I do see the issue with having no password (`--password ""`) on the HWI side. I'm not sure if that's a problem as if you do have the `passphrase` enabled, it might make sense to require a not empty passphrase. I don't have a strong opinion about this though so if you think it should allow `--password ""` let me know and I can write a change for that.
Anyway, this should not be an issue for Specter, as even with empty passphrase given, it works just fine (at least for me...) and uses the same keys as if the passphrase was disabled, so from Specter's side this shouldn't be an issue.

As for the issue with `sign_tx`, I made a mistake there in the code and gave the wrong name when sending the passphrase parameter there. I fixed that in the last commit, it should be working well now (tested it locally with a KeepKey).

stepansnigirev:
Checked, works with and without the passphrase, including empty passphrase.
No need to change anything!
Awesome! Merging :)

PR #102, opened 2020-04-14
Fix sidebar UI issue in Safari

ben-kaufman:
fix an issue with the sidebar on Safari where too many wallets/ devices caused elements in the sidebar to overlap.


stepansnigirev:
Awesome, thanks! Works!

PR #103, opened 2020-04-17
Support bitcoin.conf network sections auto-detect

### LINKED ISSUE #10: get_rpcconfig bug
You can specify network-specific parameters in bitcoin.conf in [sections like these](https://github.com/bitcoin/bitcoin/blob/master/share/examples/bitcoin.conf#L164).

The `rpc.get_rpcconfig` function in this project can't parse this options.

ben-kaufman:
Resolve #10
Support `bitcoin.conf` network specific parameters in `auto_detect`.

### DIFF
--- a/src/cryptoadvance/specter/rpc.py
+++ b/src/cryptoadvance/specter/rpc.py
@@ -18,7 +18,12 @@ def get_default_datadir():
 def get_rpcconfig():
     path = get_default_datadir()
     config = {
-        "bitcoin.conf": {},
+        "bitcoin.conf": {
+            "default": {},
+            "mainnet": {},
+            "testnet": {},
+            "regtest": {}
+        },
         "cookies": [],
     }
     if not os.path.isdir(path): # we don't know where to search for files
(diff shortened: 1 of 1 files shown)


src/cryptoadvance/specter/rpc.py
```diff
@@ -28,12 +33,22 @@ def get_rpcconfig():
     if os.path.exists(bitcoin_conf_file):
         try:
             with open(bitcoin_conf_file, 'r') as f:
+                current = config["bitcoin.conf"]["default"]
                 for line in f.readlines():
                     line = line.split("#")[0]
+                    if '[main]' in line:
```
stepansnigirev, line 39: Here if you would use `main` and `test` as keys you could avoid copy-paste:
```py
for net in config["bitcoin.conf"]:
    if f"[{net}]" in line:
        current = config["bitcoin.conf"][net]
#...
```

src/cryptoadvance/specter/rpc.py
```diff
@@ -64,24 +79,21 @@ def get_configs(config=None):
...
+        if "rpcuser" in config["bitcoin.conf"][network]:
+            default["user"] = config["bitcoin.conf"][network]["rpcuser"]
+        if "rpcpassword" in config["bitcoin.conf"][network]:
+            default["passwd"] = config["bitcoin.conf"][network]["rpcpassword"]
+        if "rpcconnect" in config["bitcoin.conf"][network]:
+            default["host"] = config["bitcoin.conf"][network]["rpcconnect"]
+        if "rpcport" in config["bitcoin.conf"][network]:
+            default["port"] = int(config["bitcoin.conf"][network]["rpcport"])
+        if "user" in default and "passwd" in default:
+            if "port" not in default: # only one bitcoin-cli makes sense in this case
```
stepansnigirev, line 92: This will work only once, then it will be populated by mainnet config and all other nodes will not be detected.
Should be
```py
if "port" not in config["bitcoin.conf"]["default"]
```

src/cryptoadvance/specter/rpc.py
```diff
@@ -18,7 +18,12 @@ def get_default_datadir():
 def get_rpcconfig():
     path = get_default_datadir()
     config = {
-        "bitcoin.conf": {},
+        "bitcoin.conf": {
+            "default": {},
+            "mainnet": {},
+            "testnet": {},
```
stepansnigirev, line 24: please use `main` and `test` as keys here, otherwise lookups in RPC_PORTS don't work and other things are more complicated as well. See comments below.

ben-kaufman:
@stepansnigirev thanks, fixed it.

stepansnigirev:
Thanks, merging!

PR #97, opened 2020-04-02
Security fix

k9ert:

PR #104, opened 2020-04-17
feature: qrcode animations

gorazdko:
## Abstract

* This implements qrcode animations for large qr codes
* related to https://github.com/cryptoadvance/specter-diy/pull/58

## Status

Ready for review:
- [x] ~add buton to the component to~ toggle between static and dynamic qr mode on a mouse click
- [x] tune parameters: What do you think about speed and size of QR animation. Should we make it faster?
  * currently max 300byte frames are animated per 300ms
- [x] cleanups

clicktoanimate


stepansnigirev:
Just merged https://github.com/cryptoadvance/specter-diy/pull/58, works very smoothly.

I am thinking, if we are adding QR code generation javascript library to the frontend side, we could easily remove the dependency to Flask_QRCode - and many more underlying dependencies like Pillow and others.
I think we should migrate to browser-based QR code generation.

k9ert:
> ... and many more underlying dependencies like Pillow and others.
> I think we should migrate to browser-based QR code generation.
Just lately i fixed a security-issue in Pillow (dependency for Flask_QR just by updating the version #97. The issue got reported by the security-alerting-system:
https://github.com/cryptoadvance/specter-desktop/network/alerts

We would loose that ability. If we start to move dependencies from flask/pip to javascript, we should imho try to use npm for that purpose. Maybe i should have started to do that already with vuejs but ... yeah it's some slippery-slope.

src/cryptoadvance/specter/static/vue_qr.js
```diff
@@ -0,0 +1,83 @@
...
+ * "p3of6 q2y9L7JQQSrNYx2FoTzFdzAtSHntnGNy4aeDjfLxJa7wKE5czu",
+ * "p4of6 ExXWFzerNsiWo,[9d1dc604/48h/1h/0h/2h]tpubDEAMhNB9p",
+ * "p5of6 UL6Ej2HuiJUU1TDMqHNxRKjGFxLfCB7cmxKjQKXNF4yR8CMpgD",
+ * "p6of6 t5eh5V3XesKDUrcFHYDMz3u3ybWwHSAZZQLjiMAi4oKy6HEg))" ]
+ *
+ * param[in] text
+ * param[in] width
+*/
+Vue.component('qrencode',{
+  template : '<div :title="title" v-on:click="clicked()"><qrcode :value="qrval" :options="wd"></qrcode></div>',
```
stepansnigirev, line 17: Alt is not very obvious, I suggest also adding a note below the large QR code:
`'<div :title="title" v-on:click="clicked()"><qrcode :value="qrval" :options="wd"></qrcode><div class="note">{{title}}</div></div>'`

stepansnigirev:
@gorazdko great work! I like how it works a lot!!!
I think the animation rate is good enough, let's keep it like this.

One thing I noticed - when I tried to create a transaction more than 3k in size there was no QR code for full psbt and a js error appeared stating that QR code data is too big.
I suggest showing animated QRcode right away if QR code size is larger than the library limit.
There is no way to scan such a large QR code without animations in any case :)

We should also double-check with Specter-DIY how we manage super-large QR codes, but it is a different issue...

gorazdko:
Rebased and pushed changes

We use default settings: level M with automatic encoding. Automatic encoding for random strings which consist of `abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789` means byte encoding. Therefore i set maximum limit at 2300. My phone cannot correctly decode such a large qr code.

pr55

stepansnigirev:
Awesome, merging!

PR #105, opened 2020-04-18
Fix fullbalance wrong condition

ben-kaufman:
- Fixed wrong condition in getting the full balance
- Moved locked UTXO get balance code to `getbalances` (so it imitates the `getbalances` behavior correctly for unconfirmed locked UTXO).
- Delete a `print` left by mistake.


stepansnigirev:
Looks good, merging

Issue #26, opened 2019-11-02
Improvement: mobile-friendly UI


kdmukai:
* Collapse the left nav into a burger menu
* Change the QR code displays to use the includes/overlay.html popups to maximize display without clipping?

Fonta1n3:
How can I run Specter on my mobile device? Sorry for the newb question.

stepansnigirev:
As it's a web interface to your node you can get access to Specter from mobile - over Tor or if you are using VPS like me.
You can't sign with USB hardware wallets in this case, but still you can monitor funds, use Specter-DIY, Cobo, Electrum and hot wallet from mobile.
At the moment web interface looks pretty bad on phones, but hopefully we will get it decent as well.

Fonta1n3:
So create a hidden service for the specter port and open in a tor browser? That is really very cool.

stepansnigirev:
Yes, you already can do it: https://github.com/cryptoadvance/specter-desktop/blob/master/docs/tor.md
There will be problems with the camera access if you don't use self-signed certificates, but otherwise works :)

ben-kaufman:
Closing for now as solved with #731 
We could still improve on mobile friendly-ness but for now, I think it's good enough. Please reopen/ make a new issue if you see a need.

PR #112, opened 2020-04-23
mobile-friendly-ish styles

stepansnigirev:
Getting started with https://github.com/cryptoadvance/specter-desktop/issues/26
For now only hiding sidebar if the screen is small.
Navbar appears when hovered, for touchscreens if clicked on the arrow.

ezgif-7-12ad306c0c46

### DIFF
--- a/src/cryptoadvance/specter/static/styles.css
+++ b/src/cryptoadvance/specter/static/styles.css
@@ -794,9 +794,6 @@ a:focus + .tool-tip .tool-tip__info {
 }
 
 @media (max-width: 1100px){
-  main{
-    padding-left: 15px;
-  }
   #side-content{
     position: absolute;
     height: 100%;
@@ -815,6 +812,9 @@ a:focus + .tool-tip .tool-tip__info {
     max-width: 230px;
   }
   @media (hover: hover){
+    main{
+      padding-left: 15px;
+    }
     #side-content:hover{
       position: absolute;
       left: 0px;
@@ -846,9 +846,6 @@ a:focus + .tool-tip .tool-tip__info {
   .card{
     border-width: 0px;
     width: 100%;
-  }
-  .btn.radio.left, .btn.radio.right{
-    border-radius: 0;
   }
   nav.row{
     padding-left: 0;
--- a/src/cryptoadvance/specter/static/styles.css
+++ b/src/cryptoadvance/specter/static/styles.css
@@ -120,6 +120,11 @@ nav.top{
     border-bottom: 1px solid var(--cmap-bg-lightest);
     -webkit-app-region: drag;
 }
+nav.row{
+    width: 100%;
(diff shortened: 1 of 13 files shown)


ben-kaufman:
That's very cool! great start for mobile. Tested on desktop and mobile and looks nice.
I have encountered 2 issues with that though.
First is when you have log in required, the logout button covers the top of the sidebar. (see image)

pr112


Second is that on a mobile device, it is nearly impossible to click the sidebar to expand it. I think it'd be better to increase the width a bit or maybe change it to some larger button in case the device is mobile.

stepansnigirev:
I plan to continue working on this PR, so I will mark it as WIP.

@ben-kaufman thanks for testing it, I will pay attention to the login button.

Next - scale for mobile:
`<meta name="viewport" content="width=device-width, initial-scale=1.0">`
 and making sure it fits on the screen

stepansnigirev:
Made some changes. It still looks bad on mobile in portrait mode, but in landscape looks reasonable.

@ben-kaufman can you test and give me some feedback?

ben-kaufman:
@stepansnigirev works great!
The new menu button for touchscreens is really nice :)
I tried on both a phone and a tablet. On the tablet it works great on both portrait and landscape. On a phone the landscape is fine, but I guess that for real support of such small screens we'll need many more changes, but for now this is very good.

src/cryptoadvance/specter/static/styles.css
```diff
@@ -423,7 +446,7 @@ input:hover{
 }
 .note{
     font-size: 0.7em;
-    width: 400px;
+    max-width: 400px;
```
ben-kaufman, line 449: Seems like this causes the note on the send screen (showing the available coins and sat/btc conversion to appear a bit strange.

stepansnigirev: hopefully better now, added a few tweak.
In the `.card` element note will be left-aligned.

stepansnigirev:
Yes, for mobile in portrait mode we would need many changes.
I would like to keep this PR for initial mobile & tablet support and get to the phone version a bit later.

ben-kaufman:
Looks good now on Chrome, but suddenly very bad on Safari (for example, the menu is visible on desktop, and in the Addresses tab the History/UTXO is strangely stretched).

ben-kaufman:
Hmmm but I see that without your changes Chrome looks quite bad. So it's better that we support Chrome properly but best if we can make it look good on both.

stepansnigirev:
can you make screenshots? for me in safary looks ok...
Maybe safary cached something? You need to cmd+option+e to clear the cache.

Screenshot 2020-05-12 at 17 21 13

ben-kaufman:
Yes, apparently it was caching issue so nevermind. Works great!

PR #113, opened 2020-04-23
Fix issue with Daemonize

ben-kaufman:
I've been experiencing an issue with the `--daemon` option of Specter not working, and not even outputting any crash logs or something of the sort. I've searched a bit aand looks like this line in the Daemonize lib is where it fails: https://github.com/thesharp/daemonize/blob/39a880f74c1d8ea09fa44fe2b99dec1c961201a9/daemonize.py#L136

When disabling this the daemon starts working for me.

PR #116, opened 2020-04-26
Another way to fix issue with Daemonize

stepansnigirev:
When running a `daemon` starts app with `debug=False`


ben-kaufman:
Works great, thanks!
Closing #113

PR #118, opened 2020-04-29
support scanning animated QR codes

stepansnigirev:
related to https://github.com/cryptoadvance/specter-diy/pull/55

### DIFF
--- a/src/cryptoadvance/specter/controller.py
+++ b/src/cryptoadvance/specter/controller.py
@@ -17,6 +17,7 @@ from .helpers import normalize_xpubs, run_shell
 from .descriptor import AddChecksum
 
 from .logic import Specter, purposes, addrtypes, get_cli
+from .rpc import RpcError
 from datetime import datetime
 import urllib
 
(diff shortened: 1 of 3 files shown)


gorazdko:
doesnt work for me. I get a warning: the camera stream is only accessible if the page is transferred via https.

Prepending `https://` then gives: ssl received a record that exceeded the maximum permissive length

stepansnigirev:
Are you running specter on localhost?
Camera is available either if you use localhost, or over https

gorazdko:
Yes im running localhost. Camera works perfectly fine for non-animated QRs but not for animated ones.

stepansnigirev:
Hm... Interesting... I'll double-check with firefox.

gorazdko:
i think i made mistake, didnt check out the branch correctly

stepansnigirev:
I added a QR progress that shows which parts are already scanned and which not.
Displays "👻" for every scanned QR code and "❔" for every part that is not scanned yet.

In reality I don't even notice this progress bar because all parts are scanned very fast.

stepansnigirev:
>  your master branch didnt return anything in the console, now a null is returned if the transaction cannot be parsed. A descriptive warning would be nicer.

Will do, thanks!

stepansnigirev:
@gorazdko not sure if it is what you meant,
but now web browser alerts when it failed to combine transactions.

gorazdko:
I tested also scanning a single random qr and the popup is there which is correct. Works great!

k9ert:
Great! I've tested the animated gif feature (animating on the PC). Was not able to test the other direction but my webcam-situation is ... horrible. Bad hand with webcams.
I guess we can merge that and then start final testing? 
Maybe we should release finally on halving day :-) ?

stepansnigirev:
> Maybe we should release finally on halving day
good idea!

Issue #79, opened 2020-02-27
Enhancement: add "verify address on the device" button


stepansnigirev:
HWI has a method to display address on the device. It works for single key wallets, for multisig it is not available yet, but I think it makes sense to implement this feature. Kinda critical for hardware wallets.

kdmukai:
The KeepKey (and presumably Trezor) is able to verify multisig addrs, at least on Electrum's implementation.

stepansnigirev:
Yes, Trezor and KeepKey can display multisig addresses, but it is not implemented in HWI.
We are also missing change verification on Trezor when using HWI.

stepansnigirev:
Here is the pull request that will enable this feature:
https://github.com/bitcoin-core/HWI/pull/279

PR #120, opened 2020-04-30
Display address on device

ben-kaufman:
Added "Display Address on Device" button to the Receive tab for single signature wallets. Multisig support could be added in a similar way once HWI supports that (https://github.com/bitcoin-core/HWI/pull/324)
Solve #79

### DIFF
--- a/src/cryptoadvance/specter/templates/wallet_receive.html
+++ b/src/cryptoadvance/specter/templates/wallet_receive.html
@@ -1,5 +1,8 @@
 {% extends "base.html" %}
 {% block main %}
+
+{% include "includes/hwi.html" %}
+
 <h1>{{wallet.name}}</h1>
 {% set active_menuitem = "receive" %}
 {% include "includes/wallet_menu.html" %}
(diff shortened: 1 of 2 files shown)


stepansnigirev:
Tested, works great!

@k9ert how about including this into the release as well?

k9ert:
Fine by me! Looks pretty isolated. I guess the current bottleneck for kicking the release out of the door is a hen/egg issue. We need a diy-binary to test desktop. I've tested multisig and the passphrase stuff but not yet the animated QR.

stepansnigirev:
Here is my latest build.

[specter-diy.bin.zip](https://github.com/cryptoadvance/specter-desktop/files/4574177/specter-diy.bin.zip)

k9ert:
I just tried to test this. I'm having issues, good news about this: The negtive part seems to work: "Addresses did not match! Be careful!  ...".
I'm using regtest and the expected address was something like bcr1... but the address on the trezor showed tb1qs...
Any idea why that is?

ben-kaufman:
Yes, I had this issue too and looked in the HWI code, but seems like it’s not possible to specify to the Trezor that regtest is used (seems like it’s not supported by Trezor, not completely sure though).
However what should be possible is to convert that on our side for the verification that the addresses match, so it will still show as testnet address on the device, but our verification will not make an error because of that.

k9ert:
Good enough! Merged.

PR #127, opened 2020-05-09
windows support

stepansnigirev:
Added:
- instructions for windows in docs
- wallet path fix because win uses `\\` everywhere

Daemonize doesn't work on windows, so I moved the import to the part of the code where it's necessary.

Tested on Windows with Trezor.


stepansnigirev:
omg why it detected so many changes??? gonna fix it...

stepansnigirev:
should be good now

PR #128, opened 2020-05-09
Small amounts fix

### LINKED ISSUE #123: use satoshi on send pages
*From the Telegram discussion*

Use satoshis, because floats are bad.
To improve the UX we can add a note under the amount that recalculates satoshis to BTC.

Also "Sending 3.856e-05 BTC to bc1....." doesn't look right.

ben-kaufman:
Fix #123 (note: I kept the amount the backend handles in btc and not sats since this is what Bitcoin Core expects to receive, and using sats would mean double conversion which would be even worse).

This PR fixes the way Specter displays small amounts (remove exponential notation) and allows users to specify sending amount in `sats` instead of BTC.
It also fixes two small bugs, one with the way Specter handles errors in the send screen (specifically, errors when PSBT wasn't created, for example when the amount is too small), the other is the delete button on the signing screen which stopped working properly.

### DIFF
--- a/src/cryptoadvance/specter/controller.py
+++ b/src/cryptoadvance/specter/controller.py
@@ -680,5 +680,9 @@ def derivation(wallet):
 def txonaddr(obj):
     return json.dumps(obj, indent=4)
 
+@app.template_filter('btcamount')
+def btcamount(value):
+    value = float(value)
+    return "{:.8f}".format(value).rstrip("0").rstrip(".")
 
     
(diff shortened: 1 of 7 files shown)


stepansnigirev:
Looks like we are repeating ourselves with this formatting:
`{{'{:.8f}'.format(tx["amount"]).rstrip("0").rstrip(".")`

Maybe we can make this a flask template filter? Something like:

```py
@app.template_filter()
def btcamount(value):
    value = float(value)
    return "{:.8f}".format(value).rstrip("0").rstrip(".")
```

Then in the template we could use `{{tx["amount"] | btcamount }}`

ben-kaufman:
@stepansnigirev Makes sense, add that (also in other places showing balance).

stepansnigirev:
Great! Merging :)

Issue #124, opened 2020-05-09
show more information about transaction on the send page


stepansnigirev:
At the moment it is very limited, only destination and amount.
Let's show more information: used utxo, all outputs with change marked, corresponding address labels, addresses themselves as well.

PR #130, opened 2020-05-12
Add transaction details on signing page

ben-kaufman:
Solves #124 

Adds the ability to see more transaction information (fees, size, outputs with their details, raw PSBT etc.).
I'm definitely not a designer so any ideas on how to make it look prettier are highly welcomed, though I think it will be also fine as is. Here are a few screenshots of how that looks:

Screenshot3
Screenshot1
Screenshot2


stepansnigirev:
Copy didn't work for me.
According to [w3schools](https://www.w3schools.com/howto/howto_js_copy_clipboard.asp) the script needs to select the text that will be copied before copying.

Also instead of using javascript to toggle transaction details I would recommend making it a checkbox and defining visibility in css.
Here is an example including changing the label without any js:
https://jsfiddle.net/uw5pt63b/1/

stepansnigirev:
Also, for inputs it would be nice to see address labels and addresses.
For outputs I would include addresses as well, not just labels.

ben-kaufman:
Updated with all the above, should be good now.

stepansnigirev:
Looks great!

Issue #125, opened 2020-05-09
separate broadcasting from sending


stepansnigirev:
maybe people want to broadcast from other node, or using blockstream.info or other broadcasting service.
When tx is finalized let's offer the user to:
- broadcast from their own node
- broadcast via blockstream.info
- copy raw tx in hex format
- save for later (then tx should stay in pending psbts, but marked as finalized)

We should check with `testmempoolaccept` that tx is ok.

PR #132, opened 2020-05-14
Add new tx broadcast logic and options

ben-kaufman:
Resolves: #125 
Adds a popup when a transaction is finalized allowing to broadcast the transaction via the connected node/ Blockstream node and allows to copy the raw transaction / save it.

I have not fully tested yet the Blockstream send functionality as I'm using regtest, not testnet. So if someone could try that out that'd be great!

### DIFF
--- a/src/cryptoadvance/specter/controller.py
+++ b/src/cryptoadvance/specter/controller.py
@@ -69,7 +69,7 @@ def broadcast(wallet_alias):
     if request.method == 'POST':
         d = request.json
         tx = d['tx']
-        if json.loads(wallet.cli.testmempoolaccept([tx]))['allowed']:
+        if json.loads(wallet.cli.testmempoolaccept([tx]))[0]['allowed']:
             app.specter.broadcast(tx)
             wallet.delete_pending_psbt(wallet.cli.decoderawtransaction(tx)['txid'])
             return jsonify(success=True)
--- a/src/cryptoadvance/specter/controller.py
+++ b/src/cryptoadvance/specter/controller.py
@@ -57,13 +57,25 @@ def combine(wallet_alias):
             return e.error_msg, e.status_code
         except Exception as e:
             return "Unknown error: %r" % e, 500
-        if "hex" in raw:
-            app.specter.broadcast(raw["hex"])
-
         device_name = d['device_name']
-        wallet.update_pending_psbt(psbt, txid, device_name)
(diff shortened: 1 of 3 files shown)


src/cryptoadvance/specter/controller.py
```diff
@@ -69,7 +69,7 @@ def broadcast(wallet_alias):
     if request.method == 'POST':
         d = request.json
         tx = d['tx']
-        if wallet.cli.testmempoolaccept([tx]):
+        if json.loads(wallet.cli.testmempoolaccept([tx]))['allowed']:
```
stepansnigirev, line 72: testmempoolaccept takes an array and returs an array

`if json.loads(wallet.cli.testmempoolaccept([tx]))[0]['allowed']:`

stepansnigirev:
When I copy raw transaction it says "copied PSBT", but actually copied raw transaction.

ben-kaufman:
Fixed.

stepansnigirev:
Also a comment in the code:

controller.py, array index is missing, should be `if json.loads(wallet.cli.testmempoolaccept([tx]))[0]['allowed']:`

https://github.com/cryptoadvance/specter-desktop/pull/132/files/cd2231cd7881715dece79866d09a7630cfde4e28..accb306472ebde4f70a713f874af0dbfcc67fe2e#diff-a386d70e20c16c989ef22377f5ca3398R72

ben-kaufman:
Opps... missed that comment, updated too.

stepansnigirev:
Actually it should be:

`if wallet.cli.testmempoolaccept([tx])[0]['allowed']:`

Just because `cli` returns a dictionary, so you don't need to do json.loads on it.

ben-kaufman:
Ok this time it should really be good (I hope) :)

stepansnigirev:
Looks good! I will keep it for a day, maybe @k9ert will find some time to review :)

k9ert:
Looks good to me. One thing which i think we should look a bit harder, though, is testing. We're currently heavily creating functionality but we're by far not adding enough tests for that. At least for the parts of logic.py, we have a test-concept in test_specter.py. Would be cool to have some addition  there as well?!

ben-kaufman:
I think adding more testing should be part of the refactoring process in #134

stepansnigirev:
Merging. Testing is added as one of the points in #134

Issue #134, opened 2020-05-15
Refactoring and cleaning


ben-kaufman:
Now that Specter is quite stable, there are many parts of the code which could use some refactoring. I'll list here the points which I'm already aware of and will continue to add based on comments here/ other things I notice.

- [x] Remove `nwjs` - this was an old experiment that is not used anymore). - (#135)
- [x] Break down `logic.py` into smaller files - this file has grown pretty huge and could be broken down into its composing classes - which would make it more readable and organized. - (#160)
- [x] Make reusable UI components - We currently have a lot of duplicated code in the front end, which we could turn into Jinja macros/ Custom elements - (#136).
- [x] Refactor frontend - In addition to using components, there is much old code that could be removed/ rewritten more nicely - (#136).
- [x] Refactor logic classes - Currently, these classes (`WalletManager`, `DeviceManager`, `Wallet`, `Device`) are somewhat unstructured, the managers aren't exactly a list nor an object, and the wallet contains some unnecessary data and uses a `_dict` instead of properly defining its attributes. We could refactor these classes to be better specified, more readable, and optimized. - (#160)
- [x] Connect wallet class with the devices - Currently, a wallet is completely separated from the devices it uses. For example, when deleting a device, the wallet will still remain. We should connect them and make them "aware" of each other. - (#160)
- [x] Refactor HWI - The current code we use for HWI is somewhat messy and requires unnecessary extra steps from the user. This refactoring is part of #131. (#143)
- [x] Structure Files - Files today (both front and back end) are somewhat disorganized, we could probably structure them better into directories in a way easier to reason about and navigate - (frontend: #136, backend: #160).
- [x] Remove Vue - We barely use Vue in the frontend, and where we do it's not hard to replace it with Jinja + JS. We should probably consider removing it, as it doesn't make sense to have it if we barely use it - (#136, #141).
- [x] Improve error handling - We use a lot of `try/catch` through the codebase, but in many points we just return `None` or print something or just `pass`. We should refactor those parts to send the user a proper error message/ page (maybe with an option to copy/ report the logs). - (#139, #156, #160)
- [ ] More testing - Currently our tests for Specter are quite scanty, we should probably add much more coverage of the code.

Please comment below with everything I should add to the list.

(_Update:_ Note that I edit this as stuff change.)

stepansnigirev:
I agree with most of the points, one thing that worries me is this one:

[...]

PR #136, opened 2020-05-17
Refactor frontend with macros

ben-kaufman:
This PR refactors the frontend as discussed in #134 
Here I use Jinja (mostly macros) and JS to refactor the UI: removing duplicated code, breaking large files into small components, replacing Vue with JS (except QR codes), and restructuring the files into a new folders structure. I also added documentation to the macros wherever it seemed appropriate. I am not aware of any docstring feature in Jinja, but if something like that exists it’ll be even better.


stepansnigirev:
Looks good to me, let's continue refactoring this way.

ben-kaufman:
I've continued with similar refactoring also for the sidebar, the base.html, and the overlay popups.
@stepansnigirev I think the popups refactor in particular (c83f643) might be useful for you in HWIBridge, as it separates the HWI UI we use (which are in popups) from the rest of the UI into their own components.
Let me know if that looks good or if I should do something differently. Meanwhile, I'll try to continue to refactor the rest of the UI.

stepansnigirev:
One thing I noticed,
In `getdata` function in `logic.py` we should make `self.transactions` and `self.utxo` empty arrays if `cli` request failed:
https://github.com/ben-kaufman/specter-desktop/blob/refactor-ui-macros/src/cryptoadvance/specter/logic.py#L666-L682

```py
    def getdata(self):
        self.balance = self.getbalances()
        try:
            self.utxo = self.cli.listunspent(0)
        except:
            self.utxo = []
        try:
            self.transactions = self.cli.listtransactions("*", 1000, 0, True)
        except:
            self.transactions = []
```
Otherwise iterator in `wallet_addresses` fails.

Also `getbalances` doesn't neet a try-catch case because it has one internally. Maybe makes sense to do there:
https://github.com/ben-kaufman/specter-desktop/blob/refactor-ui-macros/src/cryptoadvance/specter/logic.py#L770
```py
        except:
            r = { "trusted": 0, "untrusted_pending": 0 }
```

As I noticed `None` tend to break things, we should find a way to do it better...

stepansnigirev:
Trying to sign I got the following errors:
```
Uncaught SyntaxError: Cannot use import statement outside a module - new:1143 
```
It is about QR scanning module - you have `import` statement for qrcode scanning module in a normal `<script>` tag, not in `<script type="module">`.
Keep in mind that `script type="module"` has a different scope, so `onclick` in DOM will not work with functions defined in module, so we use `addEventListener` there.

Should be:
```html
<script type="module">
import QrScanner from "/static/qr/qr-scanner.min.js";
QrScanner.WORKER_PATH = '/static/qr/qr-scanner-worker.min.js';
//... all qr scanning stuff
</script>
```

Second error is:
```
Uncaught TypeError: Cannot read property 'forEach' of undefined
    at HTMLDocument.<anonymous> (new:879)
```
it is about this line:
```js
document.forms[0].elements.psbt_type.forEach((o)=>{
//...
```

ben-kaufman:
I removed the HWI form, replaced `forms[0]` with an id, and moved the QR code into module.
I think the QRcode stuff should be refactored into its own macro all across Specter, but we should remove Vue before we can do that. If I remember correctly you already started doing that in the HWIBridge. I'm not very familiar with the code of the QR stuff so if you could do it here too it'd be really helpful. Then e will be able to remove Vue completely, as now this is the only thing we use it for.

k9ert:
Regarding this:

```
    def getdata(self):
        self.balance = self.getbalances()
        try:
            self.utxo = self.cli.listunspent(0)
        except:
            self.utxo = []
        try:
            self.transactions = self.cli.listtransactions("*", 1000, 0, True)
        except:
            self.transactions = []
```
In general, we should be a bit hesitant when it comes to try/except to be used for something else then error-handling. Especially in java/performance it's a no go because it's slowing down a lot. I don't think that this applies here but in general such practices could cover weird errors which are difficult to track. However, on the other side it's simply so easy to use and maybe makes code more compact.
So i would suggest that we leave comments in such cases. Here is an example which is also definitely improvable:
https://github.com/cryptoadvance/specter-desktop/pull/139/files#diff-a386d70e20c16c989ef22377f5ca3398R561
I guess i should have written: "Normal behaviour because we're using that async but the call is designed sync" or something like that?! 
Does that make sense?

stepansnigirev:
I am not sure it is a normal behaviour in our case.
When does this exception happen? For me, it was breaking things on a fresh new wallet that was just created when there were no transactions yet.
I think we should look at the try-except blocks and at methods where we can return `None` and... process these errors somehow differently - show a custom error page maybe with a meaningful error message... We should add this in the refactoring roadmap and work on it next :)

Otherwise, PR looks good, would merge.

ben-kaufman:
I agree it'll probably be best to refactor the error handling all through Specter, so that instead of returning `None`s we'll show the user a proper error screen, maybe with some option to copy/ directly send the logs so we could check them when someone has an issue?
So I added error handling as a new point in #134, maybe it should be done as part of #139 (what do you think @k9ert ?)
Anyway, I think it's unrelated to this specific PR, so @k9ert if you don't have any objection to this one I think we could merge it and continue with the error handling in either #139 or some other issue/ pr.

k9ert:
> I am not sure it is a normal behaviour in our case.
Guess i made some false assumptions there. I'll double check in the other PR.

> I think we should look at the try-except blocks and at methods where we can return None and... 
> process these errors somehow differently - show a custom error page maybe with a meaningful error 
> message... We should add this in the refactoring roadmap and work on it next :)

Yes agree! As a sidenote, i've created a "SpecterError" 
https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/logic.py#L65

It's created in order to be catched by controllers to show error-messages to end-Users.
Here is an example:
https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/logic.py#L996
and in the controller:
https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/controller.py#L486
Meeh, it's inconsitent. Should be "except SpecterError:".
That way we could make sure that we create Meaningful errormessages for the user. Sooner or later (probably rather later)  we only show such messages to the user and logging the detailed problem analysis in the logs (in lower layers).
You could potentially argue whether the extra SpecterError is necessary (i think it is: i18n, security, user-friendlyness) but at least it's definitely better than magic None-values.

My comments definitely shouldn't block merging here. Pretty unrelated. Just wanted to make that remark. So, let's merge after fixing the test. I'm not sure why it's breaking, though.

PR #140, opened 2020-05-26
Fix coldcard import fingerprint bug

ben-kaufman:
There appears to be a tiny bug with how we import ColdCard fingerprint for single sig.
Currently, we parse every byte from decimal to hex, but when the hex is smaller than 10, we leave one digit instead of 2 (ie. `1` instead of `01`, `2` instead of `02` etc.). This returns an invalid fingerprint. This PR makes sure a `0` is added to the single-digit numbers.


stepansnigirev:
Thanks, merged!

PR #141, opened 2020-05-28
remove vuejs, add qr-code custom component

stepansnigirev:
Custom QR code component is defined in `templates/includes/qr-code.html`.
It is included in `base.jinja` so it is available on any page.

Usage:

```html
<qr-code value="text to encode" width="400"></qr-code>
```

Support of custom components is pretty decent in all browsers, only IE users will suffer, but even Edge and Safari support them:
https://caniuse.com/#feat=custom-elementsv1


k9ert:
Tested a bit, looks good to me, merging.

PR #142, opened 2020-05-28
fix coin selection

stepansnigirev:
A bug was introduced in #136 that broke coin selection.
This PR fixes it.


ben-kaufman:
Strange, I remember testing coin selection, but that's clearly a bug. So thanks :) Looks good to me.

Issue #131, opened 2020-05-14
HWI bridge (miniapp)


stepansnigirev:
Hi fellows!
I started writing a miniapp that works purely with HWI. It is pretty much in progress and I would like to discuss it before I go too deep into that.

Here is the link to the app https://github.com/stepansnigirev/hwibridge

The app is not relying on Bitcoin Core and any Specter logic. It works only with HWI and it is fully stateless. It detects devices, parses PSBT, and allows you to sign transactions with relevant hardware wallets. But I am reusing all `static` files of Specter and HWI code as well. So it makes sense to merge it into specter repo when it's ready. I also started playing with [eel](https://github.com/samuelhwilliams/Eel) there, it might be a good way to pack the app into a single binary.

The idea I have is to make a stand-alone app that can be used in two different use-cases:

- Airgap your USB hardware wallets. You can use it on your air-gapped laptop, scan PSBT from Specter on your online machine, sign with hardware wallets and display signed PSBT as QR code as well. Also should work with files transferred over SD card. So both Ledger and Trezor get airgap properties of Specter or Coldcard if you connect them to a cheap 100$ air-gapped laptop.
- Work as a bridge to HWI on your local machine. If Specter is running in the cloud or node-in-the-box, but you can't connect hardware wallets to that machine, you can run this as a bridge and allow certain domains to access HWI api over localhost. 

The second use case is actually my case - I can't run a full node at home because my internet is very shitty, so I run Specter on a cheap VPS. But then I can't use USB hardware wallets, and that's sad. So this bridge would help me. I know that I can configure Specter to talk to remote Bitcoin Core node, but due to large number of requests it is pretty laggy - so for the best performance Specter should run on the same machine as Core. 

The questions are:
- should it be in a separate repo or in Specter?
- what functionality can we move to the bridge and how to integrate remote bridge in Specter?
- what should the user be able to do in the bridge interface?
- what do you think about eel?

Related to https://github.com/cryptoadvance/specter-desktop/issues/37

ben-kaufman:
The second use case sounds very important, I didn't know it's not working now. I was actually also thinking of another feature which I think would be good to have but apparently would require that too (basically allowing Specter server to have multiple users and accounts - but that's a discussion for a different issue). So in general making it a mini-app sounds great.

- I think if that's a standalone app it deserves a repo of its own :) (though it actually depends on the answer to the next question).
[...]

PR #143, opened 2020-05-29
HWI JSON-RPC server

ben-kaufman:
Put all HWI commands in a single JSON-RPC server. (based on the work @stepansnigirev started in HWIBridge). Also improves HWI error handling and cleans up old code.


src/cryptoadvance/specter/controller.py
```diff
@@ -41,6 +43,22 @@ def inject_debug():
 
 ################ routes ####################
 
+hwi = HWIBridge()
+
+@app.route("/hwi_rpc/", methods=["POST"])
```
k9ert, line 48: What do you think about separating the hwi-rpc-stuff to a separate controller-file? The controller is already 700 lines and we should think about reasonable ways to split it up.

src/cryptoadvance/specter/controller.py
```diff
@@ -595,6 +613,12 @@ def new_device():
     app.specter.check()
     return render_template("device/new_device.jinja", specter=app.specter, rand=rand)
 
+@app.route('/hwi_new_device/')
+@login_required
+def hwi_new_device_xpubs():
```
k9ert, line 618: Or maybe we should rather splitup all hwi (including/excluding) rpc-stuff?

src/cryptoadvance/specter/server.py
```diff
@@ -62,7 +61,6 @@ def login():
     else:
         app.logger.info("Login enabled")
     app.logger.info("Initializing Controller ...")
-    app.register_blueprint(hwi_views, url_prefix='/hwi')
```
k9ert, line 65: The blueprint is a nice feature of flask. Maybe we used it wrong but we should think about how we can leverage it, imho.

stepansnigirev: Blueprint is a nice feature if we need more than one endpoint.
At the moment we use HWI only from javascript and if we make it a single json-rpc endpoint there is no need in a blueprint.
It would make sense to use blueprints for all `wallet/` pages for example, it helps if we have multiple routes with the same prefix.
We should consider blueprints for other things, I just don't think that HWI is the right place to use it.

src/cryptoadvance/specter/hwi_rpc.py
```diff
@@ -0,0 +1,258 @@
+from flask import Flask, url_for, render_template, jsonify, request, make_response, redirect, Response, current_app
```
stepansnigirev, line 1: why do we need this imports here?
I would like to keep `hwi_rpc.py` flask-independent

src/cryptoadvance/specter/hwi_rpc.py
```diff
@@ -0,0 +1,258 @@
...
+import json, subprocess, time, traceback
+from hwilib.serializations import PSBT
+import hwilib.commands as hwi_commands
+from hwilib import bech32
+from .helpers import normalize_xpubs, convert_xpub_prefix
+
+from .specter_hwi import SpecterClient, enumerate as specter_enumerate
+
+class JSONRPC:
```
stepansnigirev, line 10: let's also move `JSONRPC` class to a separate file, we may reuse it later for non-HWI tasks

src/cryptoadvance/specter/hwi_rpc.py
ben-kaufman: I removed it from there. It is still dependent on having access to the Specter instance to get the network and extract keys (add them to the Device Manager). Do you think it's better to separate it too so that the HWI part is completely independent from Specter itself?

src/cryptoadvance/specter/hwi_rpc.py
ben-kaufman: Makes sense, I separated it.

src/cryptoadvance/specter/controller.py
ben-kaufman: I separated it now.

src/cryptoadvance/specter/hwi_rpc.py
ben-kaufman: I think it might also make sense to refactor the extract xpubs so that instead of having a separate screen for that, it'll just be a popup like the rest of HWI functionalities.
So the `Get over USB`  button we currently have will just start the process instead of going to a separate screen. What do you think?

src/cryptoadvance/specter/hwi_server.py
```diff
@@ -6,12 +6,15 @@
...
 rand = random.randint(0, 1e32) # to force style refresh
 
-hwi = HWIBridge()
+hwi = None
 
 @app.route("/hwi_rpc/", methods=["POST"])
 @login_required
 def api():
     """JSON-RPC for ... anything. In this case - HWI Bridge"""
+    global hwi
```
stepansnigirev, line 15: maybe makes sense to create hwi when we create specter?
Like in `Specter.__init__`:
```py
class Specter:
    def __init__(self, data_folder="./data", config={}):
        self.hwi = HWIBridge(self)
        #...
```

Then we can access it as `app.specter.hwi`

ben-kaufman: I was thinking to make them just completely separated like this: https://github.com/cryptoadvance/specter-desktop/pull/143#discussion_r432836087
What do you think?

src/cryptoadvance/specter/hwi_rpc.py
stepansnigirev: Making xpub extraction with a popup is a good idea!
We want to extract xpubs both for testnet and mainnet, so I think it would be better to pass testnet flag or chain parameter with the request.

ben-kaufman:
I made here the new device process be just a single screen (instead of 2 or 3). Now you can just choose the method to extract keys on the same screen for all types.
I also separated here the HWI bridge from Specter logic.

ben-kaufman:
I see now this test fails here: https://github.com/cryptoadvance/specter-desktop/blob/57b98e2d9c533203bb22172271ad92d2b9120262/tests/test_controller.py#L10

I'm not very familiar with the tests here and having some problems running it locally.
Any tips on how to fix that? The easiest thing of course is to remove that check but I don't think that's a good idea.

ben-kaufman:
So now that we basically have a separate server which could do HWI communication, I think at least as a first step for having an independent HWI Bridge as we talked about here #131, we could add an HWIBridge only mode for running Specter. That way you could run the full Specter on a remote machine, and only the HWI part on the local machine and have it work together right away (I think that should work, right?).
What do you think? That's not yet ideal for most users I guess, but it should be a good beginning.

stepansnigirev:
Regarding tests, I don't understand how they work. I opened an issue:
https://github.com/cryptoadvance/specter-desktop/issues/146

Regarding the bridge, let's keep it to the next PR. I hope I can test this one today.

I think it makes sense to add another option on how to run specter, something like:

```sh
python -m cryptoadvance.specter hwibridge
```

One thing to mention - we may need to enable CORS for the HWI bridge. 

Ideally, we should add a whitelist of domains that can access HWI bridge, but let's keep it for later. We would need a separate UI for the HWI bridge then.

ben-kaufman:
Sounds good. So what would the HWI settings need beyond whitelist domains (if anything at all)?

stepansnigirev:
At the moment I can't think of anything else.
Let's start with a single settings page with a list of domains that are allowed to talk to the bridge.

src/cryptoadvance/specter/hwi_rpc.py
```diff
@@ -137,8 +137,12 @@ def _get_client(self, device_type=None, path=None, fingerprint=None, passphrase=
                     client = SpecterClient(device["path"])
                 else:
                     client = hwi_commands.get_client(device_type, path, passphrase)
+                if not client:
+                    raise Exception('The device was identified but could not be reached.  Please check it is properly connected and try again')
                 client.is_testnet = 'test' in chain
```
stepansnigirev, line 142: 'signet' is also testnet, but doesn't have 'test' in chain.
The only mainnet is 'main', so I think is_testnet should be if `chain == 'main'`

ben-kaufman: Right, fixed

stepansnigirev:
Makes sense to allow connection from the same domain by default.
At the moment if I just run specter without configuring HWIbridge it returns an error: 
```
HWI-Bridge call returned with an error: Unauthorized request origin.
You must first whitelist this website URL in HWIBridge settings to grant it access.
```
This shouldn't happen if I use hwi bridge from the same domain.

stepansnigirev:
Ah, sorry, it's because I use from `http://localhost:25441`, not `127.0.0.1`.
So makes sense to add `localhost` to the whitelist together with `127.0.0.1` by default

src/cryptoadvance/specter/helpers.py
```diff
@@ -262,3 +262,25 @@ def get_loglevel(app):
         logging.DEBUG : "DEBUG"
     }
     return loglevels[app.logger.getEffectiveLevel()]
+
+def hwi_get_config(specter):
+    config = {
+        'whitelisted_domains': 'http://127.0.0.1:25441/'
```
stepansnigirev, line 268: Trailing space is breaking access for some reason.

src/cryptoadvance/specter/helpers.py
```diff
@@ -262,3 +262,25 @@ def get_loglevel(app):
...
+            deep_update(config, file_config)
+    # otherwise - create one and assign unique id
+    else:
+        save_hwi_bridge_config(specter, config)
+    return config
+
+def save_hwi_bridge_config(specter, config):
+    if 'whitelisted_domains' in config and not config['whitelisted_domains'].endswith("/"):
+            # make sure the url end with a "/"
+             config['whitelisted_domains'] += "/"
```
stepansnigirev, line 284: Doesn't work for multiple domains, trailing space is added only to the last one.
But as I mentioned before, trailing space is breaking access to hwo

src/cryptoadvance/specter/templates/hwi_bridge.jinja
```diff
@@ -0,0 +1,24 @@
...
+
+{% block sidebar %}
+{% endblock %}
+
+{% block main %}
+    <div class="card">
+        <form action="." method="POST" role="form">
+            <h1>HWI Bridge Settings</h1>
+            <label>Whitelisted domains</label>
+            <input class="form-control" placeholder="e.i. http://127.0.0.1:25441" name="whitelisted_domains" type="text" value="{{ whitelisted_domains }}">
```
stepansnigirev, line 12: I suggest changing it to the textarea and separating domains with a new line, not with comma

ben-kaufman:
I fixed all the comments. Also made sure localhost and 127.0.0.1 are interchangeable

stepansnigirev:
Looks good to me! Merging.

PR #144, opened 2020-05-29
Replace default block explorer with copy functionality

ben-kaufman:
Currently, Specter has by default blockstream.info as a block explorer to open transactions and addresses in when the user clicks on one.
This might be useful for some users, but it seems like for most it is more confusing than helpful.
Here I make the default block explorer empty, and change the behavior of Specter to copy the address/ transaction id to the user's clipboard whenever the block explorer setting is empty, instead of opening an invalid URL.
In short, the default behavior now would be to copy address/ transaction id to user clipboard instead of opening a block explorer.
Also added a small note to explain that the block explorer setting is not used to fetch any data, to avoid possible confusion.


src/cryptoadvance/specter/logic.py
```diff
@@ -92,10 +92,10 @@ def __init__(self, data_folder="./data", config={}):
             },
             "auth": "none",
             "explorers": {
-                "main": "https://blockstream.info/",
-                "test": "https://blockstream.info/testnet/",
+                "main": "",
+                "test": "",
                 "regtest": None,
```
stepansnigirev, line 97: Maybe for "regtest" should be an empty string as well?

ben-kaufman:
Right, fixed.

PR #145, opened 2020-05-30
preparing release 0.4.0

k9ert:
* release_notes


ben-kaufman:
It might be worth mentioning #140 as it's a bug fix which someone using an older version might face. There was also #128 which fixed an issue with showing small amounts and also allowed using `sats` instead of `BTC`.

k9ert:
I'll try to list all changes in the future, not only features. It makes sense for multiple reasons and it's probably not that much more work.

PR #152, opened 2020-06-05
fixes arm-hwi-dependency issue see hwi bug 331

k9ert:
This is the first time we want to fix something in an older tag. Here is a suggestion of how to do that. I messed already up as i pushed the changed branch upstream. I should have done that BEFORE i make any changes. Anyway, here is a proposal for the ideal flow:
```
git checkout v0.4.0
git checkout -b branches/v0.4
git push upstream
#fix the issue, git commit
git push --set-upstream origin branches/v0.4 # origin is my fork
# create PR to master (this is this one)
# create a PR from origin-branch to upstream branch
# after that's merged, do a release
```
Maybe i'm overcomplicating this.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ ecdsa>=0.13.3
 Flask==1.1.1
 Flask-Login==0.5.0
 hidapi==0.7.99.post21
-hwi==1.1.0
+hwi==1.1.1
 idna==2.8
 itsdangerous==1.1.0
 Jinja2==2.10.3


k9ert:
Interesting. I never designed a reasonable git-process for maintaining branches. But obviously a PR from that branch to master doesn't work as i don't want to "Merge branch 'master' into branches/v0.4" which happened when i clicked on "update branch" (should have thought about that for some more seconds).

k9ert:
Closing. We need to do that differently.

PR #158, opened 2020-06-07
Showing the version with the settings-screen

k9ert:
Looks like this:

image

### DIFF
--- a/src/cryptoadvance/specter/controller.py
+++ b/src/cryptoadvance/specter/controller.py
@@ -13,7 +13,7 @@ from flask_login import login_required, login_user, logout_user, current_user
 from flask_login.config import EXEMPT_METHODS
 
 
-from .helpers import normalize_xpubs, run_shell, set_loglevel, get_loglevel
+from .helpers import normalize_xpubs, run_shell, set_loglevel, get_loglevel, get_version_info
 from .descriptor import AddChecksum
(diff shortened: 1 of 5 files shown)


src/cryptoadvance/specter/controller.py
```diff
@@ -141,6 +141,11 @@ def settings():
     protocol = 'http'
     explorer = app.specter.explorer
     auth = app.specter.config["auth"]
+    version_info={}
+    version_info["current"], version_info["latest"], version_info["upgrade"] = get_version_info()
+    app.logger.info("Upgrade? {}".format(version_info["upgrade"]))
+    if version_info["upgrade"]:
+        flash("There is a new version available. Consider strongly to upgrade to the new version {}".format(version_info["latest"]))
```
stepansnigirev, line 148: Maybe makes sense to add a hint how to update, like:
`Update to {version_info["latest"]} with "pip3 install {name} --upgrade"`
and get the package `name` from `get_version_info()` as well

k9ert: Not sure why the package_name should get returned. It's cryptoadvance.specter, right?
Fiex the other stuff.

k9ert:
I'm not sure why "some checks haven't completed yet" as they clearly did.

PR #160, opened 2020-06-09
Refactor logic + cleanup + bug fixes

ben-kaufman:
- Breaks down `logic.py` into a file for each of its classes.
- Remove the wallet and device managers iterability, replace it with a regular dict property.
- Some code cleanup and refactoring.
- Add more testing for device_manager, device, and wallet_manager (more Specter and Wallet tests soon hopefully).
- Fix bug when saving changes for settings (https://github.com/cryptoadvance/specter-desktop/pull/160/files#diff-f3f35519e1706a65feabaffbf0076002L257).
- Fix Bitcoin Core 0.20.0 address labeling compatibility issue.


ben-kaufman:
@stepansnigirev thanks for testing, I've encountered it while continuing the refactoring too, I think now it should be fine.
Not sure if we should merge this already so it doesn't get too big, or wait to complete the refactoring of wallets and devices here as a single PR, so whatever you think is better.

stepansnigirev:
I think let's continue refactoring here. It would be nice to finish splitting device and wallet classes in this PR.

k9ert:
LGTM. I was a bit concerned that renaming logic.py back to specter.py (which caused trouble in the past) would again cause trouble, but that doesn't seem to be the case.

src/cryptoadvance/specter/devices/specter.py
```diff
@@ -0,0 +1,40 @@
...
+        psbts['qrcode'] = qr_psbt.serialize()
+        return psbts
+
+    def export_wallet(self, wallet):
+        return wallet.name + "&" + get_wallet_qr_descriptor(wallet)
+
+def get_wallet_qr_descriptor(wallet):
+        return wallet.recv_descriptor.split("#")[0].replace("/0/*", "")
+
+@property
```
stepansnigirev, line 35: @property shouldn't be here
It breaks creating psbt for specter

src/cryptoadvance/specter/wallet.py
```diff
@@ -0,0 +1,510 @@
...
+        return address
+
+    def get_address(self, index, change=False):
+        pool = self.change_keypool if change else self.keypool
+        if pool < index + WALLET_CHUNK:
+            self.keypoolrefill(pool, index + WALLET_CHUNK, change=change)
+        desc = self.change_descriptor if change else self.recv_descriptor
+        if self.is_multisig:
+            desc = sort_descriptor(self.cli, desc, index=index, change=change)
+            return self.cli.deriveaddresses(desc)[0]
```
stepansnigirev, line 294: `sort_descriptor` doesn't work correctly if you provide "sortedmulti" there.
In v20 it causes invalid ordering of pubkeys, in 19 - rpc error.
And sorting descriptor is not necessary for v20 so I suggest the following:
```py
    if self.is_multisig:
            # first try with sortedmulti
            try:
                addr = self.cli.deriveaddresses(desc)[0]
            except:
                # if sortedmulti is not supported
                desc = desc.replace("sortedmulti", "multi")
                desc = sort_descriptor(self.cli, desc, index=index, change=change)
                addr = self.cli.deriveaddresses(desc)[0]
            return addr
```

src/cryptoadvance/specter/helpers.py
```diff
@@ -307,3 +204,72 @@ def save_hwi_bridge_config(specter, config):
...
+    if app.specter.chain == "main":
+        prefix = "xpub"
+    for cosigner in cosigners:
+        device = copy.deepcopy(cosigner)
+        allowed_types = ['', wallet_type]
+        device.keys = [key for key in device.keys if key.xpub.startswith(prefix) and key.key_type in allowed_types]
+        devices.append(device)
+    return devices
+
+def sort_descriptor(cli, descriptor, index=None, change=False):
```
stepansnigirev, line 238: 1. 8 spaces padding is wrong
2. `descriptor = descriptor.replace('sortedmulti', 'multi')` should be moved to the very beginning of the function, otherwise it will fail in `deriveaddress` if core v19 is used, and will get invalid ordering if v20 is used

src/cryptoadvance/specter/wallet_manager.py
```diff
@@ -0,0 +1,170 @@
...
+            wallet_alias = alias("%s %d" % (name, i))
+            i += 1
+
+        arr = key_type.split("-")
+        descs = [key.metadata['combined'] for key in keys]
+        recv_descs = ["%s/0/*" % desc for desc in descs]
+        change_descs = ["%s/1/*" % desc for desc in descs]
+        if len(keys) > 1:
+            recv_descriptor = "multi({},{})".format(sigs_required, ",".join(recv_descs))
+            change_descriptor = "multi({},{})".format(sigs_required, ",".join(change_descs))
```
stepansnigirev, line 115: let's do `sortedmulti` here from the start

src/cryptoadvance/specter/wallet.py
```diff
@@ -0,0 +1,510 @@
...
+        ]
+        r = self.cli.importmulti(args, {"rescan": False}, timeout=120)
+        # bip67 requires sorted public keys for multisig addresses
+        if self.is_multisig:
+            # we do one at a time
+            args[0].pop("range")
+            for i in range(start, end):
+                sorted_desc = sort_descriptor(self.cli, desc, index=i, change=change)
+                args[0]["desc"] = sorted_desc
+                self.cli.importmulti(args, {"rescan": False}, timeout=120)
```
stepansnigirev, line 327: Here what I came up with:
```py
        if not self.is_multisig:
            r = self.cli.importmulti(args, {"rescan": False}, timeout=120)
        # bip67 requires sorted public keys for multisig addresses
        else:
            # try if sortedmulti is supported
            try:
                r = self.cli.importmulti(args, {"rescan": False}, timeout=120)
            except:
                # first import normal multi
                desc = desc.replace("sortedmulti", "multi")
                r = self.cli.importmulti(args, {"rescan": False}, timeout=120)
                # make a huge batch or single addresses
                arg = args[0]
                arg.pop("range")
                batch = []
                for i in range(start, end):
                    sorted_desc = sort_descriptor(self.cli, desc, index=i, change=change)
                    obj = {}
                    obj.update(arg)
                    obj.update({"desc": sorted_desc})
                    batch.append(obj)
                r = self.cli.importmulti(batch, {"rescan": False}, timeout=120)
```

PR #161, opened 2020-06-14
proper pinning of bitcoind-version in different branches

k9ert:
We might want to support different bitcoind-versions on different branches in the future. This change prepares this.
The used version of bitcoin is from now on no longer "latest" (or master) but specified in pytest.ini.
This version will be expected to be existing as a docker-container or as a tag on github.com/bitcoin/bitcoin


stepansnigirev:
Why is it red?

k9ert:
Now it's green. There was an unnecessary git pull somehere in the code. Whether this successfully works will need to have a branch as well. After merge, i would do a test with a fresh branch.

In fact, we should probably change that so that all tests are run twice: Once on v0.19 and once on v0.20.

PR #162, opened 2020-06-14
Dependencies

k9ert:
I've discovered a bot called https://dependabot.com/ which is pretty amazing. It checks requirements.txt for potential updates and creates PRs for each of them.
unsurprisingly the bot needs write-access to your repo as the branches are created on directly in your repo. As i didn't wanted that for cryptoadvance/specter, i agreed on that on k9ert/specter and i'm now accumulating the dependency-upgrades and creating a manual PR for a couple of them.


stepansnigirev:
Talking about dependencies...
Why do we have sub dependencies in the list as well?
Jinja is coming from Flask, so why do we keep it there?
The same with USB stuff - Specter only uses HWI, so why there is hidapi and libusb?
Maybe we should remove all dependencies that we don't use directly in the project?

k9ert:
Very good point. I'll work on them on this PR.

k9ert:
I guess this is now ready to get merged. 1/3 of all dependencies were unnecessary.
When this is done, i'll have a look into a dependabot workflow again.

PR #165, opened 2020-06-17
Faq toc autogeneration

k9ert:
This uses a Github-Action to autogenerate Table of contents for markdown-files, first of all the faq.md.

Should we keep the TOC for the readme as well? Or should we have a TOC for all md-files?

### DIFF
--- a/.github/workflows/toc.yml
+++ b/.github/workflows/toc.yml
@@ -1,0 +1,8 @@
+on: push
+name: TOC Generator
+jobs:
+  generateTOC:
+    name: TOC Generator
+    runs-on: ubuntu-latest
+    steps:
+      - uses: technote-space/toc-generator@v2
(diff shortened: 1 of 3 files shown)


stepansnigirev:
I was thinking about using [docsify](https://docsify.js.org/#/) - I tried it on the documentation of my c++ bitcoin library, looks pretty nice: https://micro-bitcoin.github.io/

It takes markdown files and renders them as a nice webpage, only a few files are requred - `index.html` and `_sidebar.md` to configure sidebars. I think we can do something like this for documentation for both specter-desktop and specter-diy.

If we use docsify we don't need TOC because it is automatically generated in the sidebar.

k9ert:
Would agree if we would have a webpage. But we don't ... yet.
This is a quite simple solution and we can get rid of it quite easily. Or do you think it's possible to get a page up and running soonish?

stepansnigirev:
I agree. Let's do TOC for now, when I find time for a webpage we will change it.

PR #171, opened 2020-06-23
Rename Pending -> Unsigned PSBTs

### LINKED ISSUE #168: confirmed tx stays "locked in pending"
I have received a tx that took some time to confirm. Now it has been confirmed about 200 blocks deep (specter says so), but I still cannot send funds:  "- BTC, Available: 0;  Locked in pending transactions: xxxx".

ben-kaufman:
Fix #168 

Rephrases "Pending" to "Unsigned" in the Pending PSBTs screen + a small note to explain what the screen is.


ben-kaufman:
Hmmm seems like Travis is doing problems again :(
(though for this PR it shouldn't really matter, and anyway it passed on my fork).

PR #172, opened 2020-06-25
Multi-user support

ben-kaufman:
Allow multiple users to use the same Specter server.
For now, each user is separated from the rest.
The original user (the admin) can send registration links with OTP for others to register.
The settings page and configuration was split to admin managed settings like `auth` and user-specific ones like `hwi_bridge_url`.


src/cryptoadvance/specter/controller.py
```diff
@@ -99,42 +100,93 @@ def index():
...
     logout_user()
     flash('You were logged out',"info")
+    app.specter.clear_user_session()
     return redirect("/login") 
 
 @app.route('/settings/', methods=['GET', 'POST'])
 @login_required
 def settings():
     current_version = notify_upgrade()
+    print(current_user)
```
stepansnigirev, line 189: print ;)

ben-kaufman: Opps, right, removed.

stepansnigirev:
When I change in settings to Multiple Users I get an error:
`AttributeError: 'AnonymousUserMixin' object has no attribute 'is_admin'`

Where should I set up the initial admin user?

ben-kaufman:
Strange, I can't reproduce it... Though I think I saw a similar error and fixed it with the last commit (0f73225), are you using it?

Right now for simplicity and compatibility with the current RPC password as pin login option, I made the admin be the username `admin` and password as the RPC password. I didn't write it there because I think I should change it, but I'm not yet sure what to change it with, maybe it's good enough like that...

stepansnigirev:
That might be a problem - I am using cookies in Core, so there is no static username and password in my case.

stepansnigirev:
Maybe if multi user option is selected we could show two more inputs in the settings - admin username and password

ben-kaufman:
Makes sense. I added now in the settings that any user can change its username and password, and noted that for the admin the default is admin for both.

src/cryptoadvance/specter/controller.py
```diff
@@ -143,22 +192,31 @@ def settings():
...
         action = request.form['action']
+        explorer = request.form['explorer']
+        hwi_bridge_url = request.form['hwi_bridge_url']
+        if 'specter_username' in request.form:
+                specter_username = request.form['specter_username']
+                specter_password = request.form['specter_password']
+        else:
+            specter_username = None
+            specter_password = None
+        if current_user.is_admin:
```
stepansnigirev, line 212: this line fails if I am changing the authenticatoin from "none" to multi-user

src/cryptoadvance/specter/controller.py
```diff
@@ -173,26 +231,56 @@ def settings():
...
+                            new_otp=new_otp,
+                            loglevel=loglevel,
+                            specter=app.specter,
+                            current_version=current_version,
+                            rand=rand)
+                current_user.username = specter_username
+                if specter_password:
+                    current_user.password = hash_password(specter_password)
+                current_user.save_info(app.specter)
+            if current_user.is_admin:
```
stepansnigirev, line 258: this line fails if I am changing the authenticatoin from "none" to multi-user

ben-kaufman:
I already did that ([here](https://github.com/cryptoadvance/specter-desktop/pull/172/files/31b15715423e7ff937ce27fb85d51bfcd3f6ec77#diff-a386d70e20c16c989ef22377f5ca3398R105) and [here](https://github.com/cryptoadvance/specter-desktop/pull/172/files/31b15715423e7ff937ce27fb85d51bfcd3f6ec77#diff-a386d70e20c16c989ef22377f5ca3398R117)) but just noticed I missed it in one case, added it now too so it should be fine.

stepansnigirev:
I was able to get to the multi-user when I first switched from None to rpcpassword auth mechanism, and then to multiuser.

From None to multiuser still doesn't work.

How to reproduce:
- Set auth to None
- Enter incognito mode
- Try changing auth to multiuser

src/cryptoadvance/specter/controller.py
```diff
@@ -143,22 +193,31 @@ def settings():
     host = rpc['host']
     protocol = 'http'
     explorer = app.specter.explorer
-    auth = app.specter.config["auth"]
-    hwi_bridge_url = app.specter.config['hwi_bridge_url']
+    auth = app.specter.config['auth']
```
stepansnigirev, line 196: Maybe add here:
```py
    if auth=='none':
        app.login('admin')
```
Worked for me...

ben-kaufman: Oh I see why that's needed, thanks, added.

stepansnigirev:
Now works. Just tried registering a new user - works fine!

I see that the user has full access to devices that he doesn't own, in particular he can delete the device, keys of the device, or even add new keys to the device of another user.

I think we should at least disable the editing capabilities of the devices that the user doesn't own.
The next step would be to configure what keys are visible to other users.

ben-kaufman:
That’s weird, each users should be able to access only his own devices, that’s how it is for me, so it’s probably a bug, I’ll check why it’s happening.

Later I want to add sharing and visibility to devices and wallets, but that’s more complicated.

stepansnigirev:
When I am creating a device I don't see any ownership fields in the device json file, so I am not sure how you determine who owns the device...

ben-kaufman:
I create a separate devices folder for each user. This is why I’m surprised you see one users devices when logged to another.
When adding sharing I will likely need to give up the separation by folders, but for a first stage doing so made it easiest.

src/cryptoadvance/specter/specter.py
```diff
@@ -100,26 +102,36 @@ def check(self):
...
-                self.cli, 
-                chain=chain
-            )
+            user_folder_id = ''
+
+        if self.config['auth'] != 'usernamepassword' or (user and not user.is_anonymous):
+            if self.device_manager is None:
+                self.device_manager = DeviceManager(os.path.join(self.data_folder, "devices{}".format(user_folder_id)))
+            else:
+                self.device_manager.update()
```
stepansnigirev, line 114: Probably you want to pass user path here as well, similar to wallet manager.

ben-kaufman: Added. For me it works well, can you please verify now there's no persistence issue for you too?

stepansnigirev: Works now!

src/cryptoadvance/specter/helpers.py
```diff
@@ -273,3 +297,22 @@ def sort_descriptor(cli, descriptor, index=None, change=False):
         desc = f"{p}({desc})"
 
     return AddChecksum(desc)
+
+def hash_password(password):
+    """Hash a password for storing."""
+    salt = hashlib.sha256(os.urandom(60)).hexdigest().encode('ascii')
+    pwdhash = hashlib.pbkdf2_hmac('sha512', password.encode('utf-8'), 
+                                salt, 10000)
```
stepansnigirev, line 305: I suggest making a few changes here:
- store salt and password hash separately to avoid things like `stored_password[64:]`, so password key would contain `salt` and `pwdhash` keys.
- use `pbkdf2_hmac('sha256',...)` to make resulting hash x2 shorter, I don't see security benefits of sha512 here.
- store bytes not as hex digits but use base64 encoding (`binascii.b2a_base64` and `binascii.a2b_base64`) - again, shorter.

One note about `b2a_base64` function - it adds `\n` at the end of the string for some reason, but it can be removed using `.strip()` on it. It doesn't influence the result of `a2b_base64` but I am not sure how json will react to `\n` in the string

src/cryptoadvance/specter/templates/settings.jinja
```diff
@@ -3,35 +3,55 @@
...
+				{% if specter.config['auth'] == 'usernamepassword' %}
+					<button type="submit" class="btn" style="width: 100%;" name="action" value="adduser">Generate New Registration Link</button>
+					<br>
+					<span class="note">Default is username: admin, password: admin.</span><br>
+				{% endif %}
+			{% endif %}
+			{% if specter.config['auth'] == 'usernamepassword' %}
+				Specter Username:<br><input type="text" name="specter_username" type="text" value="{{ current_user.username }}"><br><br>
+				Specter Password:<br><input type="password" name="specter_password" type="text" placeholder="Set new password"><br>	
+				<br><br>
```
stepansnigirev, line 45: Let's add this part dynamically with javascript when the authentication method is changed. 
UX will be more intuitive in this case - the user selects `usernamepassword` setting and can set up admin login and password right away.
When authentication method is changed to `rpcpasswordaspin` or `none` - we remove this part of the form.

ben-kaufman:
Updated based on both comments :)

PR #176, opened 2020-06-27
Separate setttings into tabs

ben-kaufman:
Since the Settings page has become quite big and will get bigger in the near future, this PR splits it to General, Auth, and Bitcoin Core settings pages, with tabs like we use for the wallet.


stepansnigirev:
Nice, much cleaner now! A few comments...

1. I think the most important is Bitcoin Core configuration, it should be the first one and open by default when you click on the settings.
2. Let's rename Auth to Authentication, I think we have enough space for the full word :)
3. I also think it would be nice to have a link to the HWI bridge settings in the General tab or even iframe it there.

It might make sense to redirect from `/` to `/hwi/settings` if specter runs in hwibridge mode, because otherwise it's not clear where to go... But maybe in another PR?

stepansnigirev:
Weird behaviour happens when you choose multiuser mode and instead of saving it you click on the OTP link button.
I think OTP button should appear only when you save the authentication.

And maybe it's better not to redirect to the main page from the settings any more, not sure...

ben-kaufman:
Very strange behavior of the controller test suddenly. When the same code is inside a single test it seems to pass, but when they are in 2 tests I get a 404 suddenly, despite each one separately passing. For now, I just added them together into a single test.

stepansnigirev:
Might be because tests run in random order, so at the end of the test we need to make sure we are logged in.
At the moment if login logout runs first we are left logged out from the last invalid credentials test.
Can you try moving a valid login test at the end?

ben-kaufman:
This doesn't seem to help :( The problem is that I just get a `404` response error, regardless of which URL I'm trying to query in the second test, as if the client has become inaccessible.

PR #179, opened 2020-06-30
Support mulisig display address

### LINKED ISSUE #79: Enhancement: add "verify address on the device" button

ben-kaufman:
Fix #79


stepansnigirev:
Tested, works!

PR #189, opened 2020-07-04
Cobo Vault support

stepansnigirev:
New device - [Cobo Vault](https://cobo.com/hardware-wallet/cobo-vault).
For now single-sig only as they don't support multisig yet.

[bcur standard](https://github.com/BlockchainCommons/Research/blob/master/papers/bcr-2020-005-ur.md) is used for communication, but it may change and evolve over time.


llxado123:
Thanks Stepan!

PR #194, opened 2020-07-05
Manage users panel

ben-kaufman:
Adds a list of existing users and the option to delete each user to the authentication settings page, visible only to the admin.


stepansnigirev:
Breaks when I try to delete the user. `auth` key is not in `request.form`.

These lines need to be moved to `if action == "save":`:

```py
if current_user.is_admin:
    auth = request.form['auth']
```

https://github.com/cryptoadvance/specter-desktop/pull/194/files#diff-a386d70e20c16c989ef22377f5ca3398L334-L335

ben-kaufman:
Ok, I moved that, should be good now.

PR #201, opened 2020-07-08
Allow changing device type

### LINKED ISSUE #199: cant import a wallet
## Abstract

* idk a bunch of problems for me, meybe i'm missing something
* did a fresh clone, installation, before that deleted .specter, registered new hwws, created a multisig wallet, but no qr code shown
* device.exportable_to_wallet evaulates to false
* ubuntu, firefox

image

ben-kaufman:
Close #199 
- Allows the user to change the device type from the device screen.
- Show a warning to the user when adding a device if the type submitted is "Other".


src/cryptoadvance/specter/templates/device/device.jinja
```diff
@@ -1,6 +1,11 @@
 {% extends "base.jinja" %}
 {% block main %}
-	<h1>Public keys of {{ device.name }}</h1>
+	<h1>{{ device.name }}</h1>
+	<form action="./" method="POST">
+		{% include "device/components/device_type.jinja" %}
+		<button type="submit" class="btn hidden" id="settype" name="action" value="settype" style="float: right;">Update</button>
+	</form>
```
stepansnigirev, line 7: "Update" button is not aligned with the rest of the elements.

What do you think about doing it like this?
```html
<form action="./" method="POST">
	<div class="row">
		{% include "device/components/device_type.jinja" %}
		<button type="submit" class="btn hidden" id="settype" name="action" value="settype" style="height:15px; align-self:center;">Update</button>
	</div>
</form>
```

ben-kaufman: Oh I didn't notice that as I use Safari, on which it looks the same as with the fix, I should remember to test on Chrome too...

stepansnigirev:
And I normally use Chrome and often forget to check in Safari, Kim is using Firefox, so we complement each other :)
We need one more contributor who uses Edge, but I think it may be problematic :D

Issue #203, opened 2020-07-10
Ledger - HWI-Bridge call returned with an error: Internal error: open failed


stefanwouldgo:
When signing a multisig tx with Specter 0.5.2:

`[2020-07-10 10:28:19,276] INFO in _internal: 127.0.0.1 - - [10/Jul/2020 10:28:19] "POST /hwi/api/ HTTP/1.1" 200 -
Traceback (most recent call last):
  File "/home/stefan/.local/lib/python3.8/site-packages/cryptoadvance/specter/json_rpc.py", line 36, in jsonrpc
    response["result"] = method(**request["params"]) # dict -> **kwargs
  File "/home/stefan/.local/lib/python3.8/site-packages/cryptoadvance/specter/hwi_rpc.py", line 108, in sign_tx
    client = self._get_client(device_type=device_type, fingerprint=fingerprint, path=path, passphrase=passphrase, chain=chain)
  File "/home/stefan/.local/lib/python3.8/site-packages/cryptoadvance/specter/hwi_rpc.py", line 139, in _get_client
    client = hwi_commands.get_client(device_type, path, passphrase)
  File "/home/stefan/.local/lib/python3.8/site-packages/hwilib/commands.py", line 34, in get_client
    client = client_constructor(device_path, password, expert)
  File "/home/stefan/.local/lib/python3.8/site-packages/hwilib/devices/ledger.py", line 104, in __init__
    device.open_path(path.encode())
  File "hid.pyx", line 72, in hid.device.open_path
OSError: open failed
`
This is on Ledger Nano S secure element 1.6, microcontroller 1.12, bitcoin app 1.4.2

stepansnigirev:
I made a PR that will hopefully fix this issue - #205 
Could you check if the issue remains with this PR?

stepansnigirev:
I will reopen until @stefanwouldgo could confirm that the issue is solved.

stefanwouldgo:
Just tried Specter 0.5.3. Unfortunately, the issue isn't quite solved. The error described above doesn't appear anymore, but I still cannot sign the tx. In fact, the behavior now is even stranger. First, ledger got into look showing me the same tx to approve again and again. After some time, I got a few errors:
`[2020-07-11 18:28:34,253] INFO in _internal: 127.0.0.1 - - [11/Jul/2020 18:28:34] "POST /hwi/api/ HTTP/1.1" 200 -
Traceback (most recent call last):
[...]

PR #205, opened 2020-07-10
thread locks and daemon

stepansnigirev:
Flask is running in multithreading mode, this causes problems with file access and device access as well.
This PR adds thread locks to make sure two threads don't interfere if they want to use the same resource.

Maybe close https://github.com/cryptoadvance/specter-desktop/issues/203


src/cryptoadvance/specter/wallet.py
```diff
@@ -192,8 +192,9 @@ def json(self):
         }
 
     def save_to_file(self):
-        with open(self.fullpath, "w+") as f:
-            f.write(json.dumps(self.json, indent=4))
+        with self.lock:
+            with open(self.fullpath, "w+") as f:
+                f.write(json.dumps(self.json, indent=4))
```
ben-kaufman, line 197: Isn't that also needed in the device class fo the `_update_keys` and `set_type` functions?

stepansnigirev: Thanks! Updated!

stepansnigirev:
I returned back `Daemonize` as it was only causing issues on MacOS with python3.7.
So instead it warns the user that better to upgrade to python3.8.

PR #210, opened 2020-07-11
Bitcoin Core hot wallet

### LINKED ISSUE #58: New device type - this computer
I am thinking about including a hot storage functionality into the app by implementing a new device type - call it a "Hot storage" or "This computer".
### Motivation:
I really like using Specter-desktop for my cold storage, but for small transactions I don't want to use my hardware wallet — my laptop is always with me, but I don't take my hardware wallets with me all the time.
At the moment I have to use another app for these small transactions without a hardware wallet - I would prefer to use Specter-desktop for this as well.
### Future bonuses
If we have hot storage functionality we could later add CoinJoin and Lightning to the app.
Implementing hot storage as a device type also allows us to add the computer to the multisig setup. Or maybe even run Specter on an airgapped computer or raspberry pi purely for signing.

I think this signing functionality can be implemented as a new miniapp that can be included in Specter as an addon. But I don't know yet what would be the best architecture for a plugin system.
In principle future extensions like CoinJoin and Lightning can be also implemented as plugins.

@k9ert , @kdmukai , what do you think? Makes sense or totally insane?

ben-kaufman:
Closes #58 

Adds a new device type: Bitcoin Core. This will create a Bitcoin Core hot wallet from a BIP39 mnemonic as a new signing device, allowing to use it in as any other device (exception is that it is limited to the network type, i.e. the same wallet can't be used both on mainnet and testnet).

To utilize this new device type, go to the new device screen and select Bitcoin Core as the device type. This will open a text field with a proposed, random mnemonic, which you can either use or replace with your own mnemonic. Once created, you can use it just like any other device.

This is still WIP, things left to do:
- [x] Encrypted wallet support (either allow only encrypted for security or just do it by default)
- [x] UI/UX improvements (open for suggestions).
- [x] Foolproof it and proper error handling (e.g. error if invalid mnemonic entered etc.).
- [x] Clean up the code a bit.

Edit:
Added optional support for encrypted wallets (an empty passphrase is treated as unencrypted). Also a few improvements to the UI according to @stepansnigirev's [comment](https://github.com/cryptoadvance/specter-desktop/pull/210#issuecomment-657193254).


stepansnigirev:
Tested, works!
A few comments:
- I would suggest generating a 12-word mnemonic by default, not 24 words. 12 words are unfeasible to brute force as well. Or maybe have a button to regenerate mnemonic with 12 / 24 words.
- Do you plan to implement bip39 passwords? It could be just another input field.
- UX: We can have two tabs for mnemonic - new and recovery. In the new tab we can display mnemonic in a nice table, in the recovery tab we can keep just a text area.

ChristopherA:
You should talk with us (@fonta1n3 @wolfmcnally and I) about how our standards for exchanging seeds, BIP39, Shamir, and derived keys (soon), and account maps (aka public key only wallet descriptors) for Bitcoin Standup & Fully Noded 2. It would be great if this Bitcoin-Core tool used the same standards.

stepansnigirev:
> You should talk with us (@Fonta1n3 @wolfmcnally and I) about how our standards for exchanging seeds, BIP39, Shamir, and derived keys (soon), and account maps (aka public key only wallet descriptors) for Bitcoin Standup & Fully Noded 2. It would be great if this Bitcoin-Core tool used the same standards.

I opened an issue on this topic. Would be nice to support export to Fully Noded and other wallets.

Fonta1n3:
> Closes #58 
> 
> 
> 
> Adds a new device type: Bitcoin Core. This will create a Bitcoin Core hot wallet from a BIP39 mnemonic as a new signing device, allowing to use it in as any other device (exception is that it is limited to the network type, i.e. the same wallet can't be used both on mainnet and testnet).
> 
> 
> 
> To utilize this new device type, go to the new device screen and select Bitcoin Core as the device type. This will open a text field with a proposed, random mnemonic, which you can either use or replace with your own mnemonic. Once created, you can use it just like any other device.
> 
> 
> 
> This is still WIP, things left to do:
> 
> - [ ] Use `[secrets](https://docs.python.org/3/library/secrets.html)` to generate the mnemonic.
> 
> - [ ] Encrypted wallet support (either allow only encrypted for security or just do it by default)
> 
> - [ ] UI/UX improvements (open for suggestions).
> 
> - [ ] Foolproof it and proper error handling (e.g. error if invalid mnemonic entered etc.).
> 
> - [ ] Clean up the code a bit.

Is this a single sig or a signer in a multisig wallet? 

Fwiw I added these instructions to easily recover any Specter wallet with Fully Noded [here](https://github.com/Fonta1n3/FullyNoded/blob/master/Readme.md#specter-pairing), certainly plan to make it easier by supporting your QR code directly. After importing the descriptors a user can simply add signers to make it spendable in the form of bip39 mnemonics.

stepansnigirev:
> Is this a single sig or a signer in a multisig wallet?

It's both, it creates a "device" that can be used as a co-signer in multisig or in a single-sig wallet.
Keys are imported to the Bitcoin Core itself, so they are not stored in Specter.

ben-kaufman:
> Is this a single sig or a signer in a multisig wallet?
The way I implemented this is that I separated the signing functionality from the key management. For the key and address management, we just generate the pubkeys from the mnemonic and save it in the same format we do for cold storage devices (like hww). So like now you can add a hardware wallet by importing its pubkeys, here you'll have a similar experience after adding the mnemonic.
For signing, we create a dedicated wallet which manages the private keys, but is not actually used to track balances or anything, but only as a "signing device", i.e. get a PSBT, sign it, and return the result.
This means you can use the hot wallet both as a single-sig and in a multisig construction like you can now with any other device.

Another recent change we made actually made exporting the wallet depend on the device type you use (whatever it supports importing a wallet or not, and its specific format), which means the exporting QRcode you wrote about will not appear in the wallet settings page by default. I didn't have in mind the case of importing the wallet into another wallet software like FN, but I think it is actually an important one. So it will be very good to add a standard exporting method, which will be available in all wallets, regardless of their devices. Adding that will also mean we can export it exactly as FN would expect it, so the user won't need to do the modifications like described in the doc now. Let's continue talking about this in #212 or a new dedicated issue for that.

src/cryptoadvance/specter/templates/device/new_device.jinja
```diff
@@ -7,77 +7,185 @@
...
+				<label>Passphrase: </label>
+				<input type="password" name="passphrase" class="inline" placeholder="Wallet passphrase">
+				<br><br>
+				<button type="submit" class="btn centered" name="action" value="newhotdevice">Continue</button>
+				<br>
+				<p class="note" style="border-radius: 5px; padding:20px; background: rgba(0,0,0,0.1); color: #fff;">
+				<b style="">⚠️ A note on hot wallets:</b><br>
+				This will create a <a style="color: #fff;" href="https://en.bitcoin.it/wiki/Hot_wallet" target="_blank">hot wallet</a> on your Bitcoin Core node.<br>
+				Hot wallets are considered less secure and are not recommended for use with significant amounts. Use with caution and beware the potential risks.
+				</p>
```
stepansnigirev, line 120: Let's move it to the top, right after device type. Otherwise people will not read this warning.

src/cryptoadvance/specter/controller.py
```diff
@@ -819,23 +847,43 @@ def new_device():
...
+                err = "Failed to parse these xpubs:\n" + "\n".join(failed)
+            if err is None:
+                device = app.specter.device_manager.add_device(name=device_name, device_type=device_type, keys=keys)
+                return redirect("/devices/%s/" % device.alias)
+        elif action == "newhotdevice":
+            if not device_name:
+                err = "Device name must not be empty"
+            elif device_name in app.specter.device_manager.devices_names:
+                err = "Device with this name already exists"
+            if len(request.form['mnemonic'].split(' ')) not in [12, 15, 18, 21, 24]:
```
stepansnigirev, line 875: mnemonic module has also a check method:
```py
mnemo = mnemonic.Mnemonic('english')
mnemo.check('garlic advice under avocado scene sadness pumpkin skull old mixture drama satoshi')
>> False
```

src/cryptoadvance/specter/devices/bitcoin_core.py
```diff
@@ -0,0 +1,121 @@
...
+        xpubs += "[%s/84'/1'/0']%s\n" % (master_fpr, vpub)
+        # Testnet multisig nested Segwit
+        xpub = bip32.get_xpub_from_path('m/48h/1h/0h/1h')
+        Upub = convert_xpub_prefix(xpub, b'\x02\x42\x89\xef')
+        xpubs += "[%s/48'/1'/0'/1']%s\n" % (master_fpr, Upub)
+        # Testnet multisig native Segwit
+        xpub = bip32.get_xpub_from_path('m/48h/1h/0h/2h')
+        Vpub = convert_xpub_prefix(xpub, b'\x02\x57\x54\x83')
+        xpubs += "[%s/48'/1'/0'/2']%s\n" % (master_fpr, Vpub)
+        keys, failed = Key.parse_xpubs(xpubs)
```
stepansnigirev, line 81: Makes sense to export only xpubs for the current network (only testnet or mainnet) as others will not be available anyways.

src/cryptoadvance/specter/templates/device/components/device_type.jinja
```diff
@@ -3,6 +3,7 @@
     <fieldset style="border:none; display: inline;"> 
         <select name="device_type" id="device_type">
             <option value="other">Other</option>
+            <option value="bitcoincore">Bitcoin Core</option>
```
stepansnigirev, line 6: Let's call it `Bitcoin Core (hot wallet)`

src/cryptoadvance/specter/templates/device/new_device.jinja
```diff
@@ -7,77 +7,185 @@
...
+								<td>{{ 4 * i + 2 }}. {{ wordslist[4 * i + 1] }}</td>
+								<td>{{ 4 * i + 3 }}. {{ wordslist[4 * i + 2] }}</td>
+								<td>{{ 4 * i + 4 }}. {{ wordslist[4 * i + 3] }}</td>
+							</tr>
+						{% endfor %}
+					</table>
+				</div>
+				<br>
+				<label>Passphrase: </label>
+				<input type="password" name="passphrase" class="inline" placeholder="Wallet passphrase">
```
stepansnigirev, line 112: To avoid confusion we should call it "Encryption password" as it's not a bip39 password (unlike passwords in hardware wallet).

PR #213, opened 2020-07-12
Refactor HWI

stepansnigirev:
Refactor HWI with async functions.
Code should be more reliable and easier to read.

Core javascript functionality is implemented in HWIBridge class in `static/hwi.js`.
GUI and Specter-specific functions are in two files - `hwi.jinja` everything that doesn't need wallet information, and `hwi_wallet.jinja` that builds functions depending on the currrent wallet.

Could help with https://github.com/cryptoadvance/specter-desktop/issues/203


stepansnigirev:
@ben-kaufman thanks for the review!
I fixed everything you found. I hope :)

stepansnigirev:
I agree, cancelling requests would be nice. I will open an issue on that.
At the moment most of the problems are due to `hwi.enumerate()` command that takes really long time on some machines, mostly because of the simulators. I've heard that enumerate can take ~20 seconds on Windows.

PR #214, opened 2020-07-13
Remove requests timeout

stepansnigirev:
Sometimes the node is slow...
This PR removes all timeouts (except when trigger blockchain rescan).


stepansnigirev:
For some reason Safari doesn't animate the gif sometimes... Probably because it's already loading the next page.
I will figure out later how to fix it.

PR #218, opened 2020-07-13
Update keys table

stepansnigirev:
Small PR, changing the QR code format and table for device keys.
These QR codes can be imported to another Specter (for example to online Specter from airgapped one)

### DIFF
--- a/src/cryptoadvance/specter/key.py
+++ b/src/cryptoadvance/specter/key.py
@@ -168,8 +168,6 @@ class Key:
     def __str__(self):
         if self.derivation and self.fingerprint:
             return f"[{self.fingerprint}/{self.derivation[2:]}]{self.original}"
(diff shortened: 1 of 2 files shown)


src/cryptoadvance/specter/key.py
```diff
@@ -165,6 +165,14 @@ def json(self):
     def purpose(self):
         return purposes[self.key_type]
 
+    def __str__(self):
+        if self.derivation and self.fingerprint:
+            return f"[{self.fingerprint}/{self.derivation[2:]}]{self.original}"
+        elif self.derivation:
```
ben-kaufman, line 171: I think it should be `elif self.fingerprint:` right?

stepansnigirev: ups...

Issue #212, opened 2020-07-12
Allow exports to other wallets


stepansnigirev:
At the moment if Specter fails people don't have an easy way to regain access to their money.
We are still in development so crashes and failures may happen pretty often.
Would be nice to have an option to export wallet to another wallet.
Candidates:
- [ ] sequence of Bitcoin Core and HWI commands
- [x] [Electrum](https://electrum.org/)
- [ ] [Caravan](https://unchained-capital.github.io/caravan/#/)
- [x] [Fully Noded](https://apps.apple.com/us/app/fully-noded/id1436425586)
- [x] [Fully Noded 2](https://github.com/BlockchainCommons/FullyNoded-2)
- [ ] [Sparrow](https://sparrowwallet.com/)
- [ ] JoinMarket
- [ ] Wasabi

Fonta1n3:
I think it’s worth adding Fully Noded to the list as it is capable of spending any watch-only wallet that exists on a users node simply by loading it and adding bip39 mnemonics as signers. The signers are stored locally to the device and act as psbt signers independently of bitcoind.

stepansnigirev:
> I think it’s worth adding Fully Noded to the list as it is capable of spending any watch-only wallet that exists on a users node simply by loading it and adding bip39 mnemonics as signers. The signers are stored locally to the device and act as psbt signers independently of bitcoind.

Added :)

Fonta1n3:
Just to continue the conversation with @ben-kaufman here. To import the wallet all we need is the receive descriptor, and ideally the block-height the wallet was created at to automate rescanning. 

I know json is not ideal for QR codes but it is what we are using in FN2 for now and it gets the job done. We call them "Account Maps" and they include the receive descriptor (Bitcoin Core format) with full path and checksum, a user defined label, and a blockheight. When we scan an Account Map we know it includes the receive descriptor so we simply replace `/0/*` with `/1/*`, and carry out two importmulti commands to derive all the receive and change keys.

An example:
```
[...]

PR #220, opened 2020-07-13
Wallet export to other software

ben-kaufman:
Starts work on #212.
Adds an export wallet button to the wallet settings page. Clicking the button will prompt the popup as in the picture below.
The format is as described here: https://github.com/cryptoadvance/specter-desktop/issues/212#issuecomment-657319858
The blockheight property for the wallet is retrieved from the block before the first transaction, or if no transactions exist, the latest block number. The wallet data is available either as a QR code, JSON file, and simple text to copy.
Screen Shot 2020-07-13 at 23 40 58

Cc @Fonta1n3

### DIFF
--- a/src/cryptoadvance/specter/static/styles.css
+++ b/src/cryptoadvance/specter/static/styles.css
@@ -769,19 +769,21 @@ table tr:hover .btn.hovering{
 /************** Overlay Popup ********************************/
 .page_overlay {
     display: none;
-    position: absolute;
+    position: fixed;
     top: 0;
     left: 0;
+    bottom: 0;
+    right: 0;
     width: 100%;
     height: 100%;
-    justify-content: center;
-    align-items: center;
+    overflow: auto;
     background-color: rgba(0,0,0, 0.75);
     z-index: 2;
+    padding: 20px;
 }
 .page_overlay_popup {
+    margin: auto;
     display: none;
-    /*margin: 0 auto;*/
     max-width: 500px;
     text-align: center;
     border-radius: 0.5em;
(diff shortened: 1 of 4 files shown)


stepansnigirev:
At the moment on my screen:
- Top header got cropped (Export wallet part)
- Bottom part scrolls below visible parts of main and sidebar.
(see screenshots)

We need to make changes in css for the overlay. I suggest making it:
- with `position:fixed; top: 0; bottom: 0; left: 0; height: 0;`
- y-scrollable such that if the popup is too large we can scroll and see all of it.

Screenshot 2020-07-14 at 00 02 01

Screenshot 2020-07-14 at 00 02 06

Fonta1n3:
When I scan the QR it is not recognized as valid json as the `/` in the bip32 path needs to be escaped like `\/`, that way we can take the string returned by the QR and pass it directly to the json decoder for parsing.

cc @ben-kaufman

ben-kaufman:
@stepansnigirev I tweaked the overlay CSS, should be good now.
@Fonta1n3 I added the backslashes, could you please retry to make sure that works well now?

stepansnigirev:
Looks good!

Fonta1n3:
> @Fonta1n3 I added the backslashes, could you please retry to make sure that works well now?

Could you please share an updated QR and I can quickly test.

ben-kaufman:
Sure, try this one.
Screen Shot 2020-07-14 at 6 40 20

stepansnigirev:
> > @Fonta1n3 I added the backslashes, could you please retry to make sure that works well now?
> 
> @ben-kaufman Could you please share an updated QR and I can quickly test.
Regtest 1-of-2 multisig.

Screenshot 2020-07-14 at 05 40 11

Fonta1n3:
> Sure, try this one.
> Screen Shot 2020-07-14 at 6 40 20

Sorry to be a pain, but its not recognized as valid json with single quotes around the keys, double quotes will do the trick :) https://stackoverflow.com/a/36038497

ben-kaufman:
Like this?
Screen Shot 2020-07-14 at 7 14 40

Fonta1n3:
> Like this?
> Screen Shot 2020-07-14 at 7 14 40
> 
> Or should I also escape the double quotes?

Works like a charm :)

PR #221, opened 2020-07-13
minor fixes

stepansnigirev:
Fixes:
- If device is set to "Other" it should still have `process_psbt` function
- mnemonic is undefined error in adding keys for a device
- show proper error messages when combine failed due to server error
- allow pasting / scanning / uploading final raw transactions
- extra amount check in new tx screen (0.2 sat is a wrong amount)

### DIFF
--- a/src/cryptoadvance/specter/controller.py
+++ b/src/cryptoadvance/specter/controller.py
@@ -66,23 +66,42 @@ def combine(wallet_alias):
         app.logger.error("SpecterError while combine: %s" % se)
         return render_template("base.jinja", error=se, specter=app.specter, rand=rand)
     if request.method == 'POST': # FIXME: ugly...
-        psbt0 = request.form.get('psbt0') # request.args.get('psbt0')
-        psbt1 = request.form.get('psbt1') # request.args.get('psbt1')
+        psbt0 = request.form.get('psbt0').strip() # request.args.get('psbt0')
+        psbt1 = request.form.get('psbt1').strip() # request.args.get('psbt1')
         if "UR:BYTES/" in psbt0:
             psbt0 = bcur2base64(psbt0)
         if "UR:BYTES/" in psbt1:
             psbt1 = bcur2base64(psbt1)
         txid = request.form.get('txid')
 
-        try:
-            psbt = app.specter.combine([psbt0, psbt1])
-            raw = app.specter.finalize(psbt)
-            if "psbt" not in raw:
-                raw["psbt"] = psbt
-        except RpcError as e:
-            return e.error_msg, e.status_code
-        except Exception as e:
-            return "Unknown error: %r" % e, 500
+        raw = {}
+        # psbt should start with cHNi
+        # if not - maybe finalized hex tx
+        if not psbt1.startswith("cHNi"):
+            raw["hex"] = psbt1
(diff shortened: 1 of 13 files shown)


src/cryptoadvance/specter/templates/wallet/send/new/wallet_send.jinja
```diff
@@ -84,25 +82,32 @@
...
+		function isInvalidValue(unit){
+			return (unit == 'sat' ? amount / 100000000 : amount) < 1e-8;
+		}
+
 		// Form validation
 		function validateAmount() {
 			let amountInput = document.getElementById('amount');
 			amountInput.max = getSpendableAmount(unit);
 
 			if (isAboveWalletBalance(unit)) {
```
ben-kaufman, line 94: This gives a wrong unit error even if I just enter the address before entering the amount.
I think adding this check as the first condition before this line will make it better:
```
if (!amount) {
    return false;
}
```

stepansnigirev: done :)

src/cryptoadvance/specter/device.py
```diff
@@ -11,6 +11,10 @@ def __init__(self, name, alias, device_type, keys, fullpath, manager):
         self.fullpath = fullpath
         self.manager = manager
 
+    def create_psbts(self, base64_psbt, wallet):
```
ben-kaufman, line 14: Works :) But maybe if no method is available, we should show the PSBT/ a button to copy it and tell the user to use the paste transaction button to do the signing? Right now it just brings a popup with no options. You can just take the copy PSBT functionality from the Raw PSBT section in the transaction details.

stepansnigirev: That's the next thing I was thinking about. I thought about adding a "Generic wallet" option that would include QR code with base64 psbt, copy-text and download file - all with the same base64 psbt. Especially taking into account that "Other" option is not available anymore in the import device screen - we show an error if the type is not selected.

ben-kaufman: That sounds even better :)

src/cryptoadvance/specter/controller.py
```diff
@@ -66,23 +66,42 @@ def combine(wallet_alias):
...
+        # try converting to bytes
+        if "hex" in raw:
+            raw["complete"] = True
+            raw["psbt"] = psbt
+            try:
+                bytes.fromhex(raw["hex"])
+            except:
+                return "Invalid transaction format", 500
+
+        if "hex" not in raw:
```
ben-kaufman, line 95: Maybe pedantic, but could just say else :)
```suggestion
        else:
```

ben-kaufman:
Oh, just 2 suggestions for changes which feel too small for another PR:

- Show the Pacman loading animation on creating a hot wallet device, since it too does some lengthy communication with Bitcoin Core.
- Increment the fee steps to 0.5 instead of 0.25 in the send screen, so it doesn't start too small.

stepansnigirev:
> I think I tested most/ all cases, works great :) I guess the Generic device can be either here or in its own PR if you want to keep this to bug fixes.

Working on it already :)

PR #225, opened 2020-07-14
Support import wallet

ben-kaufman:
Add support for importing wallets. The New wallet screen now has an Import Wallet button, allowing to import a wallet (by QR/ JSON file/ text). After submitting the wallet data, (if the format is valid), the user will see a summary of the wallet, and the devices used in it. If Specter does not have a device for some key of the wallet, it will automatically add such a device, and in the summary page let the user set its name (we can add other options too to set the device).

I think the UI of the summary page could use some improvements. I'm not good at it but I'll try to make it a bit nicer, but suggestions are welcomed :)

@Fonta1n3 could you please give me some wallets exported from FN/ FN2 (as QR/ text/ file, doesn't matter) to test import with?


Fonta1n3:
> @Fonta1n3 could you please give me some wallets exported from FN/ FN2 (as QR/ text/ file, doesn't matter) to test import with?

@ben-kaufman Just FYI because FN2 is still not mainnet enabled I have not yet added Account Map support to Fully Noded, but I will get that done later today and release to App Store asap.

Here is a single sig Account Map (testnet): 
image

And a multi-sig (testnet):
image

src/cryptoadvance/specter/controller.py
```diff
@@ -467,21 +468,122 @@ def new_wallet(wallet_type):
...
+                    if createwallet:
+                        wallet_name = request.form['wallet_name']
+                        for i, unknown_cosigner in enumerate(unknown_cosigners):
+                            unknown_cosigner_name = request.form['unknown_cosigner_{}_name'.format(i)]
+                            device = app.specter.device_manager.add_device(name=unknown_cosigner_name, device_type='other', keys=[unknown_cosigner])
+                            keys.append(unknown_cosigner)
+                            cosigners.append(device)
+                        wallet = app.specter.wallet_manager.create_wallet(wallet_name, sigs_required, address_type, keys, cosigners)
+                        try:
+                            wallet.cli.rescanblockchain(startblock, timeout=1)
```
stepansnigirev, line 539: will fail if the node is pruned - maybe `flash()` an error if startblock < pruneheight and rescan from pruneheight

src/cryptoadvance/specter/controller.py
```diff
@@ -467,21 +468,122 @@ def new_wallet(wallet_type):
...
-            "wsh": "Segwit (bech32)",
-            "sh-wsh": "Nested Segwit",
-            "sh": "Legacy",
-        }
-        sigs_total = int(request.form.get('sigs_total', 1))
-        sigs_required = int(request.form.get('sigs_required', 1))
+        if action == "importwallet":
+            wallet_data = json.loads(request.form['wallet_data'])
+            wallet_name = wallet_data['label']
+            startblock = wallet_data['blockheight']
```
stepansnigirev, line 474: If this information is not in json - let's set defaults - for example "Imported Wallet" for wallet_name and current blockheight for startblock (so we don't trigger rescan)

ben-kaufman:
@Fonta1n3 thanks, had to make a small change since, in the derivation, you are using `'` for hardened instead of `h`, but the problem is that when we render it into the page as an attribute, it gets mixed with the quotes of the html attribute. I fixed it on our side so it converts that on the server side, but I'd recommend FN to use `h` in the export method just to be on the safe side with future integrations.

Anyway, after making the fix it works well :)

Fonta1n3:
> @Fonta1n3 thanks, had to make a small change since, in the derivation, you are using `'` for hardened instead of `h`, but the problem is that when we render it into the page as an attribute, it gets mixed with the quotes of the html attribute. I fixed it on our side so it converts that on the server side, but I'd recommend FN to use `h` in the export method just to be on the safe side with future integrations.
> 
> 
> 
> Anyway, after making the fix it works well :)

Will make that change for the next update!

Fonta1n3:
@ben-kaufman There is actually an issue with exporting the descriptor with `h` instead of `'`, but it's easily solved. The point of exporting the checksum along with the descriptor is so that we can save ourselves a `bitcoin-cli getdescriptorinfo` command when importing the wallet. If we manually edit the `'` to be a `h` then the client needs to know to revert the `h` back to `'` in order for the checksum to remain valid, otherwise the `h` invalidates the checksum as bitcoind always uses `'` internally it seems.

ben-kaufman:
Yeah I've noticed the checksum issue, but on our end what I did was just remove and recalculated it. But also using regexs to replace the h in derivation back would work, I just found recalculating simpler to implement.

stepansnigirev:
You can also pass descriptor without checksum to getdescriptorinfo rpc call and get a correct checksum from there.
I think simple replacement of ' to h without recalculating checksum is very unexpected behavior not compliant with descriptor specs. Better without checksum then...

Fonta1n3:
> You can also pass descriptor without checksum to getdescriptorinfo rpc call and get a correct checksum from there.
> 
> I think simple replacement of ' to h without recalculating checksum is very unexpected behavior not compliant with descriptor specs. Better without checksum then...

Agree, will update FN side of things so checksum is dropped and use `h` instead.

Fonta1n3:
Just fyi @ben-kaufman @stepansnigirev i have made all the changes and built specter from source to test and it works really well. Amazing!

Uploading the update to App Store now. Thanks for the support.

PR #234, opened 2020-07-15
Support HWI toggle_passphrase

ben-kaufman:
Adds a toggle passphrase button for Trezor and Keepkey devices in the device tab.
In theory, it would have been best to find the device by fingerprint, so you see only the specific device in the device screen, but there is an issue with that for KeepKey (and maybe Trezor One) where you can't get fingerprint if the passphrase is enabled without entering a non-empty passphrase. So that's something we should try to fix, but I think for now it's fine the way it is.

### DIFF
--- a/src/cryptoadvance/specter/static/hwi.js
+++ b/src/cryptoadvance/specter/static/hwi.js
@@ -46,18 +46,14 @@ class HWIBridge {
             { device_type: type, rescan_devices: rescan });
     }
 
-    async togglePassphrase(device, passphrase="") {
+    async togglePassphrase(device) {
         /**
             Tells the server to send the 'togglepassphrase' command to the device.
             KeepKey and Trezor only.
         **/
-        if(!('passphrase' in device)){
-            device.passphrase = passphrase;
-        }
         return await this.fetch('toggle_passphrase', {
             device_type: device.type,
-            path: device.path,
-            passphrase: device.passphrase,
+            path: device.path
         });
     }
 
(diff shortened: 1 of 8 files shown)


src/cryptoadvance/specter/static/hwi.js
```diff
@@ -45,6 +45,22 @@ class HWIBridge {
...
     }
+
+    async togglePassphrase(device, passphrase="") {
+        /**
+            Tells the server to send the 'togglepassphrase' command to the device.
+            KeepKey and Trezor only.
+        **/
+        if(!('passphrase' in device)){
+            device.passphrase = passphrase;
+        }
```
stepansnigirev, line 56: for this command we don't need a passphrase, it's ignored if I understand correctly

src/cryptoadvance/specter/templates/device/device.jinja
```diff
@@ -95,5 +100,37 @@
 				}
 			}, false);
 		});
+
+		async function togglePassphrase(){
+        	let devices = await enumerate('{{ device.device_type }}');
+			if(devices == null || devices.length == 0){
+				return;
+			}
+			let device = await selectDevice(devices);
```
stepansnigirev, line 109: It's very confusing when you disable the passphrase and GUI is asking for a passphrase.
Let's add an optional argument in `selectDevice` function - `passphrase=null`. If we don't care about the password or we know what password we want to use - we can pass it to the function and the GUI window will not appear, and password from the argument will be used.
Then here we can provide an empty passphrase and avoid a passphrase window.

PR #236, opened 2020-07-16
Backup wallets

ben-kaufman:
Add a backup wallets button in the general settings tab. This will download a zip with all wallets account maps as separate zip files, so they could be imported back into Specter if needed.


stepansnigirev:
Would it make sense to backup devices as well? Maybe with a different button?
Like two buttons in a row - "backup wallets" , "backup devices"

ben-kaufman:
Yes that's what I thought to do in another PR.

Issue #226, opened 2020-07-14
Warn the user if he is signing with a wrong device


stepansnigirev:
Related to this:
https://github.com/cryptoadvance/specter-desktop/issues/203

If fingerprint doesn't match we should show a warning. Or filter devices by fingerprint (devices with not-matching fingerprints should have "disabled" state but still visible)

ben-kaufman:
I could implement that if you haven't started yet.

stepansnigirev:
I didn't, so feel free to implement that.
Don't forget that if passphrase is required we don't know the fingerprint ;)

ben-kaufman:
Fixed in #240

Issue #78, opened 2020-02-27
Sign with an HWI device with matching fingerprint, not just with the same type


stepansnigirev:
At the moment when I sign with a USB device, HWI tries to find a device with the correct type but ignores the fingerprint. We can get the fingerprint of the device from the corresponding dictionary and filter available devices with matching fingerprints.

I bet that if at the moment one tries to sign with two Trezors connected one will fail.

PR #240, opened 2020-07-16
Disallow signing with wrong device

ben-kaufman:
Solve #226 

This works also for Trezor and KeepKey with passphrase enabled, by querying the fingerprint "manually" in the enumerate. Also made a PR for HWI to fix that behavior there https://github.com/bitcoin-core/HWI/pull/357

We should probably base on a similar approach as I did here for all device-specific communications (ref #78) by passing fingerprints, something I was wondering about is whatever we should assume a device can only have a single fingerprint (if yes we should enforce that).

### DIFF
--- a/src/cryptoadvance/specter/hwi_rpc.py
+++ b/src/cryptoadvance/specter/hwi_rpc.py
@@ -37,14 +37,16 @@ class HWIBridge(JSONRPC):
         Returns a list of all connected devices (dicts).
         Standard HWI enumerate() command + Specter.
         """
-        self.devices = hwi_commands.enumerate()
-        self.devices += specter_enumerate()
+        self.devices = hwi_commands.enumerate(passphrase)
+        self.devices += specter_enumerate(passphrase)
         for device in self.devices:
             client = self._get_client(device_type=device['type'], path=device['path'], passphrase=passphrase, chain=chain)
             try:
                  device['fingerprint'] = client.get_master_fingerprint_hex()
             except:
                 pass
+            if client:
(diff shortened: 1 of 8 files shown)


stepansnigirev:
For me it breaks "Toggle passphrase" button for Keepkey - PIN confirmation screen doesn't appear.

src/cryptoadvance/specter/hwi_rpc.py
```diff
@@ -32,13 +32,19 @@ def __init__(self):
...
         Standard HWI enumerate() command + Specter.
         """
         self.devices = hwi_commands.enumerate()
         self.devices += specter_enumerate()
+        for device in self.devices:
+            client = self._get_client(device_type=device['type'], path=device['path'], passphrase=passphrase, chain=chain)
+            try:
+                 device['fingerprint'] = client.get_master_fingerprint_hex()
+            except:
+                pass
```
stepansnigirev, line 47: I think you need to close client here.

src/cryptoadvance/specter/static/hwi.js
```diff
@@ -92,6 +94,7 @@ class HWIBridge {
         /**
             Sends the current psbt to the server to relay to the HWI wallet.
         **/
+       alert(device.passphrase)
```
stepansnigirev, line 97: alert...

ben-kaufman: removed :)

stepansnigirev:
I squashed if you don't mind :)

Issue #238, opened 2020-07-16
Server error when clicking "addresses"


Fonta1n3:
I am getting an error when clicking "addresses" (0.5.5 on macos) see the error log:
```
Traceback (most recent call last):
  File "/Users/peter/specter-desktop/src/cryptoadvance/specter/templates/wallet/history/addresses/wallet_addresses.jinja", line 79, in block_content
    }}
  File "/Users/peter/specter-desktop/.env/lib/python3.7/site-packages/jinja2/environment.py", line 430, in getattr
    return getattr(obj, attribute)
  File "/Users/peter/specter-desktop/src/cryptoadvance/specter/wallet.py", line 464, in active_addresses
    return list(dict.fromkeys(self.addresses + self.utxo_addresses))
  File "/Users/peter/specter-desktop/src/cryptoadvance/specter/wallet.py", line 402, in utxo_addresses
    key = lambda utxo: next(
  File "/Users/peter/specter-desktop/src/cryptoadvance/specter/wallet.py", line 403, in 
    tx for tx in self.transactions if tx["txid"] == utxo["txid"]
StopIteration

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/peter/specter-desktop/.env/lib/python3.7/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/peter/specter-desktop/.env/lib/python3.7/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/Users/peter/specter-desktop/.env/lib/python3.7/site-packages/flask_login/utils.py", line 269, in decorated_view
    return func(*args, **kwargs)
  File "/Users/peter/specter-desktop/src/cryptoadvance/specter/controller.py", line 757, in wallet_addresses
    return render_template("wallet/history/addresses/wallet_addresses.jinja", wallet_alias=wallet_alias, wallet=wallet, alladdresses=alladdresses, viewtype=viewtype, specter=app.specter, rand=rand)
  File "/Users/peter/specter-desktop/.env/lib/python3.7/site-packages/flask/templating.py", line 140, in render_template
    ctx.app,
  File "/Users/peter/specter-desktop/.env/lib/python3.7/site-packages/flask/templating.py", line 120, in _render
    rv = template.render(context)
[...]

Issue #239, opened 2020-07-16
Recent transactions not appearing in "recent transactions"


Fonta1n3:
My most recent (many confs) transactions are not showing up for my multisig wallet in the latest update (0.5.5 macos). The balance is correct though.

ben-kaufman:
@Fonta1n3 we removed the full Addresses view and left just the UTXO view part of it. Could you please try to pull master and check if now data is displayed correctly?

Fonta1n3:
It is there now, thanks!

PR #242, opened 2020-07-17
Remove local caching

ben-kaufman:
The goal of this PR is mainly performance improvements and bug fixes.
The current Addresses history feature is nice, but it relies on data that is very hard and network costly to retrieve from Bitcoin Core. Since it is not really in the scope of Specter's architecture to be an indexing server, and since this feature seems to be too problematic without such a server, we decided to remove it, at least until we add support for existing indexing servers like Electrum Server.

Changes:
- Remove caching of transactions
- Remove Addresses History tab
- Merge the UTXO tab with Transactions tab
- Transaction history pagination

This should resolve: #238 #239


stepansnigirev:
All transactions look like unconfirmed now...

ben-kaufman:
I don't see it on my side... Do they also say pending? Or are they saying confirmed but still displayed in unconfirmed color?

stepansnigirev:
All pending...
Screenshot 2020-07-17 at 17 36 06

ben-kaufman:
And generating a new block just to make sure doesn't help I guess? Any idea on how to reproduce or nothing special?

stepansnigirev:
Also transaction history page became much slower...

ben-kaufman:
I made a slight change which should make it a bit faster. For me it's now pretty much the same as before, maybe even slightly better (with the last commit), can you please try with that?

PR #245, opened 2020-07-18
redirect to pending transactions in /send if there are any

stepansnigirev:
Pending transactions are locking funds used in these transactions and people get confused.
This PR redirects to pending transactions screen from `/send` tab if there are any pending transactions.
If there are no pending transaction we load the usual `/send/new` tab.

### DIFF
--- a/src/cryptoadvance/specter/controller.py
+++ b/src/cryptoadvance/specter/controller.py
@@ -788,9 +788,23 @@ def fees(blocks):
     res = app.specter.estimatesmartfee(int(blocks))
     return res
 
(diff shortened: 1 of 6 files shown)


ben-kaufman:
Good idea, but currently this breaks the menu bar in the send tab if you have unsigned txs. You can't navigate to the New tab... I think it's just a matter of updating the menu stuff for `wallet_send` to `wallet_sendnew`.

stepansnigirev:
Fixed!

ben-kaufman:
Now that works, but the Open button of an unsigned PSBT stopped working with a similar error.

stepansnigirev:
Damn... I hope I fixed it everywhere now

PR #249, opened 2020-07-18
Add autodetect Bitcoin Core configuration to settings page UI

ben-kaufman:
Currently, the auto-detect setting for connecting to Bitcoin Core is hidden behind the scenes, which can be a bit confusing.
Here I added it to the Bitcoin Core settings page and made it visible to the user that he can either configure manually or set the auto-detect on.


stepansnigirev:
Nice switch :)

Issue #246, opened 2020-07-18
Feature request: sendmany


stefanwouldgo:
It would be really useful to be able to send to multiple addresses in one tx. Maybe it's even already possible, but I couldn't find the documentation for it.

stepansnigirev:
It is a very good idea, currently it's not possible with Specter, but Core supports it, so we should do it.

This would require some changes in how we represent transactions in the history tab - to allow multiple addresses there. But certainly possible and nice feature to have!

ben-kaufman:
I actually think that with the recent change to history from yesterday, it might be working out of the box. Just need a way to pass multiple addresses, labels, and amounts in send and I think we could support that fairly easily :)

stepansnigirev:
And change the way we save pending transactions - there should be not one address but a list.

ben-kaufman:
Right, we'll need to modify the pending and signing screens a bit, but not by much I think.

ben-kaufman:
Added in #252

PR #252, opened 2020-07-21
Add Multiple Recipients Transactions Support

ben-kaufman:
Solves #246.

Adds the ability to send transactions to multiple recipients. Also, since the send form has started to get a bit full and messy, I have hidden anything "advanced" (coin selection, fee options, and adding recipients button, possibly more things in the future) under an Advanced collapsible div, so it looks much cleaner but still easy to access.
Below is the default form the user first sees, and underneath the form with advanced option open and 2 recipients.

Screen Shot 2020-07-21 at 20 57 03
Screen Shot 2020-07-21 at 20 56 52


stepansnigirev:
"Add another recipient" clears the whole form.
It's a bit sad because maybe I add a new recipient when I've already entered details for the first one...

PR #259, opened 2020-07-24
remove dependency on bip32

stepansnigirev:
bip32 depends on coincurve that doesn't work on ARM like raspberry pi etc.
We only need it to derive child keys from root xprv during setup of the Core hot wallet.
This derivation is pretty easy - it's just addition modulo N.
It's not constant time, but I think that's ok-ish, there are much easier ways to get private key i.e. by sniffing communication with Core.
I also moved `seed_to_hd_master_key` function from helpers to `bitcoin_core.py` because it's used only there and not expected to be used anywhere else in Specter.

Checked against https://iancoleman.io/bip39/ - derives the same keys.

### DIFF
--- a/src/cryptoadvance/specter/devices/bitcoin_core.py
+++ b/src/cryptoadvance/specter/devices/bitcoin_core.py
@@ -1,10 +1,13 @@
 import os, shutil
-from bip32 import BIP32
 from mnemonic import Mnemonic
 from hwilib.descriptor import AddChecksum
 from ..device import Device
-from ..helpers import alias, convert_xpub_prefix, encode_base58_checksum, get_xpub_fingerprint, seed_to_hd_master_key
+from ..helpers import (alias, convert_xpub_prefix,
+                       encode_base58_checksum, decode_base58,
+                       get_xpub_fingerprint)
 from ..key import Key
+from io import BytesIO
+import hmac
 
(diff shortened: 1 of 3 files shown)


src/cryptoadvance/specter/devices/bitcoin_core.py
```diff
@@ -1,10 +1,13 @@
...
 from mnemonic import Mnemonic
 from hwilib.descriptor import AddChecksum
 from ..device import Device
-from ..helpers import alias, convert_xpub_prefix, encode_base58_checksum, get_xpub_fingerprint, seed_to_hd_master_key
+from ..helpers import (alias, convert_xpub_prefix,
+                       encode_base58_checksum, decode_base58,
+                       get_xpub_fingerprint)
 from ..key import Key
+from io import BytesIO
+import hmac
```
ben-kaufman, line 10: Seems like you are missing import `hashlib` and `b58encode` from helpers... But after adding them like so and retrying it works great :)
```suggestion
import hmac
import hashlib
from ..helpers import b58encode
```

stepansnigirev: I reused `encode_base58_checksum` instead and also deleted `b58encode` function that is not used anymore.

PR #260, opened 2020-07-25
Edit auto-detect Bitcoin datadir path

ben-kaufman:
Close #3 
Allow the user to specify his Bitcoin Core data directory path manually.

After though - maybe "Auto-detect" is not the best phrasing for the feature, as what it actually does is load from folder (which if is the default one, will result in automatic detection). Might make sense to change it to something like "Load from folder"? But not sure if that's not a bit confusing as well...

PR #261, opened 2020-07-26
Add full backup and restore options

ben-kaufman:
Add UI options for a user to fully backup the wallets and devices files and restore them from a backup.


src/cryptoadvance/specter/device_manager.py
```diff
@@ -96,3 +100,15 @@ def remove_device(self, device, wallet_manager=None):
...
+    def devices_backup_file(self):
+        memory_file = BytesIO()
+        with zipfile.ZipFile(memory_file, 'w') as zf:
+            for device in self.devices.values():
+                data = zipfile.ZipInfo('{}.json'.format(device.name))
+                data.date_time = time.localtime(time.time())[:6]
+                data.compress_type = zipfile.ZIP_DEFLATED
+                zf.writestr('{}.json'.format(device.name), device.json)
+        memory_file.seek(0)
+        return memory_file
```
stepansnigirev, line 114: I am not sure it should be a property, better use a function.
When I am using property somewhere I assume that it's fast, so I can access it many times without carying much about caching.
But reading all files and zipping them may not be fast.

src/cryptoadvance/specter/specter.py
```diff
@@ -314,3 +320,31 @@ def hwi_bridge_url(self):
...
+                for device in self.device_manager.devices.values():
+                    data = zipfile.ZipInfo('{}.json'.format(device.alias))
+                    data.date_time = time.localtime(time.time())[:6]
+                    data.compress_type = zipfile.ZIP_DEFLATED
+                    zf.writestr(
+                        'devices/{}.json'.format(device.alias),
+                        json.dumps(device.json)
+                    )
+        memory_file.seek(0)
+        return memory_file
```
stepansnigirev, line 347: Here as well, let's use a function instead of property.

ben-kaufman:
@stepansnigirev I removed the property from `specter_backup_file` and completely deleted the `wallets_backup_file ` and `devices_backup_file` as they weren't used. Can you please check if now the performance is fine again?

stepansnigirev:
Looks like we got a merge conflict as I merged #260. Could you rebase?

ben-kaufman:
Resolved, should be fine now.

stepansnigirev:
Still takes forever, not sure why. You don't have this issue, right?
I tried with python3.7 and 3.8, using bitcoin core 19.

ben-kaufman:
No for me it works the same (python 3.8, core 20), that's very strange as I don't think there's a change here which could affect every call...

src/cryptoadvance/specter/wallet.py
```diff
@@ -191,18 +201,24 @@ def from_json(cls, wallet_dict, device_manager, manager, default_alias='', defau
     def get_info(self):
         try:
             self.info = self.cli.getwalletinfo()
-        except:
+            if self.info['keypoolsize'] == 0:
+                self.keypoolrefill(0, end=self.keypool, change=False)
+                self.keypoolrefill(0, end=self.change_keypool, change=True)
```
stepansnigirev, line 206: These three lines are making all requests increadibly long.
In Core v19 keypoolsize will be always zero.

ben-kaufman: Hmmm so do you know if there's an alternative there valid for both 19 and 20?

PR #263, opened 2020-07-28
UI improvements and minor features/ bug fixes

ben-kaufman:
This PR contains many tiny changes (especially, though not only, UI improvements).

Here is a list (probably misses a few small things) of most changes:
- New status bar at the top right corner with Setting, Tor, and Logout buttons.
- Redesign the Receive tab a bit to make the actual receive address more noticeable, allow copy on click.
- Make sidebar wallets and devices lists foldable.
- HWI pin code delete (backspace) button.
- Settings page, replace tiny notes texts with larger tooltips.
- (Tiny) fix for users list not showing up when saving change from not multi-user auth to multi-user.
- Whiter text colors, more space in some places.
- Slightly reorganize New Device page.
- In the wallet transactions tab, don't show History/ UTXO navbar if there are no UTXOs.
- Replace worrying yellow warning bubbles with friendly-ish white info bubbles.

I'm not so good with UI design so I tried to stick with what we planned in most places, but in others I have made some improvisations, so if something doesn't look so good I could really use some help with ideas in how to improve it.

I think this is also a good opportunity to add other changes too small to justify their own PR, if you have any idea/ suggestions :)


stepansnigirev:
Not sure about white text everywhere and increased paddings in some places.
1. Sidebar became very defocused and chaotic.
I suggest keeping a note a bit dimmer because it's secondary information.
It's also important to keep the text aligned with the icons - they should form kind of square with top and bottom lines of the icon. It's not quite true with the key icon so we could maybe fix that as well.
We can also remove the amount as it is also a privacy concern - maybe I don't want to expose all my amounts to the public where I open Specter.
Instead of "Segwit / Nested" let's show the wallet policy instead - like `Single key` or `M of N multisig`. We could also show small devices icons that are used in this wallet but I didn't find a good way to put them there yet.
sidebar

The "New device" page has now too many different font sizes and font styles. Not sure yet how to fix that, I need to think a bit.

Also let's remove the specter-diy-related notes and device-specific notes in general (like in receive tab "specter can verify blahblahblah")

I still believe that secondary information should not be white, it should be dimmer. But I will research a bit during the week what's the best way to do it.

stepansnigirev:
I like it much better now!
One bug, buttons and text overlap:
Screenshot 2020-07-29 at 18 11 18

ben-kaufman:
Opps, right, fixed.

PR #264, opened 2020-07-29
Derive key origin automatically for depth in {0,1}

hodlwave:
This commit does ~~two~~ three things:
1) Prevent users from adding key origin with invalid derivation where a hardened index follows an unhardened index. 
2) For xpubs at depth 0 and depth 1, it derives the key origin metadata automatically (this required adding back the `bip32` library dependency). This will be particularly useful for the `electrum` device integration, which currently does not show users this metadata but nevertheless [hardcodes its multisig paths at depth 1](https://github.com/spesmilo/electrum/blob/cdda1549e91421723c677d751616660db95afb40/electrum/keystore.py#L989). 
3) Render `Key` object as a string properly when the derivation path is 'm' (removes the forward slash in the key origin)


stepansnigirev:
1. Why can't I use hardened index after non hardened? It doesn't make sense but it's not against bip32.
2. Bip32 dependency breaks compatibility with arm. You can get the parent fingerprint and index without adding this dependency - just convert from base58 and take corresponding bytes.
3. Ok

hodlwave:
1. Huh, you are totally right, I thought that was in the spec. Will remove this logic.
2. I’ll get rid of the dependency and unpack the bytes.

stepansnigirev:
Merged! Thanks :)

PR #267, opened 2020-08-02
Add Bitcoin Core node info overview

ben-kaufman:
- Show Bitcoin Core node version on the sidebar.
- Add Bitcoin Core node info overview popup when clicking on the Bitcoin sidebar logo.
- Add warning if the user is connected to an unsupported Core version (<0.19.0).
- Add recommendation in node info if the node is pruned to use unpruned.

For now, I simply added a few stats I thought might be nice to have about the node/ blockchain etc. Still thinking/ looking for suggestions on stuff to add/ remove/ change so to organize it in a way that makes more sense.

### DIFF
--- a/src/cryptoadvance/specter/controller.py
+++ b/src/cryptoadvance/specter/controller.py
@@ -1267,17 +1267,24 @@ def device(device_alias):
             device=device, device_alias=device_alias, purposes=purposes, wallets=wallets, error=err, specter=app.specter, rand=rand)
 
 
-
 ############### filters ##################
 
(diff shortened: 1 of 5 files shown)


stepansnigirev:
Let's move the version number to the bottom line, I think it's secondary information.

Should we also suggest the user set `blockfilterindex=1` if he uses an unpruned node? Rescanning will be way faster in this case and doesn't take much space.

ben-kaufman:
Makes sense, added the recommendation and moved the version number to the lower line.

PR #273, opened 2020-08-05
Create Specter Desktop tray app

ben-kaufman:
This will create a `.app` which can then be wrapped into dmg and give us simple installation (at least on Mac, didn't try Linux/ Windows yet).
The problem for now is that, for some unclear reason, the app works only when launched from the terminal (ie. `open ./path-to-app/Specter.app`, but doesn't work when launched by double click from Finder. Still investigating that so keeping this PR as a draft for now.


stepansnigirev:
I think we should run specter server not as `--daemon`, but as a thread of this app. Whenever you quit the app the server should quit as well.
This might also fix problems with launching the app.

stepansnigirev:
Would it make sense to start the server and open the web browser as soon as the server is ready at launch?

stepansnigirev:
Oh, and maybe we should add "open HWI bridge settings" in that menu!

ben-kaufman:
I was thinking to add app settings, and there allow the user to select if he uses local Specter or remote, so the remote option will switch to HWIBridge. Also the user will be able to specify URL in the remote option whih will be opened instead of the localhost, and will automatically be whitelisted for HWIBridge. What do you think?

stepansnigirev:
Yes, I think that's a good idea. Then starting server in the beginning doesn't make sense - the user has to specify the mode first.

ben-kaufman:
I added a Preference popup, which will automatically launch on first open to set Local/ remote, then it will open whatever the user chose automatically.
The only thing now left I think is to auto whitelist in HWIBridge the domain given in the settings when using a remote Specter.

stepansnigirev:
Maybe binaries should be removed from the PR and added to `.gitignore`?

pyinstaller/specter_desktop.py
```diff
@@ -0,0 +1,329 @@
...
+            return
+
+
+def run_specterd(menu):
+    global specterd_thread, wait_for_specterd_process
+    try:
+        specterd_command = [resource_path('specterd/{}'.format(
+            'hwibridge' if settings.value(
+                "remote_mode", defaultValue=False, type=bool
+            ) else 'specterd'
```
stepansnigirev, line 148: should add `.exe` extension on windows

pyinstaller/specter_desktop.spec
```diff
@@ -0,0 +1,87 @@
...
+          debug=False,
+          bootloader_ignore_signals=False,
+          strip=False,
+          upx=True,
+          upx_exclude=[],
+          runtime_tmpdir=None,
+          console=True )
+
+if sys.platform == 'darwin':
+   app = BUNDLE(exe,
```
stepansnigirev, line 64: weird spacing...

pyinstaller/specter_desktop.spec
```diff
@@ -0,0 +1,87 @@
...
+                bundle_identifier=None,
+                info_plist={
+                    'NSPrincipleClass': 'NSApplication',
+                    'NSAppleScriptEnabled': False,
+                    'NSHighResolutionCapable': 'True',
+                    'NSRequiresAquaSystemAppearance': 'True',
+                    'LSUIElement': 1
+                })
+if sys.platform == 'win32' or sys.platform == 'win64' or sys.platform == 'linux':
+  exe = EXE(pyz,
```
stepansnigirev, line 76: here as well

.gitignore
```diff
@@ -10,3 +10,4 @@ dist
 .vscode
 cert.pem
 key.pem
+*.dmg
```
stepansnigirev, line 13: maybe add `pyinstaller/specterd/*` here as well?

PR #276, opened 2020-08-05
change device classes structure

stepansnigirev:
I slightly refactored device classes 
- `device_type` and `name` are now defined in the class itself.
- Device selector is loaded dynamically from `device_manager.supported_devices` property.
- Class-specific properties are moved from the constructor and now available from the class as well.
- Every hwi-related device class has `enumerate` and `get_client` class methods

With these changes we can override / extend functionality of hwi, or use work from unmerged PRs by overwriting device logic.
See for example how SpecterClient is added this way. Trezor hwi class was also replaced with a multisig change verification version from https://github.com/bitcoin-core/HWI/pull/355. It will be removed as soon as this PR is merged.

Also fixes a few bugs I noticed:

when adding new device:
- even when electrum type is selected on the webpage, device is created as "Other"
- if device type is selected and user scans the QR code the type is replaced to Specter

deleting the wallet and bitcoin core device didn't work if:
- wallet is stored in datadir/wallets/... folder
- chain is not mainnet


stepansnigirev:
Looks like we have plenty of problems with hwi when running in threaded mode.
We already had these issues with enumerate before, we were able to fix them somehow, but while working on this PR I got the same problems again.
Multi-user wallet reloading issues were caused by multithreading as well.

Maybe we can use flask in a single-threaded mode? For a single user I think it's ok-ish, for multiuser it may cause problems when some long process is taking place (i.e. hwi enumerate or wallet creation).
But I would suggest moving these processes into separate threads or processes manually, without flask being involved. Then we have better control of what's going on.

PR #277, opened 2020-08-06
rescan utxo

stepansnigirev:
For pruned nodes there is no way to rescan the whole blockchain for funds, so Specter becomes pretty useless.
This PR adds functionality to rescan utxo set available on any full node and import transactions to the wallet.

The whole process happens in a separate thread so UI won't freeze.


stepansnigirev:
I figured out that it's not working with pruned nodes, as we can't get txoutproof without blocks on disk.
For pruned nodes the only options are to get these blocks from other nodes or block explorers.
For full nodes utxo rescan is much faster comparing to full rescan though.

stepansnigirev:
The problem is that utxo rescan is not binded to a particular wallet, and there is no easy way to find out which wallet is rescanned. Also two wallets can't be rescanned at the same time with this command.
But I probably can save it in a temporary parameter.

ben-kaufman:
Oh ok, so yeah whatever you think it best :)

PR #282, opened 2020-08-16
Fix bug in `der_to_bytes` helper function

hodlwave:
This commit fixes the bug where the function fails on input: `derivation == 'm'`, which results in the following error:
```
IndexError: list index out of range
```


stepansnigirev:
Thanks! Merged.

PR #286, opened 2020-08-17
blackify codebase

mflaxman:
All I did was run the following from the project root directory:
```
$ black .
```

NBD if you're not interested in this change, I saw @stepansnigirev point out a spacing issue on a PR [here](https://github.com/cryptoadvance/specter-desktop/pull/273#discussion_r470073742) and figured this type of thing would be better handled via automation.

If you'd rather run the code yourself so you don't have to proof-read/trust what I've done, I wouldn't be offended.

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -1,0 +1,12 @@
+name: black
+
+on: [push,pull_request]
+
+jobs:
+  black:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2.3.2
+      - uses: lgeiger/black-action@v1.0.1
+        with:
+          args: ' --verbose --check --force-exclude "\.yml\|\.md" . '
--- a/.github/workflows/zblack.yml
+++ b/.github/workflows/zblack.yml
@@ -1,4 +1,4 @@
-name: black
+name: Black Python Linter
 
 on: [push,pull_request]
 
@@ -9,4 +9,4 @@ jobs:
       - uses: actions/checkout@v2.3.2
       - uses: lgeiger/black-action@v1.0.1
         with:
-          args: ' --verbose --check --force-exclude "\.yml\|\.md" . '
+          args: " --check . "
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -1,4 +1,5 @@
 # requirements for testing
+black==20.8b1
 docker==4.1.0
 pytest==5.2.2
 PySocks==1.7.1


k9ert:
Great idea! Happy to automate this! However i'd vote for a githubAction over this PR. @mflaxman would you do a new PR doing this with GithubAction?

mflaxman:
I haven't used this but it sounds promising (free for open source projects): https://deepsource.io/transformers/

k9ert:
The Github-actions are good and easy to be tested in your own fork. Very appreciated if you would implement it.
Just did a very short look on your link. I would prefer to keep the number of tools to a limit we're using. If there is a reason to use it over github-actions ... let's discuss it.

mflaxman:
This PR will now run `black` on each push/PR and display the output!

Note that this will correctly fail until the whole codebase has been blackified. I don't have admin privileges to confirm, but I believe this passing is not required in order for code to be merged (travis-ci is required for example).

Here is my proposal for blackifying the whole codebase:
1. Merge this PR and try it out in development for a little bit. It will confusingly tell contributors that "some checks were not successful."
2. Someone runs the command to blackify and merges that into master (I'm happy to do it but probably would make sense for it to be a maintainer): 
`$ black .`
3. The repo settings are updated to require all PRs to pass this check in order to be merged. It will be a little annoying for outstanding PRs to blackify their branches, but since black is deterministic it should be simple to produce clean enough diffs. Future PRs will likely need run to black to make their code compliant, so it's actually not that different.

I spent a few hours trying to make `black` run only on new/modified files for a smoother adoption, but GitHub makes this surprisingly very difficult. It would be much easier to run my own git server and add a few commands to do this, but that's bad for obvious reasons.

While gradual change would have been the path of least resistance, I think it's actually a bad idea. Doing it all at once gives maximum view into linting changes and then allows future code review to be focused on what that code is actually doing. Better to rip off the band-aid when it comes to linting instead of having each PR for months (or even years!) contain massive formatting changes that are tangential to the features/bugs being addressed.

mflaxman:
Travis CI appears stuck again (`Expected — Waiting for status to be reported`), so I renamed the file from `black.yml` to `zblack.yml` in case it makes a difference in terms of ordering.

Anyway, my point is that despite (correctly) failing tests I think this is **ready to merge**.

stepansnigirev:
Looks good, thanks! We'll merge it before the next minor release (when we have no open PRs).

mflaxman:
Just to clarify, you can merge this *now*! The linter will correctly fail, but that won't stop you from merging.

stepansnigirev:
I understand that but when it's merged we will have red crosses everywhere and they are very depressing =)

k9ert:
I agree and i think it'll also be a big motivation to quickly blackify the whole codebase when it's finally merged. However, let's do that after 0.8.1.

stepansnigirev:
Merged, will blacklify the whole codebase now.

PR #296, opened 2020-08-20
List Past Addresses

ben-kaufman:
Add a list of the wallet's past addresses in the receive tab.


mflaxman:
I just checked this branch out and it worked for me. Thanks!

PR #300, opened 2020-08-22
escape single quotes that break javascript

luclefleur:
Wallet names with a single quote/apostrophe will break the javascript that enables editing the wallet name on the settings page.

To replicate, change a wallet name to include ' and try re-loading. See javascript console for errors.

This patch escapes any single quotes in the affected fields containing the wallet name. Not sure if this is the best way to do this--I didn't see any jinja filters for escaping js, like django used to have.


stepansnigirev:
Works great, thanks!

PR #304, opened 2020-08-24
Add calculate total supply from Specter

ben-kaufman:
Add calculating the total UTXO set button to the node info popup.


stepansnigirev:
Maybe the button should be called "Run the numbers!"? Like in your tweet, and an info tooltip with an explanation of what it means? Or better be serious?

PR #307, opened 2020-08-25
Warn user if trying to upload ZIP instead of a folder from backup

### LINKED ISSUE #293: Tooltip when restoring a backup needs to be more clear or should accept zip files
Screen Shot 2020-08-20 at 00 17 05

This tooltip needs to either mention you need to extract the zipfile to restore the wallet to avoid some confusion

ben-kaufman:
Allowing to upload a zip would have been better, but the way it is currently built it is a bit problematic, so for now at least just tell the user to unzip and select the extracted folder.
Fix #293

### DIFF
--- a/src/cryptoadvance/specter/templates/settings/general_settings.jinja
+++ b/src/cryptoadvance/specter/templates/settings/general_settings.jinja
@@ -44,7 +44,7 @@
 					<span class="info">
 						<span class="info__title">Restoring Specter from backup<br><br></span>
 						Here you can restore your wallets and devices from an existing Secter backup.<br>
-						Please make sure to unzip the backup file first, then to select the exctarcted folder for upload.<br>
+						Please make sure to unzip the backup file first, then to select the extracted folder for upload.<br>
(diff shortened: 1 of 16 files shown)


stepansnigirev:
Actually I can't select .zip file from the backup, so I never see the warning.
Maybe makes sense to add a note to the info tooltip as well?

ben-kaufman:
Ok, added a note on that there.

src/cryptoadvance/specter/templates/settings/general_settings.jinja
```diff
@@ -44,6 +44,7 @@
 					<span class="info">
 						<span class="info__title">Restoring Specter from backup<br><br></span>
 						Here you can restore your wallets and devices from an existing Secter backup.<br>
+						Please make sure to unzip the backup file first, then to select the exctarcted folder for upload.<br>
```
stepansnigirev, line 47: exctarcted =)

ben-kaufman: 🤦‍♂️  right, fixed

PR #310, opened 2020-08-25
Typo fix in Tor tooltip

TheCharlatan:


ben-kaufman:
Good catch, thanks :)

PR #314, opened 2020-08-26
Fix wrong sat conversion x10

### LINKED ISSUE #313: Extra decimal on conversion from BTC to sats on Send screen
Hi guys,

There's an extra decimal when converting from BTC on the Send screen display.
See attached screenshot.

image

ben-kaufman:
Fix #313


stepansnigirev:
Maybe we should use `1e8` and `1e-8` notation then?

PR #315, opened 2020-08-27
move test requirements away from normal requirements

stepansnigirev:


stepansnigirev:
@k9ert can you look into this one?
I think we should find a way to separate test dependencies away from the main pip installation. What would be the best way to do it?
The method I used here (moving to another requirements file) breaks CI...

test_requirements.txt
```diff
@@ -0,0 +1,4 @@
+# requirements for testing
```
k9ert, line 1: can't we call them reuqirements_tests.txt ?

DEVELOPMENT.md
```diff
@@ -11,7 +10,6 @@ git clone https://github.com/cryptoadvance/specter-desktop.git
 cd specter-desktop
 virtualenv --python=python3 .env
 source .env/bin/activate
-pip3 install -r requirements.txt
```
k9ert, line 14: Not sure why you remove it here. IMHO for Development, you need both?
You want to be able to run the tests locally.

k9ert:
Ah, just realizing that this is already merged.

DEVELOPMENT.md
stepansnigirev: It installs all the requirements with `pip3 install -e .` anyways

Issue #323, opened 2020-08-28
Automate Binary Upload for release


k9ert:
Doing them manually is a waste of time. Might get difficult for MacOS and Windows, though.

stepansnigirev:
Maybe we can look into how HWI does it, AFAIK they build binaries for mac and windows as well

k9ert:
HWI is using travis which do not allow private runners. Also hwi has a lot less releases than we have. 
They had 10 releases in 1,5 years. We have 26 releases since in 3-4 month! That's legit as we 're much more customer-facing than HWI is. My guess is that Andrew is creating the release by hand with some scripts on his machines.

Imho we should try to automate as much as possible. Today i'll create another PR for a windows-binary.

ben-kaufman:
Closing for now since we have it on Gitlab for Linux and Windows but probably no intention to add macOS there right?
Anyway please reopen if you think it should stay open.

PR #322, opened 2020-08-28
Nit - clean json_rpc.py

PulpCattel:
Hey all,

I'm curious about the codebase, so I'm working my way through it. Very nice project, kudos.

I've seen #286 so I'm intentionally ignoring every PEP stylish stuff. 
IDK if you have strong opinions about readability, imho the shorter version is also clearer.

Cheers

PR #324, opened 2020-08-28
Release binaries

k9ert:
see #323 

For now, only linux binaries get created and uploaded.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -36,7 +36,7 @@ test:
     # - python3 tests/conftest.py 
     - pytest --docker
 
-release:
+release_pip:
   stage: releasing
   only: 
     - tags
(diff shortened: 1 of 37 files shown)


stepansnigirev:
I cherry-picked `fix the build` commit to fix the build on master. Was broken by #322 
Do you think it's realistic to get build jobs for osx, win and aarch64 in this PR?

k9ert:
Gitlab provides its runner-binary for [windows](https://docs.gitlab.com/runner/install/windows.html) and [MacOS](https://docs.gitlab.com/runner/install/osx.html). They even have publicly available windows runners. I'm not too happy to use them for security-reasons, though. Building and testing is one thing (which we're already not covering with public but a private runner) but releasing is even more security critical.

stepansnigirev:
So if I understand this PR correctly, it will create a unix binary and upload it to your fork, right?

k9ert:
Yes, that was wrong (fixed). Also there is documentation missing which i should put to https://github.com/cryptoadvance/specter-desktop/blob/master/docs/continuous-integration.md
Give me some more time and i'll try to create a windows-binary as well. There i have to fiddle around with picking the right runner. When i have that, maybe someone else can do the same for MacOS binaries (I don't own a Mac but can provide a windows-runner).
By the way i really think we should restrict ourself to private runners. Sooner or later we want to sign the builds (and hash them before that). Even though it might be possible on public runners, it's a very bad idea imho.

PR #333, opened 2020-08-31
add current specter version to sidebar

mflaxman:
I've been playing around with various features of Specter for a guide I'm writing, and when I stumble on bugs I realized it'd be helpful if the Specter version was visible on screenshots:
Screen Shot 2020-08-30 at 7 28 23 PM

I'm new to this codebase, so unsure whether to use `{{ current_version }}` in the template instead. Locally, this was displaying `vx.y.z-get-replaced-by-release-script` (I think this comes from [setup.py](https://github.com/cryptoadvance/specter-desktop/blob/f9b04003a13c79465f0e02da20ddfcad9546b070/setup.py#L17)).

### DIFF
--- a/src/cryptoadvance/specter/templates/includes/sidebar/sidebar.jinja
+++ b/src/cryptoadvance/specter/templates/includes/sidebar/sidebar.jinja
@@ -94,7 +94,15 @@
             <a href="https://github.com/cryptoadvance/" target="_blank">
                 <img src="/static/img/ca.png">
             </a>
-	    <span>Specter v<strong>{{ specter.get_version }}</strong></span>
+	    <span>Specter Version: 
+		<strong>
+		    {% if specter.get_version == 'vx.y.z-get-replaced-by-release-script' %}
+			custom
+    		    {% else %}
+			{{ specter.get_version }}
+		    {% endif %}
+		</strong>
+	    </span>
         </div>
     </nav>
     <div id="side-expand">
--- a/src/cryptoadvance/specter/templates/includes/sidebar/sidebar.jinja
+++ b/src/cryptoadvance/specter/templates/includes/sidebar/sidebar.jinja
@@ -94,6 +94,7 @@
             <a href="https://github.com/cryptoadvance/" target="_blank">
                 <img src="/static/img/ca.png">
             </a>
+	    <span>Specter v<strong>{{ specter.get_version }}</strong></span>
         </div>
     </nav>
     <div id="side-expand">
--- a/src/cryptoadvance/specter/templates/includes/sidebar/sidebar.jinja
+++ b/src/cryptoadvance/specter/templates/includes/sidebar/sidebar.jinja
@@ -94,15 +94,7 @@
             <a href="https://github.com/cryptoadvance/" target="_blank">
                 <img src="/static/img/ca.png">
             </a>
-	    <span>Specter Version: 
-		<strong>
-		    {% if specter.get_version == 'vx.y.z-get-replaced-by-release-script' %}
-			custom
-    		    {% else %}
-			{{ specter.get_version }}
-		    {% endif %}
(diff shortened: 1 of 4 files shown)


ben-kaufman:
I think it's good to have that there, but I think you should probably use the `{{ current_version }}` as we do in the settings page. The `get_version_info()[1]` returns the latest specter version, not that which is run by the user.
About the `vx.y.z-get-replaced-by-release-script`, this is basically shown when you're running Specter directly from source, so I think it'd be best to just add an if check to see if that's returned and then if so display `master` or `self-compiled` or something like that.

src/cryptoadvance/specter/templates/includes/sidebar/sidebar.jinja
```diff
@@ -94,6 +94,7 @@
             <a href="https://github.com/cryptoadvance/" target="_blank">
                 <img src="/static/img/ca.png">
             </a>
+	    <span>Specter v<strong>{{ specter.get_version }}</strong></span>
```
ben-kaufman, line 97: I'm not really a designer, but I think it would be a bit nicer to just add a small `margin-bottom` or `<br>` at the end of the line so it's not too close to the screen bottom

mflaxman: Good call, done. It was `padding-bottom` that worked.

mflaxman:
Turns out `{{ current_version }}` is only set on some templates, so I had to stick with the one I made with the `get_version` property. This probably makes more sense though, as it's only called at boot and not on a bunch of pageloads.

I went with `custom` instead of `self-compiled` because I figured it would be more obvious to non-technical users.

Then I went with `version: ` instead of just `v` as `vself-compiled` doesn't really make sense.

Look good?

mflaxman:
Screenshot: Screen Shot 2020-08-31 at 8 39 14 PM

I should also mentioned, without the release script I can't test it's working post-deploy!

stepansnigirev:
I think calling `get_version_info()` every time doesn't make sense if we are interested only in the current version - let's cache it.
I suggest having something like `self._current_version = None` in the constructor of `Specter`, and on the first call of the `specter.version` property get it with `get_version_info()[0]`.
Maybe `get_version_info` should return `custom` or `unknown` string depending on the situation and we should keep the jinja template non-conditional. At the moment it is returning `Unknown version` if the version check over pip didn't work.

mflaxman:
Update: I made all these changes. I also changed `Unknown version` -> `Unknown` (asking for a version shouldn't return the word version).

Again, can't test how it works after release but it's correctly showing `custom` for me and the caching is working. I think not just setting `current_version` on init (and using that in templates) is probably overkill as it will be set on first pageview with this init/getter/setter construction anyway but I don't feel strongly.

ben-kaufman (approved):
Looks good to me, thanks!

mflaxman:
Is Travis CI integration working? I can't see what checks aren't completing:
Screen Shot 2020-09-01 at 5 45 12 PM

I see something like this on this PR as well as this one: https://github.com/cryptoadvance/specter-desktop/pull/334

ben-kaufman:
Yeah strange, @k9ert do you think Travis is blocking us again? This seems to happen only for new contributors I think...

k9ert:
IMHO this is a result of how the abuse-detection from Travis CI. From the support:

> The abuse detection system is actually per user account basis and this can be triggered more 
> than one if your builds still behave unexpectedly suspicious.

I'm not sure how many tickets i already openen to github about it. @mflaxman  can you maybe write a short mail to support@travis-ci.com asking to get whitelisted for this repository?

stepansnigirev:
Looks good, merging!
Yes, travis is not running for new contributors. For every new active contributor we need to ask travis to add him to the whitelist.

Issue #332, opened 2020-08-30
Way to See Extended Keys, Paths, and Address Encoding for Existing Wallet via GUI?


mflaxman:
This page on wallet creation looks great:
![Screen Shot 2020-08-30 at 12 29 56 PM](https://user-images.githubusercontent.com/448644/91670165-987b1b00-eae0-11ea-8177-74dc3f5a1de3.png)

When you want to verify the info on (at least a quorum of) your devices it's harder to find. This is decent but it shows devices (not keys) and doesn't give you your `m-of-n`:
![coordinate-multisig-view-policy-specter](https://user-images.githubusercontent.com/448644/91670685-48528780-eae5-11ea-98cb-36b6eeb63eb4.png)

You can clickthrough and see the key details for that device, but it might have many keys (due to derivation paths, address encoding, etc). The `Export Wallet` functionality is good, but a script descriptor is too complicated for 99% of GUI users.

Basically, I'm just asking for a GUI view similar to the one on wallet creation.

Is that something you'd consider adding?

Thanks!

ben-kaufman:
Hi, yes that definitely makes sense. I've opened a PR #335 to add such list to the settings page, let me know what you think and if that's about what you had in mind.

PR #335, opened 2020-08-31
Add wallet keys list to wallet settings

ben-kaufman:
Solve #332

### DIFF
--- a/src/cryptoadvance/specter/controller.py
+++ b/src/cryptoadvance/specter/controller.py
@@ -1186,6 +1186,7 @@ def wallet_settings(wallet_alias):
 
         return render_template(
             "wallet/settings/wallet_settings.jinja",
+            purposes=purposes,
             wallet_alias=wallet_alias,
             wallet=wallet,
             specter=app.specter,
@@ -1195,6 +1196,7 @@ def wallet_settings(wallet_alias):
     else:
         return render_template(
             "wallet/settings/wallet_settings.jinja", 
+            purposes=purposes,
             wallet_alias=wallet_alias,
             wallet=wallet, 
             specter=app.specter,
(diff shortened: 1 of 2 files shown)


mflaxman:
Looks good!

Is there any way to see the number of signatures need (`m` in `m-of-n`)? I'm not seeing it here but could be missing something.

Nitpick: any reason that my coldcard seed is `xpub...` and my others are `Zpub...`? That will like confuse newbies.

ben-kaufman:
Well the number of signatures is always visible at the left sidebar under the wallet name, but I could add it also somewhere there if you think that'll be better.

Not sure about the xpub zpub, for me it shows zpub for all, including coldcard, so seems strange but I'll take another look

stepansnigirev:
xpub representation depends on how you imported them:
- import via USB automatically converts xpubs to slip-132 (zpubs, Zpubs etc)
- import via SD card or pasting in the text field keeps them in the same form they were imported. Probably coldcard uses xpubs in multisig wallet file, so we keep them as xpubs.

Maybe we should always convert them to slip-132 so a simple copy-paste to Electrum and other software wallets would work.

stepansnigirev:
Actually all hardware wallets display xpubs, so better to display xpubs in this table

mflaxman:
> all hardware wallets display xpubs, so better to display xpubs in this table

Not true @stepansnigirev!: 
46C53153-580F-4D67-BC55-B6CD5EBCEDEF_1_105_c

I currently don't have a strong opinion on which to use/how (xpub vs Zpub), but it does seem inherently confusing to end-users to have to deal with both :(

ben-kaufman:
Oh, we didn't check Cobo :( I'll add there a Toggle SLIP-132 button (or maybe any simpler name suggestion?) which will just toggle between xpub and Zpub in the table, is that good?

mflaxman:
Toggle seems like a good solution! Maybe in the future there will be more of a standard, but giving users access to all the info for now seems like the best choice to me.

ben-kaufman:
Cool, added it in #343

PR #337, opened 2020-09-01
Allow specify block number when creting a wallet

ben-kaufman:
We've already received a few requests to add the option to scan from a specific block from the waallet creation screen, so added it here.


ben-kaufman:
Oh maybe I should also add here option to do just rescan UTXO instead so it's possible to choose here between the two

stepansnigirev:
Yes, maybe we could use a radio button - rescan blockchain from block, or rescan utxo only

PR #344, opened 2020-09-01
fetch missing data from block explorer

stepansnigirev:
For pruned nodes there is no way to import old wallet without full blockchain redownload.
This PR makes a workaround - it allows to rescan utxo set and fetch missing information from the block explorer.
At the moment it's hardcoded to blockstream.info.

Todo:
- [x] checkbox "fetch from blockstream.info" in wallet settings
- [x] fetch checkbox when creating a wallet
- [x] use block explorer url defined in settings


PulpCattel:
What do you think about adding in the tooltips something like:
`careful, this may have privacy implication`

ben-kaufman (approved):

stepansnigirev:
> What do you think about adding in the tooltips something like:
> `careful, this may have privacy implication`

Added a warning. Should it be like this or better to hide it into the info tooltip?

Screenshot 2020-09-02 at 14 44 19

PulpCattel:
No strong opinion.
My personal preference would be the more visible the better.

stepansnigirev:
Also added keypool adjustments based on found utxo.

So the best way to import an old wallet is:
- create wallet
- run utxo scan to get utxo (fast)
- run full rescan to get history (slow)

ben-kaufman (approved):

PR #345, opened 2020-09-01
Add message signing using HWI

ben-kaufman:
Support message signing for HWI devices.


stepansnigirev:
Verified on Trezor, Coldcard and Ledger.
Ledger shows the hash of the message only, but it's what they do...

PR #348, opened 2020-09-01
fix logging avoiding dictConfig

k9ert:
Not sure why, maybe python versioning issues but the dictConfig doesn't seem to work properly.

Remark: This behaves for some reason differently if you pip-install a release than when you're running a dev-setup.


k9ert:
Also, since some time, the warning to upgrade is broken because "pip show" seems to return the version-number without the preceding "v". So, making the "v" optional for the check.

stepansnigirev:
Then we should also update `pyinstaller/specterd.py` ( https://github.com/cryptoadvance/specter-desktop/blob/master/pyinstaller/specterd.py#L5-L30 )
Or move it to some function that we can call from there.

stepansnigirev:
I have problems with changing the logger lever from Specter settings.
Whenever I change debug level to INFO or WARNING it saves the value, but I still see debug messages in the stdout. 
When I restart specter it resets to debug again.

k9ert:
> Whenever I change debug level to INFO or WARNING it saves the value, but I still see debug messages in the stdout.
There might be still debug-messages, i get e.g.:
``` 
[2020-09-03 13:15:30,949] DEBUG in connectionpool: Starting new HTTP connection (1): localhost:18443
``` 
I'm not sure why that is but for me this feature works at least for all packages cryptoadvance.specter*

stepansnigirev:
Merging now.

P.S. Somehow got TOC recreated with your PR...

PR #351, opened 2020-09-03
fixes #256 - btc-node-config via env-vars

### LINKED ISSUE #256: Flexible Core-configuration via environment-variables
In certain scenarios, that might make a lot of sense.

k9ert:
copied from the info-popup below:

When auto-detect is on, Specter will check for Environment-Variables (BTC_RPC_USER, BTC_RPC_PASSWORD, BTC_RPC_HOST, BTC_RPC_PORT) to configure the connection or attempt to automatically locate your Bitcoin data 
directory and load your node configurations from it.

### DIFF
--- a/src/cryptoadvance/specter/cli.py
+++ b/src/cryptoadvance/specter/cli.py
@@ -187,6 +187,7 @@ def bitcoind(debug, mining, docker_tag):
     from .bitcoind import (BitcoindDockerController,
                            fetch_wallet_addresses_for_mining)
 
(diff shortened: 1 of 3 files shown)


k9ert:
Usually you probably use the env-vars if you connect to a remote-host. That should be https by default. If you really want to do something stupid (e.g. for testing reasons) you can still specify BTC_RPC_PROTOCOL=http .

stepansnigirev:
Oh, got it. Then it should be fine, sorry!

PR #352, opened 2020-09-03
Cleanups

stepansnigirev:
I started some minor refactoring of the codebase, created a `util` folder with potentially useful isolated modules.
Ideally the whole `helpers.py` should be splitted into modules and moved to util.

At the moment `util` contains:
- `base58` encoding-decoding tools
- `xpub` tools for slip132 conversion and fingerprint extraction
- `bcur` tools for Cobo
- `merkleblock` proofs
- `tor` hidden service utils
- `json_rpc` class used by HWIBridge (and by SpecterRPC in the future)

Even though I almost didn't change the code, things may break because I didn't change imports properly.


k9ert:
LGTM. As long as checks have passed, i'd merge this PR.

PR #354, opened 2020-09-04
Add gen-certificate tool for self-signed certificates

stepansnigirev:
An easier way to generate self-signed certificates.


k9ert:
LGTM

PR #368, opened 2020-09-08
set logger level to INFO by default

stepansnigirev:
Otherwise we get a huuuuge number of debug messages from requests module.
If someone wants to see debug messages, we should be able to configure it from general settings of Specter.


k9ert:
Good point. Looks good, merging.

PR #375, opened 2020-09-10
fix load_jsons and wallet manager update

stepansnigirev:
load_jsons was failing even if only one of the json files is damaged.
wallet manager was ignoring all errors.
This caused problems when a single empty json file was preventing all wallets from appearing in Specter.

With this PR load_jsons will load all files it can even if some of them are damaged.
Wallet manager will only ignore RPC errors (i.e. Core is not running).


ben-kaufman (approved):
Works great!

PR #383, opened 2020-09-12
Use Tor to get pruned node blockchain data and connect to remote Core node

### LINKED ISSUE #363: Add support for connecting to a remote node over Tor
Connecting to a bitcoin core over Tor would be nice to have.
Same for block explorer url.
Probably can be done with requests sessions.
See for example https://stackoverflow.com/questions/42971622/fetching-a-onion-domain-with-requests

ben-kaufman:
Fix #363 

Notes:
- Not tested yet, should work in theory though (if Tor is running) and should anyway not break anything if Tor couldn't be connected to. But needs to first be tested (I'll try to check connecting to remote Tor node, @stepansnigirev can you try maybe the rescan UTXO on a pruned node, can set explorer to Blockstream's `http://explorerzydxu5ecjrkwceayqybizmpjjznk5izmitf2modhcusuqlid.onion/`)
- I made connecting to remote Bitcoin Core node use Tor by default, since Tor will cause calls to be significantly slower, it might make more sense to add this as an option in settings? Or only for `.onion`s?

Edit:

Tested connecting to myNode Bitcoin Core over Tor, works.

### DIFF
--- a/src/cryptoadvance/specter/rpc.py
+++ b/src/cryptoadvance/specter/rpc.py
@@ -265,7 +265,7 @@ class BitcoinRPC:
         if "wallet" in kwargs:
             url = url+"/wallet/{}".format(kwargs["wallet"])
         r = None
-        if 'localhost:' not in url and '127.0.0.1:' not in url:
+        if '.onion' in url:
             try:
                 requests_session = requests.Session()
                 requests_session = requests.Session()
--- a/src/cryptoadvance/specter/rpc.py
+++ b/src/cryptoadvance/specter/rpc.py
@@ -264,8 +264,29 @@ class BitcoinRPC:
         url = self.url
         if "wallet" in kwargs:
             url = url+"/wallet/{}".format(kwargs["wallet"])
-        r = requests.post(
-            url, data=json.dumps(payload), headers=headers, timeout=timeout)
+        r = None
+        if 'localhost:' not in url and '127.0.0.1:' not in url:
+            try:
+                requests_session = requests.Session()
+                requests_session = requests.Session()
+                requests_session.proxies = {}
+                requests_session.proxies['http'] = 'socks5h://localhost:9050'
+                requests_session.proxies['https'] = 'socks5h://localhost:9050'
+                r = requests_session.post(
+                    url,
+                    data=json.dumps(payload),
+                    headers=headers,
+                    timeout=timeout
+                )
+            except Exception:
+                pass  # Tor call failed
+        if r is None:
+            r = requests.post(
+                url,
+                data=json.dumps(payload),
+                headers=headers,
+                timeout=timeout
+            )
         self.r = r
         if r.status_code != 200:
             raise RpcError(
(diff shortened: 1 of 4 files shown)


src/cryptoadvance/specter/rpc.py
```diff
@@ -264,8 +264,29 @@ def multi(self, calls: list, **kwargs):
         url = self.url
         if "wallet" in kwargs:
             url = url+"/wallet/{}".format(kwargs["wallet"])
-        r = requests.post(
-            url, data=json.dumps(payload), headers=headers, timeout=timeout)
+        r = None
+        if 'localhost:' not in url and '127.0.0.1:' not in url:
+            try:
+                requests_session = requests.Session()
+                requests_session = requests.Session()
```
stepansnigirev, line 271:
```suggestion
                requests_session = requests.Session()
```

src/cryptoadvance/specter/rpc.py
```diff
@@ -264,8 +264,29 @@ def multi(self, calls: list, **kwargs):
         url = self.url
         if "wallet" in kwargs:
             url = url+"/wallet/{}".format(kwargs["wallet"])
-        r = requests.post(
-            url, data=json.dumps(payload), headers=headers, timeout=timeout)
+        r = None
+        if 'localhost:' not in url and '127.0.0.1:' not in url:
```
stepansnigirev, line 268: why for all domains? it will certainly fail for any local network ip addresses.
Maybe better to do it only for onion?

```suggestion
        if '.onion' in url:
```

ben-kaufman: ok

src/cryptoadvance/specter/rpc.py
ben-kaufman: Good catch, for some reason github doesn't let me accept the suggestion so I'll just close it here and edit manually

src/cryptoadvance/specter/rpc.py
```diff
@@ -264,8 +264,28 @@ def multi(self, calls: list, **kwargs):
         url = self.url
         if "wallet" in kwargs:
             url = url+"/wallet/{}".format(kwargs["wallet"])
-        r = requests.post(
-            url, data=json.dumps(payload), headers=headers, timeout=timeout)
+        r = None
+        if '.onion' in url:
+            try:
+                requests_session = requests.Session()
+                requests_session.proxies = {}
```
PulpCattel, line 271: Is `requests_session.proxies = {}` needed?

`Session` object already [initialize](https://github.com/psf/requests/blob/967a05bfffcb68f97296eda197b062221c2ebc0d/requests/sessions.py#L373) the attribute `proxies` to {}

src/cryptoadvance/specter/wallet.py
```diff
@@ -432,18 +432,26 @@ def _rescan_utxo_thread(self, explorer=None):
             for tx in existing
         ])
         # handle missing transactions now
-        # TODO: do it over Tor
+        # if Tor is running, requests will be sent over Tor
         if explorer is not None:
+            try:
+                requests_session = requests.Session()
+                requests_session.proxies = {}
```
PulpCattel, line 439: Same as above

src/cryptoadvance/specter/rpc.py
ben-kaufman: Right, that's redundant, removed. Thanks for the review!

PR #387, opened 2020-09-12
FAQ: Troubleshooting rescanning

k9ert:


docs/faq.md
```diff
@@ -346,6 +347,10 @@ Specter only works with python3, so use pip3 to install it
 
 Run it with the
  command `python3 -m cryptoadvance.specter server` - then it will use python3
+ 
+## *I created an existing wallets but even after rescanning, specter couldn't find any funds?*
+
+Make sure you're using the right type of wallet. There are "Nested Segwit" and "Native segwit". Make sure to choose the right one. Also, it's relevant whether you're watching enough addresses. By default only 20 addresses are watched. Maybe your wallet needs more so increase them in the settings-menu of the wallet.Also  if you're using a pruned might be difficult depending on much data your node holds. There are some ideas on how to circumvent that which probably involves privacy issues so it's definitely recommended to use a non-pruned-node if dealing with older wallets.
```
ben-kaufman, line 353:
```suggestion
Make sure you're using the right type of wallet. There are "Nested Segwit" and "Native segwit". Make sure to choose the right one. Also, it's relevant whether you're watching enough addresses. By default only 20 addresses are watched. Maybe your wallet needs more so increase them in the settings-menu of the wallet.
If you're running a pruned node, it's not possible to scan for the entire transaction history without doing a full re-download of the blockchain (IBD). Alternatively, we also support scanning for only the existing wallet balance (UTXO) which is very quick and supports both full and pruned nodes. However, for pruned nodes to support this feature, we must query some external data from an outside source (such as a block explorer, configurable by the user). This does not constitute a security risk, as the validity of the data can be verified against the hash existing on the pruned node itself, but can be a potential privacy risk, although it's possible to get the data over Tor to reduce the potential privacy leak. Yet, we still strongly recommend using a non-pruned-node (if possible) when dealing with older wallets.
```

k9ert:
Guess i have to update after the next release, though. Thanks so far!

PR #392, opened 2020-09-14
Add singlesig wizard

### LINKED ISSUE #390: Replacing Cobo Vault's icon with attachment
Replacing Cobo Vault's icon with attachment
[Cobo Vault.svg.zip](https://github.com/cryptoadvance/specter-desktop/files/5218159/Cobo.Vault.svg.zip)

ben-kaufman:
Adds a wizard for setting up a new device + singlesig wallet.

Also fix #390


stepansnigirev:
Looks good. A few comments:
1. I would expect that when the user selects the device in the wizard he is forwarded to the next step as there is nothing else to do on this page.
2. The same on the next page - after I click "Connect over USB" and "Device data was retrieved succesfully" I expect to get to the next page. Maybe makes sense to show somewhere how many keys were imported (not necessary to show the key, just the number of them). The same should happen for SD card and QR code import.
3. On "Create new wallet" page let's change the wizard icon to something more wizard-y. Something like this maybe: https://www.svgrepo.com/svg/190631/magic-wand-wizard, and rename to "New wallet wizard"
4. On the key import step the wizard should check if at least one of the keys (scanned or uploaded) is single sig (has bip 49 / 84 derivation path or is ypub / zpub / upub / vpub). It's annoying to go through the whole wizard and find out you scanned wrong xpub at the very end.
5. If the wizard fails (i.e. I only scanned multisig xpub) the device should not be created. So makes sense to check if everything went well and if not - clean up.
6. On wallet creation screen disable nested segwit / native segwit radio button if the corresponding key type is missing. For example I shouldn't be able to scan only zpub and create a nested segwit wallet.

PR #393, opened 2020-09-15
Bitbox2 support

ben-kaufman:


stepansnigirev:
`Internal error: The BitBox02 does not accept a passphrase from the host. Please enable the passphrase option and enter the passphrase on the device during unlock..`

This happens in `hwi_rpc.py, line 72, in enumerate` `devcls.get_client(dev["path"], passphrase)`. Maybe we can overwrite `get_client` method in BitBox and make sure the password is not transmitted to the device.

With HWI command line interface it works.

Also getting hwi from git commit didn't quite work when I tried to update using pip in my old virtual environment, I had to make a fresh new virtual environment for this PR. Are we sure we want this before the HWI release?

ben-kaufman:
Can you please just check what `passphrase` on line 72 is (just print it out?). I'm assuming it's `None`, but if you can please confirm it that'd help.

I think it will be nice to have it already, HWI can take a long time for releases... though if it's causing issues we'll have to wait.

stepansnigirev:
Yes, it works when passphrase is an empty string `""`, but fails if it's `None`.
So changing the code in enumerate to this helped with the issue:
```py
                if passphrase is None:
                    client = devcls.get_client(dev["path"])
                else:
                    client = devcls.get_client(dev["path"], passphrase)
```

Now I am getting another one:
"Internal error: Input error: error code: 103, message: generic error. A keypath might be invalid. Supported keypaths are: m/49'/0'/<account'> for p2wpkh-p2sh; m/84'/0'/<account'> for p2wpkh; m/48'/0'/<account'>/2' for p2wsh multisig; account can be between 0' and 99'; For address keypaths, append /0/<address index> for a receive and /1/<change index> for a change address.."

I wasn't able to get m/44h... key and none of testnet keys (m/49h/1h/0h), only works with mainnet keys for some reason.

We may need to `try` getting each key and keep going if it fails.

ben-kaufman:
Makes sense. I can do these changes. But I think it'd also make sense to allow setting the passphrase to None in HWI. I'll propose that in a PR there and also ask if they plan to make a release soon with the Bitbox2 in it.

But for now yes I'll just add the check for none and the try catch for key export.

stepansnigirev:
Ok, to get testnet keys we need to set `client.is_testnet = True`, and then it works.
So I assume we can drop support for m/44h keys, and change client to testnet and back when fetching keys for different networks.

PR #418, opened 2020-09-20
Use random hidden service for pruned nodes rescan

### LINKED ISSUE #408: Fetch Block Explorer data over Tor
#344 added the ability for users connected to pruned nodes to fetch missing data from blockstream.info.

This leaks IP to blockstream in addition to allowing blockstream to link the utxo's (should they wish).

If you connect via tor to blockstream's [onion address](http://explorerzydxu5ecjrkwceayqybizmpjjznk5izmitf2modhcusuqlid.onion/) this will stop leaking the IP address.

If you refresh the tor connection between requests this may stop linking the utxo's based off a common tor circuit.

Timing analysis would still be an issue, this could be resolved by staggering the lookups by random time intervals (in addition to the refreshed tor circuits).

Finally, there are other services (like mempool.space) which have a [tor address](http://mempoolhqx4isw62xs7abwphsq7ldayuidyx2v2oethdhhj6mlo2r6ad/) so it would be best to add both and have the lookup randomly switch between these to further improve lookup privacy.

**Summary**
1. Use tor lookup to stop IP leak
2. Also refresh tor circuit with each lookup
3. Also stagger lookup times to stop timing analysis
4. Also randomly switch between multiple tor services

ben-kaufman:
Fix #408 

Points to consider:
- I added just a small time delay which might not be ideal but otherwise it would probably take too long to do that, might be too slow even now?
- Only found 2 explorers supporting Tor - blockstream.info and mempool.space, ideally we'd want more to choose from, so if anyone knows more we can use please comment.
- Needs to be tested as I don't have a pruned node...

### DIFF
--- a/src/cryptoadvance/specter/util/tor.py
+++ b/src/cryptoadvance/specter/util/tor.py
@@ -1,4 +1,7 @@
 import os
+import secrets
+import time
+from stem import Signal
 from stem.control import Controller
 
 
@@ -59,3 +62,9 @@ def stop_hidden_services(app):
             app.tor_service_id = None
     except Exception:
         pass  # we tried...
+
+def renew_tor_ip():
+    with Controller.from_port(port=9051) as controller:
+        controller.authenticate(password='')
+        controller.signal(Signal.NEWNYM)
+        time.sleep(secrets.SystemRandom().randint(15, 30)) # Ensure new IP used
--- a/src/cryptoadvance/specter/wallet.py
+++ b/src/cryptoadvance/specter/wallet.py
@@ -469,6 +469,9 @@ class Wallet():
                     if requests_session.proxies.get('http', None) == 'socks5h://localhost:9050':
                         try:
                             renew_tor_ip()
+                            requests_session = requests.Session()
+                            requests_session.proxies['http'] = 'socks5h://localhost:9050'
+                            requests_session.proxies['https'] = 'socks5h://localhost:9050'
                         except Exception:
                             pass # just continue
                     if len(explorers) > 0:
--- a/src/cryptoadvance/specter/wallet.py
+++ b/src/cryptoadvance/specter/wallet.py
@@ -442,6 +442,8 @@ class Wallet():
             'http://explorerzydxu5ecjrkwceayqybizmpjjznk5izmitf2modhcusuqlid.onion',
             'http://mempoolhqx4isw62xs7abwphsq7ldayuidyx2v2oethdhhj6mlo2r6ad.onion'
         ]
+        if self.rpc.getblockchaininfo()['chain'] != 'main':
+            tor_explorers = [tor_explorer + '/testnet' for tor_explorer in tor_explorers]
         explorers = []
         if explorer is not None:
             try:
--- a/src/cryptoadvance/specter/wallet.py
+++ b/src/cryptoadvance/specter/wallet.py
@@ -464,7 +464,7 @@ class Wallet():
                 proofs = []
                 # get proofs
                 for tx in missing:
-                    if 'http' in requests_session.proxies and requests_session.proxies['http'] == 'socks5h://localhost:9050':
+                    if requests_session.proxies.get('http', None) == 'socks5h://localhost:9050':
                         try:
                             renew_tor_ip()
                         except Exception:
--- a/src/cryptoadvance/specter/wallet.py
+++ b/src/cryptoadvance/specter/wallet.py
@@ -12,7 +12,9 @@ from io import BytesIO
 from .specter_error import SpecterError
 import threading
 import requests
+import random
 from math import ceil
+from .util.tor import renew_tor_ip
 
 logger = logging.getLogger()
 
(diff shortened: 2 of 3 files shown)


src/cryptoadvance/specter/wallet.py
```diff
@@ -436,27 +438,45 @@ def _rescan_utxo_thread(self, explorer=None):
...
-                        for tx in missing]
+                raws = []
+                proofs = []
                 # get proofs
-                proofs = [requests_session.get(
-                            f"{explorer}/api/tx/{tx['txid']}/merkleblock-proof"
-                            ).text
-                          for tx in missing]
+                for tx in missing:
+                    if 'http' in requests_session.proxies and requests_session.proxies['http'] == 'socks5h://localhost:9050':
```
stepansnigirev, line 467:
```suggestion
                    if requests_session.proxies.get('http', None) == 'socks5h://localhost:9050':
```

src/cryptoadvance/specter/util/tor.py
```diff
@@ -59,3 +62,9 @@ def stop_hidden_services(app):
             app.tor_service_id = None
     except Exception:
         pass  # we tried...
+
+def renew_tor_ip():
+    with Controller.from_port(port=9051) as controller:
+        controller.authenticate(password='')
+        controller.signal(Signal.NEWNYM)
+        time.sleep(secrets.SystemRandom().randint(15, 30)) # Ensure new IP used
```
stepansnigirev, line 70: So this means there will be a 15-30 second delay for each utxo?

src/cryptoadvance/specter/wallet.py
```diff
@@ -436,27 +438,45 @@ def _rescan_utxo_thread(self, explorer=None):
         ])
         # handle missing transactions now
         # if Tor is running, requests will be sent over Tor
+        tor_explorers = [
+            'http://explorerzydxu5ecjrkwceayqybizmpjjznk5izmitf2modhcusuqlid.onion',
+            'http://mempoolhqx4isw62xs7abwphsq7ldayuidyx2v2oethdhhj6mlo2r6ad.onion'
```
stepansnigirev, line 443: These are mainnet only, on testnet requests will fail to fetch data.

stepansnigirev (commented):
Doesn't work for me. Tor service is running, using http://explorerzydxu5ecjrkwceayqybizmpjjznk5izmitf2modhcusuqlid.onion/ as blockexplorer.

```
[2020-09-21 11:51:08,435] INFO in log: Error while receiving a control message (SocketClosed): received exception "[Errno 9] Bad file descriptor"
[2020-09-21 11:51:08,436] INFO in log: Error while receiving a control message (SocketClosed): received exception "peek of closed file"
[2020-09-21 11:51:08,436] INFO in log: Error while receiving a control message (SocketClosed): received exception "peek of closed file"
```

Without tor running and using https://blockstream.info worked.

src/cryptoadvance/specter/util/tor.py
ben-kaufman: Yes, too much I guess? What do you think will be best, maybe none?

ben-kaufman:
Strange, I'll try to check what's the error about...

src/cryptoadvance/specter/util/tor.py
stepansnigirev: It's ok as soon as we show that something is still happening.
So I assume we need a new progress status that utxo rescan is done and we are fetching data over tor now.
At the moment the user only sees that rescan is done but nothing appeared in the wallet.

6102bitcoin:
You may need to add /testnet at the end of the url

src/cryptoadvance/specter/util/tor.py
stepansnigirev: Just checked, for me it either changes IP address without any delay or doesn't change it at all.

src/cryptoadvance/specter/wallet.py
```diff
@@ -436,27 +438,47 @@ def _rescan_utxo_thread(self, explorer=None):
...
+                proofs = []
                 # get proofs
-                proofs = [requests_session.get(
-                            f"{explorer}/api/tx/{tx['txid']}/merkleblock-proof"
-                            ).text
-                          for tx in missing]
+                for tx in missing:
+                    if requests_session.proxies.get('http', None) == 'socks5h://localhost:9050':
+                        try:
+                            renew_tor_ip()
```
stepansnigirev, line 471: after renew of a TOR ip we need to create new session. The old one will use the old IP address.

src/cryptoadvance/specter/wallet.py
ben-kaufman: Fixed in f1ae57da1371a27e9368c1309ee72a0afe51e483

src/cryptoadvance/specter/wallet.py
ben-kaufman: Fixed with 6a900d95e9f77975a1ed826a77dd079a1a29c9f2

ben-kaufman:
I will close this for now... Not giving up, but we will have to first refactor our use of Tor anyway.

I'll first work on improving our Tor integration, have some research to do, then will return to it...

PR #419, opened 2020-09-20
faster rpc

stepansnigirev:
reuse of requests session makes communication with Bitcoin Core faster.
In most cases it's not very noticeable, but sometimes it gives a huge improvement - only first call is slow, following are much faster.
I was able to reproduce such "slow connection" and had 5s requests times without session reuse whereas with session reuse only first call was slow and all others were within a few milliseconds.
It may solve problems for some people if specter's page loads are extremely slow.

It doesn't solve all issues, but at least some improvement.

Also renamed passwd to password.

### DIFF
--- a/src/cryptoadvance/specter/controller.py
+++ b/src/cryptoadvance/specter/controller.py
@@ -496,7 +496,7 @@ def bitcoin_core_settings():
                 err = 'Fail to connect to the node configured: {}'.format(e)
         elif action == "save":
             if current_user.is_admin:
-                app.specter.update_rpc(
+                success = app.specter.update_rpc(
                     user=user,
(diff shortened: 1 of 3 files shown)


ben-kaufman:
This causes unexpected behavior when changing a valid RPC configuration to an invalid one - what happens is that the settings show as if it switched to the invalid one but in practice, it's still on the older valid config. We should either make sure we switch to the invalid config, or tell the user we couldn't connect and revert the changes.
Besides that looks good though :)

stepansnigirev:
fixed. both showing the error now and saving settings to invalid.

ben-kaufman (approved):

PR #421, opened 2020-09-20
move version checker to a thread

stepansnigirev:
We are triggering version check on index and all settings pages.
For me, it is slowing down page loading time by a factor of 2.
If the internet is slow or unavailable it is taking several seconds.

This PR moves version checker to a separate thread which checks for new versions once per hour.
As it is a thread it doesn't slow down page loading, but upgrade messages still can be shown to the user.

I also changed the upgrade message because pip upgrade instructions are confusing for people using mynode / rapiblitz and other nodes-in-the-box.


ben-kaufman:
Works great. If I understand correctly though it will still not show the exact version for binary app installs, maybe it belongs to another PR but would be nice to fix that. Also I see it won't show an upgrade message to binary app users, maybe we can check somehow also for that from pip and see if the latest there has a higher version?

stepansnigirev:
For notifications in binaries we need to put the version number somewhere in the code during the build process. I think it's possible with pyinstaller spec files.
We could fetch the latest release from github api:
https://api.github.com/repos/cryptoadvance/specter-desktop/releases 

I will look into that and make a separate PR for that.

stepansnigirev:
Done. Now binaries should show version as well and do version checks.

ben-kaufman (approved):

PR #423, opened 2020-09-20
add .DS_Store to .gitignore for MacOS developers

mflaxman:
MacOS creates these hidden `.DS_Store` files that shouldn't be committed and makes using git annoying (they pollute `$ git status` and `$ git add .` will not end well).


ben-kaufman (approved):
Yes this file is annoying, thanks.

PR #429, opened 2020-09-23
Show balance as percentage of the total supply of Bitcoin

ben-kaufman:
Seems like [some people are interested](https://twitter.com/_benkaufman/status/1308381774863642624?s=20) in knowing this info, so I added a small note for it.


ben-kaufman:
Moved

PR #430, opened 2020-09-23
Add labels to wallet backup file

### LINKED ISSUE #445: Unable to load wallet folder on both app and browser.
was able to upload files via browser originally but deleted wallets to test recovery and now continually getting the server error below.

The browser does find the files but error show when you tey to load them. 

server error specter wallet

Windows 10
Core 0.20.1
specter 0.8.0

ben-kaufman:
This add the wallet labels to the wallet file and allows auto importing them when restoring a full wallet backup.

Also, fix #445

### DIFF
--- a/src/cryptoadvance/specter/controller.py
+++ b/src/cryptoadvance/specter/controller.py
@@ -351,14 +351,8 @@ def general_settings():
                 as_attachment=True
             )
         elif action == "restore":
-            if request.form['restoredevices']:
-                restore_devices = json.loads(request.form['restoredevices'])
-            else:
-                restore_devices = []
-            if request.form['restorewallets']:
-                restore_wallets = json.loads(request.form['restorewallets'])
-            else:
-                restore_wallets = []
+            restore_devices = json.loads(request.form.get("restoredevices", "[]"))
+            restore_wallets = json.loads(request.form.get("restorewallets", "[]"))
             for device in restore_devices:
                 with fslock:
                     with open(
--- a/src/cryptoadvance/specter/controller.py
+++ b/src/cryptoadvance/specter/controller.py
@@ -351,8 +351,14 @@ def general_settings():
                 as_attachment=True
             )
         elif action == "restore":
-            restore_devices = json.loads(request.form['restoredevices'])
-            restore_wallets = json.loads(request.form['restorewallets'])
+            if request.form['restoredevices']:
+                restore_devices = json.loads(request.form['restoredevices'])
+            else:
+                restore_devices = []
+            if request.form['restorewallets']:
+                restore_wallets = json.loads(request.form['restorewallets'])
+            else:
+                restore_wallets = []
             for device in restore_devices:
                 with fslock:
                     with open(
@@ -399,6 +405,7 @@ def general_settings():
                     wallet_obj = app.specter.wallet_manager.get_by_alias(
                         wallet['alias']
                     )
+                    wallet_obj.import_labels(wallet['labels'])
                     try:
                         wallet_obj.rpc.rescanblockchain(
                             wallet['blockheight']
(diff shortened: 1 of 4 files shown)


src/cryptoadvance/specter/controller.py
```diff
@@ -399,6 +405,7 @@ def general_settings():
                     wallet_obj = app.specter.wallet_manager.get_by_alias(
                         wallet['alias']
                     )
+                    wallet_obj.import_labels(wallet['labels'])
```
stepansnigirev, line 408: What if the old backup is imported? 'labels' will not be there then.
You can use `wallet.get("labels", [])` to get empty if not found.

src/cryptoadvance/specter/wallet.py
```diff
@@ -363,6 +364,28 @@ def rescanutxo(self, explorer=None):
...
+        res = self.rpc.multi([
+            ("getaddressesbylabel", label)
+            for label in labels
+        ])
+        return [{ labels[i]: list(result['result'].keys()) } for i, result in enumerate(res)]
+    
+    def import_labels(self, labels):
+        # format: [
+        #   {'label': ['address', 'address2']}
+        # ]
```
stepansnigirev, line 380: Why not like this?
```
{
 "label": [addresses],
 "another label": [more addresses]
}
```
What's the point of an array?

src/cryptoadvance/specter/wallet.py
```diff
@@ -363,6 +364,28 @@ def rescanutxo(self, explorer=None):
...
+        rpc_calls = []
+        for label_row in labels:
+            label = list(label_row.keys())[0]
+            addresses = label_row[label]
+            for address in addresses:
+                rpc_calls.append(("setlabel", address, label))
+        self.rpc.multi(rpc_calls)
```
stepansnigirev, line 387: makes sense only if there are any rpc calls (that might not be the case if imported wallet doesn't have any labels)
```suggestion
        if len(rpc_calls)>0:
            self.rpc.multi(rpc_calls)
```

src/cryptoadvance/specter/controller.py
```diff
@@ -351,8 +351,14 @@ def general_settings():
...
-            restore_devices = json.loads(request.form['restoredevices'])
-            restore_wallets = json.loads(request.form['restorewallets'])
+            if request.form['restoredevices']:
+                restore_devices = json.loads(request.form['restoredevices'])
+            else:
+                restore_devices = []
+            if request.form['restorewallets']:
+                restore_wallets = json.loads(request.form['restorewallets'])
+            else:
+                restore_wallets = []
```
stepansnigirev, line 361: This is a bit shorter:

```suggestion
            restore_devices = json.loads(request.form.get("restoredevices", "[]"))
            restore_wallets = json.loads(request.form.get("restorewallets", "[]"))
```

stepansnigirev (commented):
Looks good! I left a few comments.

src/cryptoadvance/specter/wallet.py
ben-kaufman: No point, I initially wanted to structure it slightly differently and didn't think about that, fixed.

src/cryptoadvance/specter/controller.py
ben-kaufman: Right, changed it.

src/cryptoadvance/specter/wallet.py
```diff
@@ -363,6 +364,31 @@ def rescanutxo(self, explorer=None):
...
+        #       'label2': ['address3', 'address4']
+        #   }
+        #
+        rpc_calls = []
+        for label in labels.keys():
+            addresses = labels[label]
+            for address in addresses:
+                rpc_calls.append(("setlabel", address, label))
+        if len(rpc_calls)>0:
+            self.rpc.multi(rpc_calls)
```
PulpCattel, line 390:
```suggestion
    rpc_calls=[("setlabel", address, label) for label, addresses in labels.items() for address in addresses]
        if rpc_calls:
            self.rpc.multi(rpc_calls)
```
What about this?

ben-kaufman: Much cleaner, thanks :)

PR #437, opened 2020-09-24
Addition to FAQ:  Backup files not showing when trying to load backups.

calkob:
1st ever pull request on github so hope ive done this right.  Basically had an issue with loading a backups and thought i would put the answers i found in the FAQ.


k9ert:
Cool, thanks for contributing! Looks good but you don't have to insert the headline. In that case i think it might be even broken the way you added it.
We've setup a "github-Action" which will create the Table of content automatically.

PR #440, opened 2020-09-26
fix message signing for specter-DIY

stepansnigirev:
Some time ago we added support for ascii and base64 encoding of the messages in DIY. And the format has to be included in the command.
This PR converts messages to base64 if the message contains `\r` or `\n` characters that are considered as the end of the command in communication with DIY.


ben-kaufman (approved):
Tested, works :)

PR #457, opened 2020-09-29
Add pagination for the UTXO tab

### LINKED ISSUE #451: Add pagination to UTXO tab
The UTXO tab currently becomes nearly unusable when there is a large number of UTXOs in the wallet (100s+).
We should add pagination to that tab the same way we already do for the History tab so the max UTXOs loaded per request is not too much that it causes performance issues.

ben-kaufman:
This both adds pagination to the UTXO tab and does some very significant optimizations to it which speeds it up very significantly with a large number of UTXOs. There probably is some more room for improvement, optimization, and maybe general refactoring around this, but I think for now this should be good enough.

Fix #451

### DIFF
--- a/src/cryptoadvance/specter/wallet.py
+++ b/src/cryptoadvance/specter/wallet.py
@@ -228,8 +228,10 @@ class Wallet():
     def getdata(self):
         try:
             self.utxo = parse_utxo(self, self.rpc.listunspent(0))
+            self.utxo_labels_list = self.getlabels(utxo["address"] for utxo in self.utxo)
         except Exception:
             self.utxo = []
+            self.utxo_labels_list = {}
         self.get_info()
         # TODO: Should do the same for the non change address (?)
         # check if address was used already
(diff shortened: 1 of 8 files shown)


src/cryptoadvance/specter/wallet.py
```diff
@@ -624,11 +626,12 @@ def balance_on_address(self, address):
         return sum(balancelist)
 
     def utxo_on_label(self, label):
-        utxo = [tx for tx in self.utxo if self.getlabel(tx["address"]) == label]
+        utxo = [tx for tx in self.utxo if self.utxo_labels_list[tx["address"]] == label]
+
         return len(utxo)
```
PulpCattel, line 631:
```suggestion
        return len([tx for tx in self.utxo if self.utxo_labels_list[tx["address"]] == label])
```
What about a one-liner here?
If you don't like it then I would suggest renaming `utxo` to `utxos` and removing the blank line.

src/cryptoadvance/specter/wallet.py
```diff
@@ -624,11 +626,12 @@ def balance_on_address(self, address):
...
 
     def balance_on_label(self, label):
-        balancelist = [utxo["amount"] for utxo in self.utxo if self.getlabel(utxo["address"]) == label]
+        balancelist = [utxo["amount"] for utxo in self.utxo if self.utxo_labels_list[utxo["address"]] == label]
         return sum(balancelist)
```
PulpCattel, line 635:
```suggestion
        return sum(utxo["amount"] for utxo in self.utxo if self.utxo_labels_list[utxo["address"]] == label)
```
What about a one-liner here? If you don't like it then I would suggest renaming `balancelist` to `balances`.
Should also be good to remove the list and use a generator comprehension instead, so that no unused list is created.

PR #459, opened 2020-09-30
Update README.md with note about setting server=1 for Bitcoin Core GUI

Maple44:
I ran into the issue of needing to set `server=1` when connecting the Bitcoin Core GUI to Specter. I found the answer in the Telegram group along with some related info. This pull request adds that info to the Readme so that others can find the info there.


stepansnigirev:
Thanks! Merged!

PR #460, opened 2020-09-30
Connect to port 8332 by default if none specified

### LINKED ISSUE #456: Use port 8332 by default when port setting is empty
atm if port is empty in Core settings we connect to port 80. We should probably use port 8332 if nothing is specified in the port field.

ben-kaufman:
Fix #456


stepansnigirev:
This actually doesn't work - it breaks autodetect.
Before I was able to connect to a running node in autodetect mode without specifying the port.
The issue was only if credentials are defined *without* autodetect.
I made another PR that fixes it:
https://github.com/cryptoadvance/specter-desktop/pull/462
So if you are ok with it I will close this one and merge #462

Issue #358, opened 2020-09-06
Wallet name in Coldcard export file must be ascii, 1...20 long


stepansnigirev:
We need to make sure the coldcard file contains a name that CC can understand

PR #467, opened 2020-09-30
fix coldcard wallet export

stepansnigirev:
1. limits wallet name to 20 ascii chars for ColdCard, Cobo and Specter-DIY
2. adds `.txt` extension to the wallet file

Fix https://github.com/cryptoadvance/specter-desktop/issues/358


ben-kaufman (approved):
Works

PR #480, opened 2020-10-04
Docker-compose support (draft PR for comments)

pox:
This effort introduces:

* A dockerfile for specter docker image
* A docker-compose file that can build and run both bitcoind and specter

### Known issues

* RPC credentials not discovered using the `.bitcoin` dir - just hardcoded for now to get it working.
* Node is pruned to speed up testing of this branch (otherwise anyone trying to test this out has to pay 300GB..)
* The specter docker image just does `pip install`, instead of copying the sources from the repo's working dir.

### Further work planned (aside from the known issues above)

* Tor support
* Optimizing the docker images for size
* PGP verify the sources in the bitcoind image
* Harden the bitcoind image (least privileged user)


docker/Makefile
```diff
@@ -0,0 +1,13 @@
+
```
k9ert, line 1: do we really need a Makefile for this. It would be the first one for this project. If we accept it, we should probably use it also for other purposes.
@stepansnigirev what do you think about Makefiles?

docker/specter-desktop/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM python:3
+
+RUN apt-get update && apt-get install -y libusb-1.0-0-dev libudev-dev
+RUN pip3 install cryptoadvance.specter
```
k9ert, line 4: To be honest, for the official docker-image, i really like the idea to install a specifc version via pip.
More about that in a general comment.

k9ert (commented):
Thank you very much, this looks promising. I created some comments already. Here is something more general:
* You did pip-install for the docker-image. You could potentially also did a ADD od the current master. How to decide what to do?
* It depends a lot on the use-case. We definitely want to provide official proper versioned/tagged dockerimages. On these, it's quite reasonable to install a specific version via pip. It makes sense to automate that for releases.
* What's then the purpose of a docker-compose-setup? It's probably an "easy way to demo". So what's wrong doing this with tagged-dockerimages? What would be the usecase for using "master" here? If you don't have a fullnode but want to develop/test, there is also the regtest!

Hope my questions make sense.

nioncode:
You might want to check out my docker-compose setup in https://github.com/cryptoadvance/specter-desktop/pull/481#issuecomment-703307467.

In general, I'd suggest you base your work on the already existing Dockerfiles for bitcoind (https://github.com/lncm/docker-bitcoind) and specter-desktop (https://github.com/lncm/docker-specter-desktop), which already solve many of the things (using tagged releases, dropping privileges, etc.).

docker/specter-desktop/Dockerfile
```diff
@@ -0,0 +1,6 @@
...
+
+CMD ['python3', '-m', 'cryptoadvance.specter', 'server']
```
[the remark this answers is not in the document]
pox: Because of how docker networking works, I think this is not necessary. Traffic gets forwarded to the localhost interface anyway.

pox:
@k9ert thank you for the review!

You raise a good point about how to build the image: from pip or from sources.

I am currently using `pip install` but the plan is to switch to `ADD`ing the files from the repo, for the following reasons:

1. Relying on pypi is yet another later of trust that ought to be removed (or at least be optional).
2. When `pip install` is done in a docker build, there's essentially no way to update a local image without just rebuilding the entire image. Meaning, for development iterations, it's very impractical. Not to mention the fact that even if you would build without the benefit of the docker layer cache, you'd still need to upload to pypi just to rebuild the image. It also makes the docker image build non-deterministic, since it relies on the engine's cached layers in order to determine if the `pip install` `RUN` line must be rerun or not.
3. The purpose of having a compose setup is not just for quick demoing. It can also be the main way one runs a node, since compose is a very mature tool at this point. It can be set up to relaunch containers after reboots, etc. It also unlocks a bunch of best practices that would otherwise be challenging for users to apply, like having dedicated networking that's not exposed to the host.

docker/Makefile
pox: I made a note of this in the commit message, but probably should have said something in this PR draft, too:

the makefile is 100% a convenience script, and not required for this setup. 

Having said that, being able to do `make up` after cloning the repo, from any linux machine, and have the entire thing built and running automatically is pretty darn convenient, so I wouldn't necessarily throw it away just because makefiles aren't used elsewhere. It's a nice-to-have.

Also, Makefiles are powerful and very easy to use (just not write...). You should consider using Make, because it has the potential to make building/testing and installing this project a lot easier. There's a reason this ancient technology is sticking around ;) it keeps proving itself. I use it for basically every project out of convenience, even when it's not a C/C++ project. It's a swiss army knife tool for any kind of local build, and a shortcut to a CLI for development purposes (`make {build,run,test,install,rebuild,lint,...}`).

docker/specter-desktop/Dockerfile
pox: This could be a build `ARG`, where by default it copies the repo files into the image (good for developers, people who want unreleased version), but if a specific version is passed it uses `pip` to install it.

docker/Makefile
k9ert: Yes, i agree here. It's sometimes annoying though when things get complex which are quite easy in a bas-script are quite complicated in a Makefile.
So i'm happy to keep it. CC @ben-kaufman

docker/specter-desktop/Dockerfile
k9ert: Oh, that would be cool! But we can have it as a separate PR, i'd say.

k9ert:
> 1. Relying on pypi is yet another later of trust that ought to be removed (or at least be optional).
Fully agree! We need to create (+ sign) pip-packages in tar.gz-form and upload them to the Github-releases (on top of pushing to pypi). These should then be used for the docker-images.
https://packaging.python.org/tutorials/installing-packages/#installing-from-local-archives

> 2. Meaning, for development iterations, it's very impractical.
Why do you want to develop locally but run the results in docker? I mean there are reasons people doing it (in other projects) but what would be your reason? There is skaffold for this if really needed.

> 3. The purpose of having a compose setup is not just for ...
yeah, ok, but why are these ways of using it are an argument to use ADD instead of pip-install releases?

k9ert:
I really like the work of @nolim1t in the repo @nioncode mentioned. We should try to get that work directly in here. 
Really would like to see some more thought on that.

nolim1t:
You can use bitcoind cookie support too. No need to hardcode rpc if everything is all running on the same machine.

nioncode:
> You can use bitcoind cookie support too. No need to hardcode rpc if everything is all running on the same machine.

Unfortunately that does not work easily, yet (see discussion in #481). It only works if you configure the containers to use the host network, which is not a good practice.

nolim1t:
> > You can use bitcoind cookie support too. No need to hardcode rpc if everything is all running on the same machine.
> 
> Unfortunately that does not work easily, yet (see discussion in #481). It only works if you configure the containers to use the host network, which is not a good practice.

ah.. I see what you mean. It depends on what the objectives are I guess. For embedded devices, take a look at these two projects which auto configures stuff such as [this one](https://github.com/lncm/thebox-compose-system) or [this one - which is a derivative of the first](https://github.com/getumbrel/umbrel)

docker/Makefile
pox: 100%. The trick is realizing when bash script get out of hand, and when makefiles get out of hand...
With makefiles I find the rule of thumb is whenever you find yourself splitting it to multiple files with a whole include hierarchy then it's probably time to reconsider the tooling. It's very powerful and very succinct, so often it goes a long way before it reaches that stage.

docker/specter-desktop/Dockerfile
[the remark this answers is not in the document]
nolim1t: Yep I can confirm its not needed if you use Docker

k9ert:
I'm very happy about the constructive discussion here. The surprising thing to me is, that the work @nolim1t already did is only conceptually related to this PR. So this even further amplifies that we have to think about whats's the purpose of this PR.

If cryptoadvance/specter wants to have docker-images for its releases (and it definitely want to) the only thing we need to do is forking the https://github.com/lncm/docker-specter-desktop and adjusting operational parameters. Haven't digged that much but i'm quite sure that's the essence, right?

We don't even need to change anything in this repo. Also his [embracing of gitlab](https://nolim1t.co/2020/09/21/docker-buildx-on-gitlab.html) very much fits to us as we're releasing on gitlab using private runners which is important when start signing hashes of binaries.

So probably the plan would be to fork his repo, adjust the operational stuff, create our own images and push them to cryptoadvance on dockerhub. We'll probably follow up on all the changes in his repo, being a bit more conservative at the same time. We'll probabyl need a versioning for that repo as well and it needs to be related to the versions here.

Does something like this make sense? Did i miss anything?

nolim1t:
The versioning for the my docker repo follows the same versions as upstream to keep things consistent.

I do have a gitlab equivalent too which does mirroring.


https://gitlab.com/lncm/docker/spector-desktop

nolim1t:
re-forking, the repo is portable and will build on either github or gitlab with little changes.

You'll need to set up some environment variables (I haven't documented this as of yet), and follow the [maintainer release notes](https://github.com/lncm/docker-specter-desktop#maintainer-release-notes) for releasing versions (its as simple as just creating a tag on the repo. It will follow and build whatever versions [in upstream](https://github.com/cryptoadvance/specter-desktop)

k9ert:
Yes, we're also using mirroring. I'm surprised that your gitlab-mirror doesn't have pipelines running? Or that simply not visible to me? I'm also not sure how you're using registry.gitlab.com/ericvh/docker-buildx-qemu. I'm building images on gitlab since a longer time but DIND. Something like:
```
.image_building_job: &image_building_job
  #image: registry.gitlab.com/.../imagebuilder:20190810205300
  image: docker:19.03.0
  services:
  - docker:19.03.0-dind
```
My intentional plan was to also make some sort of monorepo approach and i have some more or less sophisticated image-dependency-bootstrapping stuff where i can dynamically create all my images with timestamps like the above. But your approach definitely looks also very promising.

nolim1t:
I'm using qemu so that buildx can build cross platform stuff. Mostly for ARM images

k9ert:
I guess i would immediately start using that. All that underlying buildx-quemu stuff looks good to me as well. I was first concerned about using that image and at some point we'll probably for the creation-process there as well.
You probably agree that for a security relevant project which is using docker, you probably should own/control all the images which are involved in anything on your project?

nolim1t:
Yes I do agree. 

I'm not really using the `.gitlab-ci.yml` stuff too much. Just made it available to use as an experiment (for now). But im using the github built docker images day to day on multiple devices.

k9ert:
I've forked @nolim1t repo, deleted all tags in it and started to create new ones from the latest release of every major release up from v0.6.0.
The resulting images are super-small (50-60MB) and available for 4 different Archs. 

The repo is here:
https://github.com/cryptoadvance/docker-specter-desktop

The docker-hub link:
https://hub.docker.com/repository/docker/cryptoadvance/specter-desktop/general

So i think we should now think again how we want to go with this PR. Have a look in the other repo, there is also an example for docker-compose.

nolim1t:
I suggest changing this and just putting an example docker-compose in /docs

pox:
The thing that motivated me to start this effort is I wanted a way to trustlessly build and run an entire full node from source code. My goal was to avoid relying on docker hub, github, pypi, gitlab, etc. 
Setting up and upgrading a full node should ideally be as simple as doing `docker-compose up`.

Of course, this doesn't mean the compose file has to be in this repo. It could, if the maintainers see the value of the above, but it can also be separate. It would have to clone this repo and do more stuff in the docker build, but that's not a big deal.

The motivation wasn't to create release builds as docker images or to enable some CI pipeline. I personally wouldn't want to `docker pull` my core node images. I want to build them myself and have a simple way to audit the build (dockefiles solve the problem elegantly, because especially core's build has a lot of annoying little quirks and configuration switches that's a PITA to remember each time you build it).

Anyway, that's just my 2c about the motivation here. Glad I opened this draft PR to have the discussion.

k9ert:
> Anyway, that's just my 2c about the motivation here. Glad I opened this draft PR to have the discussion.
And i think it's very useful so far.

I think you made your point now much clearer and i also think this is a valid usecase. It's just not a very common one. So maybe as a kind of more formal requirement for things like that, can you:
* create a directory like /contrib/docker-compose-zerotrust (or something like that? I like the contrib-directory and then subfolders for specific special solutions
* put all your stuff there with a Readme.md explaining what this solved (and that this is not necessarily the easiest/best way to build/use docker(-compose)? 

Does that make sense?

nolim1t:
I do like the ```contrib/``` suggestion

k9ert:
Are you still interested in this @pox ? I don't want to push but currently we should try to clarify all open PRs ...
Also happy to hear your opinion on the contrib-idea?!

pox:
Putting things in `contrib` makes sense. But if it means having alternate versions of the dockerfiles (e.g. to copy the working directory, rather than `pip install`) then it's unclear whether this should belong in this repo. Could possibly work around this issue with a build arg and some conditional scripting, but not sure it's worth it.

It's probably best to take this to a separate repo that's 100% focused on the above usecase.

nolim1t:
```pip3 install .``` essentially installs whatever is in the current working directory into the system (or container if docker)

PR #481, opened 2020-10-04
cli: allow overriding of bitcoin rpc host during auto-detection

nioncode:
This allows to use the existing auto-detection mechanism from the bitcoin
data dir, but to use a different host than the default 'localhost' that
is used at the moment. This is helpful in docker environments where
specter-desktop and bitcoind run in separate containers, since it is not
possible to connect to another container through localhost (except when
sharing the host network in the containers, which is not recommended).
By providing the hostname of the bitcoind container as `--bitcoin-rpc-host`
and mounting the bitcoind data volume into the specter-desktop container,
auto detection from the bitcoin.conf / cookie file works again without
having to supply rpc user, password, host, and port manually.

### DIFF
--- a/src/cryptoadvance/specter/cli.py
+++ b/src/cryptoadvance/specter/cli.py
@@ -34,12 +34,26 @@ def cli():
 @click.option("--debug/--no-debug", default=None)
 @click.option("--tor", is_flag=True)
 @click.option("--hwibridge", is_flag=True)
-def server(daemon, stop, restart, force, port, host, cert, key, debug, tor, hwibridge):
+@click.option("--bitcoin-rpc-host")
+def server(
+    daemon,
+    stop,
+    restart,
+    force,
+    port,
+    host,
+    cert,
+    key,
+    debug,
+    tor,
+    hwibridge,
+    bitcoin_rpc_host,
+):
     # create an app to get Specter instance
     # and it's data folder
     app = create_app()
     app.app_context().push()
-    init_app(app, hwibridge=hwibridge)
+    init_app(app, hwibridge=hwibridge, bitcoin_rpc_host=bitcoin_rpc_host)
 
     # we will store our daemon PID here
     pid_file = path.join(app.specter.data_folder, "daemon.pid")
--- a/src/cryptoadvance/specter/rpc.py
+++ b/src/cryptoadvance/specter/rpc.py
@@ -142,7 +142,7 @@ def detect_rpc_confs_via_env():
     return rpc_arr
 
 
-def autodetect_rpc_confs(datadir=get_default_datadir(), port=None):
+def autodetect_rpc_confs(datadir=get_default_datadir(), port=None, host=None):
     """Returns an array of valid and working configurations which
     got autodetected.
     autodetection checks env-vars and bitcoin-data-dirs
@@ -157,6 +157,8 @@ def autodetect_rpc_confs(datadir=get_default_datadir(), port=None):
     available_conf_arr = []
     if len(conf_arr) > 0:
         for conf in conf_arr:
+            if host is not None and "host" not in conf:
+                conf["host"] = host
             rpc = BitcoinRPC(**conf)
             if port is not None:
                 if int(rpc.port) != port:
--- a/src/cryptoadvance/specter/server.py
+++ b/src/cryptoadvance/specter/server.py
@@ -41,14 +41,17 @@ def create_app(config="cryptoadvance.specter.config.DevelopmentConfig"):
     return app
 
 
-def init_app(app, hwibridge=False, specter=None):
+def init_app(app, hwibridge=False, specter=None, bitcoin_rpc_host=None):
     """  see blogpost 19nd Feb 2020 """
     # Login via Flask-Login
     app.logger.info("Initializing LoginManager")
     if specter is None:
         # the default. If not None, then it got injected for testing
         app.logger.info("Initializing Specter")
-        specter = Specter(DATA_FOLDER)
+        conf = {}
+        if bitcoin_rpc_host:
+            conf = {"rpc": {"host": bitcoin_rpc_host}}
+        specter = Specter(DATA_FOLDER, config=conf)
 
     # version checker
     # checks for new versions once per hour
--- a/src/cryptoadvance/specter/specter.py
+++ b/src/cryptoadvance/specter/specter.py
@@ -27,14 +27,11 @@ def get_rpc(conf, old_rpc=None):
         conf["autodetect"] = True
     rpc = None
     if conf["autodetect"]:
-        if "port" in conf:
-            rpc_conf_arr = autodetect_rpc_confs(
-                datadir=os.path.expanduser(conf["datadir"]), port=conf["port"]
-            )
-        else:
-            rpc_conf_arr = autodetect_rpc_confs(
-                datadir=os.path.expanduser(conf["datadir"])
-            )
+        rpc_conf_arr = autodetect_rpc_confs(
+            datadir=os.path.expanduser(conf["datadir"]),
+            port=conf.get("port"),
+            host=conf.get("host"),
+        )
         if len(rpc_conf_arr) > 0:
             rpc = BitcoinRPC(**rpc_conf_arr[0])
     else:


k9ert:
Thank you for the PR. Let me please try to understand your issue with the existing functionality. Why don't you simply set the existing env-vars instead of:
* specifying the additional newly created --bitcoin-rpc-host
* and mounting the datadir inside the specter-container

So to me your PR is a bit counterintuitive and only working in that specific scenario: From specters-perspective the rpc-host is somewhere else but on the same "machine" at the same time. However, why is it then separated in the first place? The beauty of docker is imho a clear separation.

The only advantage i see here is that you don't have to fiddle with the password which would be easy for a one-time-setup and also quite easy to automate in some sort of custom installation-procedure.
I guess you're using docker-compose here, right?

nioncode:
Honestly, I did not know about the env vars before digging into the code and applying my patch and then I was a bit confused why all of the env vars have to be supplied and not only partial env vars are allowed.

I guess this is done for simplicity, so that specter either a) connects on localhost with auto-detection, or b) connects to a given host that requires all parameters (host, port, user, password).

Yes, I'm using a docker-compose setup with the following `docker-compose.yml` (requires this PR):
```yaml
version: '3.8'
services:
  bitcoind:
    image: lncm/bitcoind:v0.20.1
    container_name: bitcoind
    volumes:
      - type: bind
        source: /storage/ssd/bitcoin
        target: /data/.bitcoin
    restart: unless-stopped
    stop_grace_period: 20m
    # Use the same ip subnet for rpcallowip as in the network
    # configuration at the bottom of this file to allow access
    # to the rpc interface to other containers.
    # Other containers can simply use the name of this container
    # (i.e. `bitcoind` to connect and don't need to specify the IP).
    command: -rpcallowip=172.28.0.0/16 -rpcbind=0.0.0.0 -testnet=1
  specter:
    # TODO: wait for PR (https://github.com/cryptoadvance/specter-desktop/pull/481) to get merged and released as new version and then go back to lncm/specter-desktop image and drop custom build
    #image: lncm/specter-desktop:v0.8.1
    build: specter-desktop
    container_name: specter
    command: --host 0.0.0.0 --force --bitcoin-rpc-host bitcoind
    restart: unless-stopped
    stop_grace_period: 5s
    volumes:
      - type: bind
        source: /storage/ssd/bitcoin
        target: /data/.bitcoin
      - type: bind
        source: /storage/ssd/specter
        target: /data/.specter
  nginx-proxy:
    image: nginx:alpine
    depends_on:
      - specter
    environment:
      - NGINX_PROXY_HOST=specter
      - NGINX_PROXY_PORT=25441
    volumes:
      - type: bind
        source: $PWD/nginx-reverse-proxy.conf.template
        target: /etc/nginx/templates/default.conf.template
      - type: bind
        source: $PWD/cert.pem
        target: /etc/nginx/ssl-cert.pem
      - type: bind
        source: $PWD/key.pem
        target: /etc/nginx/ssl-key.pem
    ports:
      - 80:80
      - 443:443

networks:
  default:
    ipam:
      config:
        - subnet: 172.28.0.0/16
```

So I guess instead of using my PR, it would also work by adding the `rpcuser` and `rpcpassword` to the bitcoind command line (which are deprecated though and might be removed in favor of `rpcauth` – which has to be calculated), and then setting the env variables (`BTC_RPC_USER`, `BTC_RPC_PASSWORD`, `BTC_RPC_HOST`, `BTC_RPC_PORT`, `BTC_RPC_PROTOCOL`) for specter-desktop, which would slightly complicate the setup, but not require adding more code to specter-desktop, which might be preferable.

k9ert:
> I guess this is done for simplicity, so that specter either a) connects on localhost with auto-detection, or b) 
> connects to a given host that requires all parameters (host, port, user, password).
Yes exactly! It's a bit lazy, we could definitely make that a bit more clever: If one is set, then defaults for the others are used, until the others are set as well.

> might be removed in favor of rpcauth
That's definitely a point. I'm not so familiar with rpcauth, though but i'm still convinced that one should decide whether to separate in containers or not. Avoiding the calculation looks like not doing either, or.

nioncode:
So I thought about this a bit more and played with the settings on the regular specter-desktop version (i.e. without my patch).

Without this patch, the following has to be done:
- run bitcoind with `rpcuser` and `rpcpassword` being set
- run specter-desktop with the same credentials in `BTC_RPC_USER` and `BTC_RPC_PASSWORD` env vars, and additionally set the `BTC_RPC_HOST`, `BTC_RPC_PORT`, and `BTC_RPC_PROTOCOL` variables
- if you are on testnet, you have to take care to set a different `BTC_RPC_PORT` depending on whether you run bitcoind on testnet or mainnet
- setting the variables will probably be taken care of with an `.env` file in docker-compose, so you have to be careful to not check this into version control

With my patch, the following has to be done:
- run specter with `--bitcoin-rpc-host` command line parameter
- mount the bitcoin data volume into the specter container (possibly even read-only), auto-detection takes care of the rest
- no credentials are part of the docker setup, every run of the container will create a new authentication cookie and everything will just work


From a user perspective, I feel like my approach is easier to configure, less error-prone and avoids potentially leaking credentials through the `.env` file. I think using the auto-detection is a nice way of connecting to bitcoind and if this mechanism could be preserved while overriding `localhost` with another host, then this would be great.

Since you don't seem to be against the idea in principle, maybe the following would be a better way to achieve the desired effect:
- do not use the new command line option
- instead, if `BTC_RPC_HOST` is set, simply use that instead of `localhost` when using auto-detection
- if all env vars are set use a connection that is exclusively specified with the env vars, just like it is now

k9ert:
> instead, if BTC_RPC_HOST is set, simply use that instead of localhost when using auto-detection

But that doesn't solve the main issue that you have: setting the password.

The general thing here is that i don't get for whom (other then you personally) this feature is for. See also the docker-compose-discussion where we discussed which person would potentially use this feature: A developer? An End-User?
User stories help here as they force you to the format "As someone of the role X i want to easily ...".
So for this change, we need to pin down the real-world-problem a bit more. See also:
https://hintjens.gitbooks.io/social-architecture/content/chapter4.html

nioncode:
> But that doesn't solve the main issue that you have: setting the password.

The password will be taken from the bitcoind cookie file as part of the auto-detection just like it is now. There is nothing more to do for the user.

> The general thing here is that i don't get for whom (other then you personally) this feature is for. See also the docker-compose-discussion where we discussed which person would potentially use this feature: A developer? An End-User?

This would make setting up a bitcoind + specter-desktop combination very easy for end-users that prefer to use docker (e.g. on a home server / NAS / single board computer, where using docker is usually very easy, but installing binaries directly might not be). The containers are separated from each other, but can easily communicate with each other through the default docker network of the compose file, and credentials are automatically detected and don't have to be provided. You can just take the docker-compose file as it and you are up and running without leaking your credentials if you copy-paste the docker-compose file around.

I outlined the concrete problem in my previous post. Obviously it is possible to make this work right now without my patch, but it just takes more configuration and is prone to leak credentials if the user is not careful. The same question ('which real-world problem does this solve') could be asked for why the environment variables even exist, since you can just go to the config file and change the values there, it is just more effort to do so.

nolim1t:
I think ENV variables do have its merit in docker setups. But I could probably set up an entrypoint in [this project](https://github.com/lncm/docker-specter-desktop) to pass on the env variables to as CLI switches as well if that will solve a lot of your issues. So the maintainers can focus more on their goals rather than get sidetracked.

Docker setups don't have a lot of place for desktops (particularly windows and mac where the implementation is shitty), but on IoT devices (such as the Pine64 Phone or Tab) or home linux servers such as the RPI which is gaining in popularity for full node projects for ease of building modules it might actually be useful.

k9ert:
I also think that docker (-compose) is for integrators, not so much end-users. For them, some more integration-work is a good trade-off to get a bit more security or better architecture.

> The same question ('which real-world problem does this solve') could be asked for why the environment variables even exist, since
> you can just go to the config file and change the values there, it is just more effort to do so.

Imho the justification is that the one who's specifying the ENV-vars might not be the one who controls the config-file (which is partially the one who has admin-rights on specter) and also not the end-user. So effectively it's a never-ending splitup of deciding for everyone whether to trust or to verify.

That was my intention for the Env-vars in the first place: The Env-var-operator (the one who starts the binary or is in control doing that indirectly) can suggest his trust and the admin of specter can decide to not trust but verify (turn off auto-detection and use his own node).

nioncode:
In what scenario has the operator of the ENV vars no control over the config file of specter?

I want to raise a separate question: is there any benefit of giving specter access to the bitcoind data directory except of the auto-connection? If there is, then I think it makes sense to mount the bitcoind data directory to the specter container regardless of my patch. And if the data directory is mounted anyways, then it makes sense to allow overwriting `localhost` in the auto-detection. If you would prefer to do this over a single env var, as proposed by me before, instead of a command line argument, please let me know.

To me it feels like you are strongly against this PR no matter what solution I propose, although I think it makes for a better user experience when one wants to use docker to set up a bitcoind + specter combination (whether this is a integrator that builds a box for end users or simply an end user like me who wants to use specter on a constrained device), since you don't have to fiddle with bitcoin rpc credentials at all.

k9ert:
> In what scenario has the operator of the ENV vars no control over the config file of specter?
I guess "control" is the wrong word. The admin can, at any point in time change the config which node is used. Therefore the user has "control" over that.

> is there any benefit of giving specter access to the bitcoind data directory except of the auto-connection?
As said, i only see drawbacks, mostly from a security point of view. The "solution" looks rather like a dirty hack to me.

> To me it feels like you are strongly against this PR no matter what solution I propose
It's the mounting of the data-dir which concerns me.

In the other PR, there is a clearer goal formulated but it seems to be a very similiar use-case. I'd suggest to close this PR and followup on the other PR. The use-case is definitely valid but we should really still respect docker-best-practices.

nioncode:
I'm closing this, since mounting the bitcoind volume seems to be not suggested (although it is the same as running specter on the same system) and use the env variables in my docker-compose file now.

nioncode:
Actually, I want to try one more time to convince people here that the feature is worth having, since I just found out that `bitcoin-cli` is now broken with the env variables, since it cannot get the authentication from the cookiefile anymore (since the cookie file is not created with hardcoded rpc user + password).

To fix bitcoin-cli, I'd need to:
- either a) move the credentials from the bitcoind command line arguments into `bitcoin.conf`, so that `bitcoin-cli` can parse them from there. This would then mean that the credentials must be specified twice, once in `bitcoin.conf` and once in either specter's env variables, or in specter's config file. Both options are not so nice, since it duplicates the credentials in two places, and it requires to keep them updated / in sync.
- or b) pass `rpcuser` and `rpcpassword` on the bitcoin-cli command line for each invocation, which seems equally unpleasant

So to summarize the current issues:
- it is very beneficial to run bitcoind with rpc authentication being setup through the default cookie file, e.g. to support painless interaction with bitcoin-cli
- however, when wanting to use the authentication file (i.e. auto-detection) with specter-desktop, it always chooses `localhost` as rpc host, which does not resolve correctly when bitcoind is running in a separate container
- if there would be a way to simply replace `localhost` with another host (in this case the container name of the container that runs bitcoind), everything would be solved without having to take care about the credentials AT ALL

Only real workaround right now:
- set up rpc credentials through bitcoin.conf + specter config separately, which requires modifying two files and keeping them in sync instead of modifying 0 config files with my patch

As you see the current way of things would require a much more complex setup to get things working smoothly, so I want to kindly ask to revisit your opinion to include this change (or an equivalent change that works through a single env var if you don't like the cli argument) in order to simplify setups and make them more robust and secure (since no credentials have to be configured at all and thus cannot accidentally be leaked when sharing bitcoin/specter config files) – instead of making them less secure as you implied (remember the bitcoin data dir could even be mounted read-only).

k9ert:
> ... so I want to kindly ask to revisit your opinion to include this change ...
It's not only about me. As a good maintainer i might even merge things which i'm not 100% convinced of because a lot of people make the same arguments. So, you should also convince others but i think there is a reason that we see so less engagement in this PR.

Imho it boils down to two things:
* It's not suprising that credentials need to be maintained in two places because that's why they are credentials. One is setting it and the other part part must know it to get success. For sure there is overhead involved for such kind of setups.
* Docker makes it possible to separate things. Merging them again via docker-magic is misusing docker.

Sorry, i hope you understand my reasoning here and i hope you could potentially engage in the other docker-things going on in the project.

k9ert:
Closing the PR. Outdated and no more engagement.

PR #485, opened 2020-10-05
Minor spelling change

mflaxman:
I'm testing doing this all via github's web tool, it's unclear how this will affect git commit history (I believe everything will be automatically squashed down nicely).


k9ert:
The annoying thing here is that PR-builds are only taken from whitelisted people (scam-detection-code trom travis-ci is detecting crypto-mining-code in the project but base the block on the person who did the PR) and you haven't so far. Will probably write yet another mail to travis in order to get the people whitelisted which are making more frequent changes.

Issue #476, opened 2020-10-03
Verify Multisig Receive Address on Coldcard Fails


liquidso:
### Problem Description:
Multisig receive address fails to display on Coldcard

### To reproduce:

- 1. On Specter-Desktop: Your Wallet > Receive > Display address on device:
- 2. Verify address on your Coldcard but fails with below error msg.

Traceback (most recent call last):
  File "/home/liquid/.local/lib/python3.8/site-packages/cryptoadvance/specter/util/json_rpc.py", line 35, in jsonrpc
    response["result"] = method(**request["params"])  # dict -> **kwargs
  File "/home/liquid/.local/lib/python3.8/site-packages/cryptoadvance/specter/helpers.py", line 38, in wrapper_fn
    return fn(*args, **kwargs)
  File "/home/liquid/.local/lib/python3.8/site-packages/cryptoadvance/specter/hwi_rpc.py", line 214, in display_address
    raise Exception(status["error"])
Exception: Multisig descriptor must include all pubkeys

Environment:

- Python 3.8.5
- Specter Version: 0.8.1
- Linux Distro: MANJARO

mflaxman:
Confirmed I'm getting the same on 98bbcf28455504a1335597946c74b7878d8b6458 on python 3.8.5 MacOS 10.15.6.
```
[2020-10-08 17:58:23,395] INFO in _internal: 127.0.0.1 - - [08/Oct/2020 17:58:23] "POST /hwi/api/ HTTP/1.1" 200 -
Traceback (most recent call last):
  File "/Users/mflaxman/workspace/specter-desktop/src/cryptoadvance/specter/util/json_rpc.py", line 35, in jsonrpc
    response["result"] = method(**request["params"])  # dict -> **kwargs
  File "/Users/mflaxman/workspace/specter-desktop/src/cryptoadvance/specter/helpers.py", line 38, in wrapper_fn
    return fn(*args, **kwargs)
  File "/Users/mflaxman/workspace/specter-desktop/src/cryptoadvance/specter/hwi_rpc.py", line 214, in display_address
    raise Exception(status["error"])
Exception: Multisig descriptor must include all pubkeys
```

mflaxman:
This is a confusing one...

[...]

PR #496, opened 2020-10-09
add pip-compile and hash support for security/reproducibility

mflaxman:
I'm running into a strange bug in https://github.com/cryptoadvance/specter-desktop/issues/476, and I suspect that broken versioning is at play (though I'm not certain) so I decided to add this to specter, especially since it is very good for security. Read more about `--require-hashes` [here](https://lil.law.harvard.edu/blog/2019/05/20/improving-pip-compile-generate-hashes/).

I updated the steps for pip to first install from `requirements.txt` and then second `install -e .` as this [seems to be the best practice for validating those hashes on build](https://stackoverflow.com/questions/46835953/hash-checking-in-setup-py-install-requires) (kind of hackey).


mflaxman:
Foiled by my own `black`ifier being down! This does indeed pass locally:
```
$ black . --check
All done! ✨ 🍰 ✨
64 files would be left unchanged.
```

mflaxman:
Black is now fixed in https://github.com/cryptoadvance/specter-desktop/pull/497

Once that is merged this should pass.

mflaxman:
Black passes again, this is ready to merge! (Don't know how to get TravisCI to run)

k9ert:
Thank you very much! This is really awesome PR. I didn't knew that this is possible and also haven't understood all implications but this is definitely very usefull. I tested the branch and the pip-installation worked. So this LGTM, i'd merge.
@ben-kaufman  and @stepansnigirev please confirm, at least one of you!

stepansnigirev:
Looks good to me!

stepansnigirev (approved):

PR #502, opened 2020-10-11
rework discovery of pip installed version

ZeroPointEnergy:
I ran across an issue when attempting to create a system wide installation with a package manager that the version could not be identified correctly.

This patch reworks the version discovery and used the importlib_metadata library to find the installed version (https://importlib-metadata.readthedocs.io/en/latest/using.html#using). This is part of python >=3.8, so it is only temporarily a new dependency.

To find the latest version I just made a direct request to the pypi API, which is probably also what pip does, just that we don't have to parse the CLI output and get an easy to parse json.


ZeroPointEnergy:
Fixed the formatting issue. Had to redo the whole commit, could not get git to preserve the content of the merge commit on a rebase for some reason. Sorry

k9ert:
Always a pleasure to merge a PR which removes more code than it creates :-).

Issue #492, opened 2020-10-08
Improvements for reverse-proxy setups


max302:
Hello, 

I've been trying my hand at integrating specterd into BTCPayServer, I'm almost there but I've run into a bind with regards to the reverse proxy setup. 

The is probably a very hacky way of working this out with a complex setup of URI rewrites within NGINX, but I'm guessing that as adoption increases, other people will run into my problem. 

The problem itself is that linking to external resources (images, stylesheets) seems to be inconsistent, which breaks proxy actions; some are linked with trailing slashes, some are not. Here's an example:

icon.png in div `footer` (does not work): 
`<img src="/static/img/icon.png" style="width: 35px;">`

icon.png in div `messages` (works):
`<img src="static/img/icon.png" width="40px" style="vertical-align: top; margin-right: 20px;">`

I see a few ways of fixing this, but since I'm not familiar with Workzeug, I thought I'd run it by you guys. 

1. Fix the markup to use relative links for all resources
2. Configure Workzeug to serve the application at a specific base URI, which would reflect what I'm trying to use in BTCPayServer. (domain.com/specter -> 0.0.0.0:/specter/)
3. Document a full list of endpoints and asset locations to help people with their proxy setups. 

My docker image is [https://github.com/max302/specter-docker](url). I'm using the specterd binaries. 

Let me know what sounds like the simplest/best, hopefully I can help out and propose a PR.

k9ert:
Can you please have a look here:
https://github.com/cryptoadvance/specter-desktop/pull/504

Would that help here as well?

max302:
This is exactly what I need, @kexkey's use case is almost identical to mine!

PR #504, opened 2020-10-11
Support for url context paths

Kexkey:
Hi guys, first of all this is an awesome project!  Thanks for your amazing work!

I'm not a Python dev, so I did my best here but not sure it's the best solution and am relying on you to make it better.

I've been working on integrating Specter to Cyphernode.  The problem I was facing was Specter doesn't support having a non-root context path.

If we run a reverse proxy in front of specter with a url path forward (X-Forwarded-Prefix), we have problems.  For example, when my specter instance is running at `https://mydomain/specter/`, the application will replace the URLs with `https://mydomain/`.

In this PR, I use Werkzeug's ProxyFix middleware to do the context path substitutions.  I had to change several references to URLs in the app so it uses `url_for` instead of absolute paths, as well of course as to add the WSGI middleware to the Flask app object.

The tests passed and I made a Trezor work with a remote instance running behind a proxy (`https://whatever/specter`).  I didn't change the local hwibridge instance's url path (`./specterd --hwibridge`) because I assumed it would always run locally... but maybe it would be a nice to have.

Let me know if it makes sense and if there's a better way to do it.  I'll be happy to learn and change the PR.  Thanks again!


k9ert:
Hey, thank you very much for this PR, that really sounds very reasonable. I'm personally a cyphernode fan and tested it for some time, although i'm not using it, it's very good to have something like that for bigger-business-requirements. Thank you for that project as well!
I think we need to manually test that PR a bit better, give me some more time and i'm sure i have feedback or can merge it until end of the week.

k9ert:
BTW, cyphernode is mostly built from docker-containers, right? Have a look at the excellent work which weforked here:
https://github.com/cryptoadvance/docker-specter-desktop

https://hub.docker.com/r/cryptoadvance/specter-desktop

Kexkey:
Hi @k9ert!  Awesome, I understand that it needs some time to check everything, since the URLs this PR affect are all over the place.  :)  I couldn't myself test everything because I don't have everything needed to do so.

Thanks for the kind words about Cyphernode, it's very appreciated!  Yes, Cyphernode is fully dockerized (except the installer, of course!).  I wrote my own Dockerfile for Specter, based on Alpine.  The Alpine images are usually lighter than other OS images.  It works pretty well.  I build all the Cyphernode images for amd64, arm32 and arm64 (using Docker's manifests) and make sure all is compatible.  I also made a glibc-based Alpine image to avoid using musl, when needed.  Sometimes challenging, but it gives smaller image footprint (RAM and disk space).  I'll push the Dockerfile this week and let you know so you can have a look... it's straight forward though.

BTW did you try using the USB ports from Docker on Mac?  Not sure it's possible without using VirtualBox...

Maybe we should move this conversation somewhere else if you want :)

Kexkey:
BTW I'm looking at your docker image and it's very good!!  Good work!!  Hi-Five!!

max302:
Critical support for this PR on my part. This would fix issue #492 and also allow easy integration to BTCPayServer.

k9ert:
Merged, thank you very much again! Awesome PR.

k9ert:
> allow easy integration to BTCPayServer.

@max302  can you elaborate on that part? How would this improve integration with btc-pay-server?

PR #510, opened 2020-10-13
Bugfixes/479

### LINKED ISSUE #479: 0.8.1 upper case address cause exception.
I am getting exceptions "KeyError :  BC1....."   when trying to create a unsigned transaction.  If I lower case the address first it works fine. 

I am scanning from Samurai wallet.

k9ert:
fixes #479


k9ert:
Yeah, should have known that. Only bech32-addresses are case insensitive.
https://en.bitcoin.it/wiki/Address
Removed the intrusive part and only fixing the key-error.

stepansnigirev (approved):

PR #514, opened 2020-10-14
Hot wallet advanced configurations

ben-kaufman:
This PR adds some more advanced configurations for using Bitcoin Core hot wallet, including:

- BIP39 passphrase support (close #490)
- Custom derivation paths support (close #498)
- Custom range for import of addresses (Bitcoin Core 0.21 should make it obsolete, eventually...)
- Allow adding more keys to an existing hot wallet.
- Improves terminology to distinguish between wallet file encryption password and BIP39 passphrase.

PR #521, opened 2020-10-16
add support for non-hardened derivation for hot storage

stepansnigirev:
using rpc to convert private key to public key.
should help with #514


ben-kaufman (approved):
Awesome, tested with http://iancoleman.io/bip39/ , works great

PR #524, opened 2020-10-17
fix for proper namespace package cryptoadvance

k9ert:
In #69 we have created the proper layout and infrastructure to do pip-releases. We decided to have "cryptoadvance.specter" as the package name and used src/cryptoadvance/specter as the directory-layout in order to maybe later have other packages within the namespace "cryptoadvance".

However, i didn't properly did this as a "namespace" package. The [dot is fine](https://www.python.org/dev/peps/pep-0423/#use-a-single-name) for the package-name but for a proper [native namespace-package](https://packaging.python.org/guides/packaging-namespace-packages/) we need to remove the __init__.py from the cryptoadvance-package and help the setup-procedure a bit with finding the package.

Not doing that results in the inability to have any other subpackage below "cryptoadvance". I've tested this on a clean checkout and the package-creation (and installation in a virtualenv) works.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,6 @@
 from glob import glob
 
-from setuptools import find_packages, setup
+from setuptools import find_namespace_packages, setup
 
 with open("requirements.txt") as f:
     install_reqs = f.read().strip().split("\n")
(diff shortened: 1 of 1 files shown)


stepansnigirev:
tested, works. merging.

PR #530, opened 2020-10-20
make electrum watch-only export

mflaxman:
https://github.com/cryptoadvance/specter-desktop/issues/212#issuecomment-699587232

Note that this only works for multisig (single sig would require adding a crypto library).


ben-kaufman (approved):
Tested, works :)

stepansnigirev:
Awesome, thanks!
Why do we need a crypto library for single sig?

mflaxman:
I was confused at the relationship between devices/keys and was wrong. Single-sig now works and the code is a bit easier to read.

Give this a shot and see if you like it, I think it's a nice addition (and I want to feature it for out of band receive address verification).

stepansnigirev:
Yes, I think it's a great feature. Let me some time to test it tonight and we can merge :)

Issue #376, opened 2020-09-10
rename devices


stepansnigirev:
we can rename wallets but not devices. would be nice to rename devices as well.

PR #534, opened 2020-10-21
rename devices

stepansnigirev:
close https://github.com/cryptoadvance/specter-desktop/issues/376


ben-kaufman (changes requested):
Looks good, just one thing, we should change the `devices_signed` in out psbt format to use `alias` instead of name. Currently in case you rename while having an existing psbt it could "loose" existing signature of the renamed device.
These areas need to change:
https://github.com/cryptoadvance/specter-desktop/blob/f68904de2a76079e0f6d30da84c2ea3c821d8fdb/src/cryptoadvance/specter/controller.py#L190
https://github.com/cryptoadvance/specter-desktop/blob/f68904de2a76079e0f6d30da84c2ea3c821d8fdb/src/cryptoadvance/specter/controller.py#L1390
https://github.com/cryptoadvance/specter-desktop/blob/f68904de2a76079e0f6d30da84c2ea3c821d8fdb/src/cryptoadvance/specter/wallet.py#L350
https://github.com/cryptoadvance/specter-desktop/blob/f68904de2a76079e0f6d30da84c2ea3c821d8fdb/src/cryptoadvance/specter/wallet.py#L1127
https://github.com/cryptoadvance/specter-desktop/blob/cda5851da09582c011c6f193bb26a591d7ff4414/src/cryptoadvance/specter/templates/wallet/send/sign/wallet_send_sign_psbt.jinja#L155

PR #541, opened 2020-10-22
add page loader

stepansnigirev:
In the app we don't have the "loading" UI, so it's hard to say if the page is loading or just UI freezes.
This PR adds a page loader bar on top of the screen.

We could also show it after timeout, not right away, not sure. What do you think?

Screenshot 2020-10-22 at 16 00 25


ben-kaufman:
Looks good! I think we could give a tiny timeout before activating it, but doesn't matter so much I guess.

stepansnigirev:
ok, added 500ms delay

ben-kaufman (approved):
Works great! I think the timeout could even be a bit shorter, like 200 or so, but like that is good too if you prefer that.

PR #543, opened 2020-10-22
change wallet export to use alias instead of name

mflaxman:
This is useful because you get filenames like `my_wallet.json` instead of `My Wallet.json`.

These are easier to pass around via CLI tools.


stepansnigirev:
Good point, thanks!

PR #548, opened 2020-10-24
add tx hash to psbt filename

mflaxman:
It is useful when managing PSBTs to have more precise names like `base64_00451c6a635872149eedf1bc1e521081ba0dc0f3185d97062e4e2c49af1c3c1d.psbt` instead of `base64.psbt`.

As you do more transactions and dowload more files, `base64.psbt` is likely to turn into nastier things like `base64(1).psbt` (which may have no relation to `base64.psbt`).


stepansnigirev:
thanks, merged!

PR #556, opened 2020-10-27
Features/linux binary hashing

k9ert:
This is some random bits and pieces on the build/release side of this story:
* it skips the test-stage in the gitlab-build if it's not running in the cryptoadvance-gitlab-project. As we're mostly using gitlab for releasing, if people are running gitlab-builds on their own gitlab-projects, they are probably developing on the release-process and therefore have an urgent need for fast feedback.
* Also the release-parts which are uploading to the releases-page of gitlab are now automatically uploaded to the github-project this build is running on.
* The build-ci-script is now failing (set -e) as soon as some issue occured and it also has some more output to orient better in build outputs.
* There  are now sha256-sum created for both electron-apps and the linux-specterd and the pip-package. The windows-specterd is missing, though.
* We're now creating coverage-reporting for tests

Here is how the release-page looks like after the test-build finished:
image


k9ert:
I think this is ready for review. I've also added some signing code which is not yet working but which also doesn't hurt so far. In short:
* A gpg-key is accessable on the gitlab-runner
* It has a passphrase which is only available on the project which is building the job
* If the gpg-key is available, it's signing the Hashsum-file and uploads that to the release-page as well.

ben-kaufman (approved):
Tested, works great!

PR #558, opened 2020-10-28
add threaded checker

stepansnigirev:
Checks happen in a separate thread when possible, once in 10 minutes (~ every block).
It's possible to make the updates more often by changing the `specter.checker.period` or resetting `specter.checker.last_check` to zero - in this case it will do the check within a second.

### DIFF
--- a/src/cryptoadvance/specter/controller.py
+++ b/src/cryptoadvance/specter/controller.py
@@ -50,9 +50,6 @@ from binascii import b2a_base64
 from .util.base43 import b43_decode
 from .util.tor import start_hidden_service, stop_hidden_services
 from stem.control import Controller
-import logging
-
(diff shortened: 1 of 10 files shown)


ben-kaufman (approved):
Works great, super fast now! :)

ben-kaufman:
Oh wait one problem I see now is that after send, the transaction screen doesn't update immediately. I guess we should trigger check (for the specific wallet at least) when sending a transaction.
Also another improvement I think would be to query the block height let's say every minute (or less even maybe?) and then run check every time the block height returned changes. So it'd be either every 10 minutes or every 1 minute if the block arrived early (we should keep the 10 minutes as max to make sure unconfirmed transactions are displayed as well eventually.
What do you think?

PR #560, opened 2020-10-28
Allow manual specterd upload

ben-kaufman:


k9ert:
This somehow didn't work for me. I downloaded 0.7.2 and tried to upload the tarball and also the unpackaged specterd.
The hash got updated and i manually changed the version but after clicking "save" nothing happens. Also so the appsettings.json did not change.

ben-kaufman:
@k9ert if nothing happens after clicking save it's almost certainly a JS error, can you toggle on the developer tools (should be in the menu bar) and check if after clicking save it gives there some error in the console?

PR #567, opened 2020-10-31
Allow build-unix.sh to work on Arch Linux

dmp1ce:
Changes shell script and python script to allow `build-unix.sh` to work on Arch Linux based distros.

### DIFF
--- a/pyinstaller/specterd.spec
+++ b/pyinstaller/specterd.spec
@@ -14,7 +14,17 @@ elif platform.system() == 'Linux':
     if platform.processor() == 'aarch64': #ARM 64 bit
         binaries = [("/lib/aarch64-linux-gnu/libusb-1.0.so.0", ".")]
     else:
-        binaries = [("/lib/x86_64-linux-gnu/libusb-1.0.so.0", ".")]
+        candidates = [
+                "/usr/lib/libusb-1.0.so.0",
+                "/lib/x86_64-linux-gnu/libusb-1.0.so.0",
+                "/lib/aarch64-linux-gnu/libusb-1.0.so.0",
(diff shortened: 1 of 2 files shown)


pyinstaller/specterd.spec
```diff
@@ -14,7 +14,10 @@ elif platform.system() == 'Linux':
     if platform.processor() == 'aarch64': #ARM 64 bit
         binaries = [("/lib/aarch64-linux-gnu/libusb-1.0.so.0", ".")]
     else:
-        binaries = [("/lib/x86_64-linux-gnu/libusb-1.0.so.0", ".")]
+        if distro.name() == "Arch Linux":
+            binaries = [("/usr/lib/libusb-1.0.so.0", ".")]
+        else:
+            binaries = [("/lib/x86_64-linux-gnu/libusb-1.0.so.0", ".")]
```
stepansnigirev, line 20: I suggest doing it slightly differently:
to set up a list of candidates where the library can live in different linux distros, and try to find one:
```suggestion
        candidates = [
                "/usr/lib/libusb-1.0.so.0",
                "/lib/x86_64-linux-gnu/libusb-1.0.so.0",
                "/lib/aarch64-linux-gnu/libusb-1.0.so.0",
                "/lib/arm-linux-gnueabihf/libusb-1.0.so.0",
        ]
        binaries = []
        for p in candidates:
            if os.path.isfile(p):
                binaries = [(p, ".")]
                break
```

stepansnigirev:
Looks good, just suggested one small tweak to make it more distro-agnostic.
See above.

dmp1ce:
Updated PR with suggestion.

dmp1ce:
Removed distro import.

stepansnigirev:
Thanks! Merging.

PR #574, opened 2020-11-03
bugfix: Specifying python version better

k9ert:
python 3.6 is no longer supported.


setup.py
```diff
@@ -37,5 +37,5 @@
         "Operating System :: OS Independent",
         "Framework :: Flask",
     ],
-    python_requires=">=3.6,<3.9",
+    python_requires=">=3.7,<=3.8",
```
stepansnigirev, line 40: why no support for 3.9?

stepansnigirev: I wasn't able to find a `<=` example in the [pep docs](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires), are you sure it will work?

k9ert: No, i'm not sure. But it doesn't matter so u've reverted that part.

k9ert:
We could potentially support 3.9 as well but i'd suggest to test that explicitely.

stepansnigirev:
I think at the moment we can't support 3.9 because HWI can't be installed on 3.9 - they have `<3.9` restriction for some reason.

PR #580, opened 2020-11-07
Support Specter file format in import wallet screen

ben-kaufman:
Allow importing specter wallet backup format in import wallet screen


src/cryptoadvance/specter/controller.py
```diff
@@ -718,8 +718,20 @@ def new_wallet(wallet_type):
...
+                "specter_format" if "recv_descriptor" in wallet_data else "account_map"
+            )
+
+            if backup_type == "specter_format":
+                name_key = "name"
+                descriptor_key = "recv_descriptor"
+            elif backup_type == "account_map":
+                name_key = "label"
+                descriptor_key = "descriptor"
+
```
stepansnigirev, line 732: `backup_type` variable is not used anywhere else so let's simplify a bit:

```suggestion
            # specter format
            if "recv_descriptor" in wallet_data:
                wallet_name = wallet_data.get("name", "Imported Wallet")
                recv_descriptor = wallet_data.get("recv_descriptor", None)
            else:
                wallet_name = wallet_data.get("label", "Imported Wallet")
                recv_descriptor = wallet_data.get("descriptor", None)

```

Also `wallet_name` assignment below is not needed (moved to the `if` case) and `recv_descriptor` variable can be used instead of wallet_data[descriptor_key]

PR #614, opened 2020-11-17
Add PDF backup

ben-kaufman:


stepansnigirev:
@k9ert the checker test is failing with 50% probability even when there are no changes in it.

PR #623, opened 2020-11-18
isascii for py3.6

stepansnigirev:
We don't officially support python3.6, but maybe we could make it working.
@k9ert do you remember what issues aside from isascii did we have with python3.6?

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ The easiest way to run Specter Desktop is by installing the Specter Desktop app,
 With this method, all you need to do is just download the right file for your operating system and install it like a normal desktop app.
 
 ### Installing Specter from Pip
-* Specter requires version 3.7 or 3.8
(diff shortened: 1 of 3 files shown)


k9ert:
If we really want to make that work, we should double all tests and test them with 3.6. But is that really worth it? Maybe we should rather do that for 3.9?

stepansnigirev:
we can't do python3.9 because HWI doesn't support it.
3.6 is the default python version on older ubuntu versions, and it is supported in HWI so it would be nice to support it as well.

stepansnigirev:
I just ran the tests on python3.6, all green.
I also tested the flow for simple signing and stuff - also seem to work, so we can keep 3.6 as the minimal version for now.
Maybe we should always run the tests on 3.6? But build on 3.7 or 3.8.

PR #641, opened 2020-11-20
Fix submit non ediable xpubs

### LINKED ISSUE #640: New device wizard problems
Trying with Specter-DIY simulator, adding new device:
xpubs are populated correctly
when I click "Add revice button" I see an error that "xpub name must not be empty"
Checking request.form I see that xpubs are not present in the form, only purposes and derivations.

Another problem is on the styles side. If the screen is slightly smaller than fullscreen (like in the app for example), the xpubs table doesn't fit in the screen anymore

ben-kaufman:
Fix #640

### DIFF
--- a/src/cryptoadvance/specter/controller.py
+++ b/src/cryptoadvance/specter/controller.py
@@ -1077,7 +1077,7 @@ def device_setup_wizard():
                 purpose = request.form.get(
                     "xpubs-table-row-{}-purpose".format(i), "Custom"
                 )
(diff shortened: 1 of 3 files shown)


stepansnigirev:
Looks good!
Thiking about importing electrum and other keys, I don't think we should have the xpubs editable - because it will cause problems if someone enters xpub with a wrong derivation path there, or without derivation path.
We should add "Paste xpubs" button instead, similar to "Paste signed transaction" - this button should appear only on Electrum and Other devices and populate / create corresponding entry in the table.

PR #692, opened 2020-11-23
Add tx data popup

ben-kaufman:
Show popup with detailed transaction data when clicking on the transaction hash.

I think the next thing with it could be allowing txid labeling or something like that. Maybe would be also nice to organize it into some general details section and advanced section, but not sure what should go to which section.


k9ert:
Seems to have difficulties with coinbase-transactions:
"TypeError: rawtx.vout[i].addresses is undefined"

But looks good otherwise!

PR #698, opened 2020-11-26
Util testing

moneymanolis:
As you can see the base58 tests fail after copying the embit tests due to embit returning strings and Specter util returning bytes. How should we deal with that? Change embit or change potential usage of embit? @stepansnigirev


k9ert:
Travis-CI is broken because of abuse-detection. I wrote a mail to support already.

stepansnigirev:
I think it makes more sense if base58 encoding and decoding are converting between strings and bytes, not bytes and bytes.
I suggest utilizing `embit.base58` and `embit.hashes` instead of maintaining separate collection of utils in Specter.

@moneymanolis do you feel confident enough removing base58 encodings from utils, finding places where we use base58 encodings and replacing them with embit versions?
I think we should keep tests on Specter side as well because embit doesn't use pytest (due to hardware limitations) and here we can use all the power of pytest to test stuff.

tests/test_util.py
```diff
@@ -0,0 +1,127 @@
...
+    assert encode_base58_checksum(testcase) == expected
+
+
+### xpub tests
+
+
+def test_convert_xpub_prefix(ghost_machine_xpub):
+    new_prefix = b"\x04\x9d\x7c\xb2"
+    testcase = ghost_machine_xpub
+    # No external source found for expected
```
stepansnigirev, line 125: you can test ypubs and zpubs against https://iancoleman.io/bip39/

k9ert:
> I think we should keep tests on Specter side as well because embit doesn't use pytest (due to hardware limitations) and 
> here we can use all the power of pytest to test stuff.
But we can also test that stuff e.b. with Github Actions over at the other repo?! Sure, pytest is more powerfull but unittest might do the job as well. testing it only on specter-desktop would mean that we would be able to only test releases but the tests should definitely be executed much earlier.

I would prefer to setup testing on the embit-repo (and maybe only additionally here). The tests has been copied over to here just to check for differences in the implementation.

stepansnigirev:
I agree, let's keep tests in embit repo but use unittest module as much as possible there, so I could test with micropython build as well.
We can use pytest where needed, and do github actions too, I will be happy to accept PRs on that.
Both Specter-Desktop and Specter-DIY will benefit from more testing on embit side :)

k9ert:
For the reference which repo we're talking about:
https://github.com/diybitcoinhardware/embit

moneymanolis:
As for the base58 refactoring, yes I can do that. I added two tests, so now the whole usage of the encoding/decoding should be covered with tests in Specter. The embit.base58 encode / decode functions are much clearer IMO. For example, in Specter right now we only have a decode function that returns without checksum, whereas in embit you have both options decode () and decode_check(). I guess it makes sense to then only use the decoding without checksum in Specter if this is helpful, for example whenever you have to manipulate / convert a xpub.

tests/test_util.py
moneymanolis: Got it and incorporated two tests for ypub and zpub conversion. For future reference, when using coleman, one has to first derive the xpubs under the BIP32 tab, convert the version bytes and then compare the results with the account extended pub keys under the BIP49 and BIP84 tabs.

moneymanolis:
Descriptor tests work in process. Base58 functions are refactored from embit with additional tests of key.py since the Key class uses those functions.

tests/test_util_descriptor.py
```diff
@@ -0,0 +1,291 @@
...
+        desc.derive(1).serialize()
+        == "sh(wsh(sortedmulti(2,03407a711574ae73aa5824f5a66bf4f9a9f49dd274407eb4c27d996019cf4a6552,0376166abb71efb6c9a497a64c1b24c484c29b8a4219a526737e1f370768b1bbe8,03a1427c178f0b1cd679464c4c90444bcfd775d5edb89cf33828170e8be008d921)))#j54e69dq"
+    )
+    # Pubkeys sorted
+    # Verified with iancoleman
+
+    assert desc.address(5) == "2NGamwat67EUkABbkY1HLYjQW3oumUcThnV"
+    # Verified with bitcoin-cli deriveaddresses
+
+    assert desc.address(5, "main") == "388fc825v9R6Ev8BKodXQMFumQRe7C8SZ5"
```
moneymanolis, line 272: I got a different address with bitcoin-cli, but can't rule out a mistake with the xpubs. Would be good, if someone else could check it.

src/cryptoadvance/specter/util/descriptor.py
```diff
@@ -171,11 +187,19 @@ def parse(cls, desc, testnet=False):
...
+            if int(multisig_M) > multisig_N:
+                raise SpecterError(
+                    f"Multisig threshold cannot be larger than the number of keys. Threshold is {int(multisig_M)} but only {multisig_N} keys specified."
+                )
         else:
             keys = [desc.split("(")[-1].split(")", 1)[0]]
 
         descriptors = []
         for key in keys:
+            origin_fingerprint = None
```
moneymanolis, line 199: @stepansnigirev that fixed the "origin propagation" problem

stepansnigirev: Awesome! Thanks!

tests/test_util_descriptor.py
stepansnigirev: Strange, running the same code returns `3R2Zsqx4VmyPxPyCrsfTvnREqThbhjFG4N` for me.

src/cryptoadvance/specter/util/descriptor.py
```diff
@@ -171,11 +187,19 @@ def parse(cls, desc, testnet=False):
                     keys.sort(key=lambda x: x if "]" not in x else x.split("]")[1])
             multisig_M = desc.split(",")[0].split("(")[-1]
             multisig_N = len(keys)
+            if int(multisig_M) > multisig_N:
```
moneymanolis, line 190: This addition ensures that no descriptor is parsed where less keys than the threshold are provided.  @stepansnigirev

tests/test_util_descriptor.py
moneymanolis: Fixed as we discussed, could now produce the correct address that bitcoin-cli gives, see latest commit.

k9ert:
This looks good to merge. Apart from testing improvements, there is one bugfix in here which is even critical. Re-requesting Stapan's review.

PR #712, opened 2020-11-28
Cypress Frontend-testing

k9ert:
copied from Development.md:

## Some hints on cypress
Cypress is just awesome. It's quite easy to create Frontend-tests and it's even recording all tests and you can immediately see how it went. So each test-run, the tests are kept for one day (see the ["artifacts-section"](https://github.com/k9ert/specter-desktop/blob/cypress/.gitlab-ci.yml#L53-L58)) and you can watch them by browsing the artifacts on any gitlab-job-page (right-hand-side marked with "Job artifacts").

```
# make sure you have npm on the path
npm ci
# run the tests
$(npm bin)/cypress run
# open the cypress application (to develop/debug/run tests interactively)
$(npm bin)/cypress open
```

The test_specifications which get executed are specified in cypress.json which looks something like this:
```
{
    "testFiles": [
      "spec_empty_specter_home.js",
      "spec_node_configured.js"
    ]
}
```

Each testfile represents a specific state specter is expected to be if running the tests in that file. However currently the states which are required for each file in the test are only created for spec_empty_specter_home.js. The second file simply assumes that the node is configured and is right about that because one of the tests in the first file did it.
That needs to be improved. Not yet sure how. There are three options:
* Using commands similiar to the way a device is a prerequisite in the second file (`cy.addDevice("Testdevice Ghost")`) which takes a lot of time.
* Copying files around in the specter-folder which makes the test brittle
* Using post-requests as described in this (very interesting) [video](https://youtu.be/LcGHiFnBh3Y?t=1367) with the example of "login" (which is necessary for almost any test)

If you're interested in cypress-test-development, make sure to watch the video above, it's really worth it. Also be sure to not miss the way to copy the css-locator from the cypress-IDE: "Click on elements to see a suggested selector". This is for me the reason why test-development is going so fast.


k9ert:
I think it's worth to create more tests but it's already some commits, so let's try to get that in and continue with more specific tests afterwards.

ben-kaufman:
It's not very clear to me how to setup the local environment to run the tests properly. Can you add details of what are the preconditions (like should I have it with an empty `.specter` folder or something, should I run bitcoind in the background etc.?)

k9ert:
Very good point. I even forget to mention that the `spec_empty_specter_home.js`will result in deleting your whole specter-folder without asking, sorry for that.
Yes, you need to run specter-desktop manually, that's not done by running the tests. So i think i will need to make it work very similiar to the pytests:
* The tests take care about starting the specter-server themself. The Developer doesn't have to do that. 
* Also a temporarily .specter folder is used for that instance
* Maybe the regtest-server is also completely automated by the test-framework (using also a different port similiar to pytest).

Guess i have some more work to do here.

src/cryptoadvance/specter/templates/wallet/history/components/total_wallet_balances.jinja
```diff
@@ -9,7 +9,7 @@
     <h1>
         <small style="line-height:30px">Total balance:</small><br>
         <span style="color: #fff">
-            {{ wallet.fullbalance | btcunitamount }}
+            <div id="fullbalance_amount">{{ wallet.fullbalance | btcunitamount }}</div>
```
ben-kaufman, line 12:
```suggestion
            <span id="fullbalance_amount">{{ wallet.fullbalance | btcunitamount }}</span>
```

div causes the BTC to go to new line, should be either a span or include the BTC/ sats in the div.

src/cryptoadvance/specter/templates/wallet/wallets_overview.jinja
```diff
@@ -25,7 +25,7 @@
         <small style="line-height:30px">Total balance:</small><br>
         <span style="color: #fff">
 			{% set fullbalance = specter.wallet_manager.wallets.values() | sum(attribute='fullbalance') %}
-            {{ fullbalance | btcunitamount }}
+            <div id="fullbalance_amount">{{ fullbalance | btcunitamount }}</div>
```
ben-kaufman, line 28:
```suggestion
            <span id="fullbalance_amount">{{ fullbalance | btcunitamount }}</span>
```
div causes the BTC to go to new line, should be either a span or include the BTC/ sats in the div.

ben-kaufman (approved):

docs/cypress-testing.md
```diff
@@ -0,0 +1,69 @@
...
+Now you can run (or open) specifically this test-file via:
+```
+$ ./utils/test-cypress.sh run spec_existing_history.js
+```
+This will restore the snapshot created above and run this test (and all subsequent ones). Opening the cypress app works the same although there you're responsible to be aware that the state is fitting to the spec-file you want to execute (just like in the case of empty state above)
+
+## Develop on tests
+The spec-files mainly select some element on a webpage and then act on them. Mainly `.click()` and `.type("some Text")`. So it's quite helpfull to have unique IDs for all the elements which we want to use in the tests. The cypress-app has a very convenient way of selecting. Make sure you don't miss that.
+
+For specific things they are `cy.tasks` which can be be implemented. Two of the already exist:
```
ben-kaufman, line 66:
```suggestion
For specific things there are `cy.tasks` which can be implemented. Two of the already existing tasks:
```

cypress/integration/spec_node_configured.js
```diff
@@ -0,0 +1,33 @@
...
+        // Download PDF
+        // unfortunately this results in weird effects in cypress run
+        //cy.get('#pdf-wallet-download > img').click()
+        cy.task("node:mine") 
+        cy.get('#btn_continue').click()
+        cy.get('#btn_transactions').click()
+        cy.get('#fullbalance_amount')
+        .should(($div) => {
+            const n = parseFloat($div.text())
+            expect(n).to.be.gte(0).and.be.lte(50)
```
ben-kaufman, line 29:
```suggestion
            expect(n).to.be.gt(0).and.be.lte(50)
```

This check looks wrong, since mining must make it greater than 0 (assuming there are any block rewards.
The mining doesn't seem to work though, as I see through cypress that the balance remains 0

utils/test-cypress.sh
```diff
@@ -0,0 +1,167 @@
...
+    if [ "$DOCKER" != "true" ]; then
+        addopts=--nodocker
+    fi
+    python3 -m cryptoadvance.specter bitcoind $addopts --create-conn-json --config CypressTestConfig &
+    bitcoind_pid=$!
+}
+
+function stop_bitcoind {
+    if [ ! -z ${bitcoind_pid+x} ]; then
+        kill $bitcoind_pid
```
ben-kaufman, line 43: This seems to fail for me for some reason, causing the stop_specter command to not execute. (if I switch their order at least specter stops properly)


I think it does stop the python process, but not the bitcoind itself...

cypress/integration/spec_node_configured.js
k9ert: Agree with the change. I guess my last commit fixed it because after the change, the tests work for me as well.

utils/test-cypress.sh
k9ert: should be fixed with 6e2dcca

k9ert:
Finally at least github is green. Now only catching up with master ...
https://gitlab.com/k9ert/specter-desktop/-/pipelines/228178185

utils/test-cypress.sh
```diff
@@ -0,0 +1,252 @@
...
+EOF
+}
+
+function send_signal() {
+  # use like send_signal <SIGNAL> <PID>
+  # whereas SIGNAL is either SIGTERM or SIGKILL
+  if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+    kill -${1} $2
+  elif [[ "$OSTYPE" == "darwin"* ]]; then
+    kill -s $signal_name $2
```
ben-kaufman, line 50:
```suggestion
    kill -s SIGTERM $2
```

ben-kaufman (approved):

k9ert:
Travis is green, gitlab is green, together with Ben we got it working on MacOS as well. Will merge tonight or tomorrow.
CC @stepansnigirev

PR #717, opened 2020-11-29
refactoring startup and cli

k9ert:
In #712 i need to inject a port and a specter-data-folder into the startup. So i decided to refactor a bit before i include that feature. 
* The cli.py is now seperated in two files, one for server and one for bitcoind
* We're now using a bit more the config.py which uses a class-inheritance mechanism to define config-values. 
* One can inject such configurations by something like `server --config DevelopmentConfig`
* Injecting `server --config CypressTestConfig` results in DATA_FOLDER is `~/.specter-cypress` and PORT is `25444` by default. You can also override that via envVars.
* Because bitcoind does not know the location of the Data-Folder anymore, it needs to use the same (default) configuration


k9ert:
https://github.com/cryptoadvance/specter-desktop/pull/717/commits/a2c91100915994652664b19738fd6da53a37cabe also contains a bugfix. The Checker did not properly went to a 20secs period because getRpc did not return a proper None for a non-working-connection.

As probably the intuitive reaction of a user realizing that core is down is clicking on the "configure Node", we're also calling check now when GET to settings/bitcoin_core.

ben-kaufman:
Would be very useful I think to be able to pass Specter data directory path as CLI argument by itself (ie. python3 -m cryptoadvance.specter server --data-folder PATH/TO/DATA/FOLDER.

Could you please add this too?

src/cryptoadvance/specter/specter.py
```diff
@@ -124,9 +126,9 @@ def __init__(self, data_folder="./data", config={}):
             self.check(check_all=True)
         except Exception as e:
             logger.error(e)
-        self.checker = Checker(lambda: self.check(check_all=True))
+        self.checker = Checker(lambda: self.check(check_all=True), desc="health")
         self.checker.start()
-        self.price_checker = Checker(lambda: update_price(self, self.user))
+        self.price_checker = Checker(lambda: update_price(self, self.user), "price")
```
ben-kaufman, line 131:
```suggestion
        self.price_checker = Checker(lambda: update_price(self, self.user), desc="price")
```
Right now the "price" is passed as the period which breaks the checker,  should be `desc="price"` instead

ben-kaufman (changes requested):

k9ert:
> Would be very useful I think to be able to pass Specter data directory path as CLI argument 
> by itself (ie. python3 -m cryptoadvance.specter server --data-folder PATH/TO/DATA/FOLDER.
Definitely useful. At latest, we'll need it to start from a specific state for the cypress-tests. But it's already in there.

https://github.com/cryptoadvance/specter-desktop/pull/717/files#diff-32ee6a89e06abc3225ec46a442bdbf8312c4f3845778e9d4164867f714e624cfR42-R46

ben-kaufman:
Oh missed that 😅  that's great!

ben-kaufman:
Hmm looking again seems like that's only for bitcoind, not when just running Specter @k9ert

k9ert:
> Hmm looking again seems like that's only for bitcoind, not when just running Specter @k9ert

Well, i decided against that but instead cover the different DATA_DIR via a configuration. So currently, in order to change DATA_DIR you can do one of two things:
* Mention a specific config-class like `--config CypressTestConfig` which might have a specific (other) DATA_DIR configured , e.g. https://github.com/cryptoadvance/specter-desktop/pull/717/files#diff-95c73b137a4d23e3200440a2feee74da79abb3877b53c1af0acfc1a602f7e08a
* You can set a env-variable in order to override that (see above `DATA_FOLDER = os.path.expanduser(os.getenv("DATA_FOLDER", "~/.specter"))`

Would that work for you? Do you have some specific use-case in your mind?

stepansnigirev:
> You can set a env-variable in order to override that (see above DATA_FOLDER = os.path.expanduser(os.getenv("DATA_FOLDER", "~/.specter"))

DATA_FOLDER looks too generic, maybe use something like SPECTER_DATA_FOLDER instead?

src/cryptoadvance/specter/specter.py
```diff
@@ -52,7 +54,7 @@ def get_rpc(conf, old_rpc=None):
         rpc = BitcoinRPC(**conf)
     # check if we have something to compare with
     if old_rpc is None:
-        return rpc
+        return rpc if rpc.test_connection() else None
```
ben-kaufman, line 57:
```suggestion
        return rpc if rpc and rpc.test_connection() else None
```

Should first also check rpc is not not or you may get an error

src/cryptoadvance/specter/cli/cli_bitcoind.py
```diff
@@ -0,0 +1,242 @@
...
+
+        def cleanup():
+            os.remove("btcd-conn.json")
+
+        atexit.register(cleanup)
+
+    signal.signal(
+        signal.SIGUSR1,
+        lambda x, y: mine_2_specter_wallets(
+            my_bitcoind, config_obj["DATA_FOLDER"], echo
```
ben-kaufman, line 188:
```suggestion
            my_bitcoind, config_obj["SPECTER_DATA_FOLDER"], echo
```

Should be `SPECTER_DATA_FOLDER ` I think...

src/cryptoadvance/specter/cli/cli_bitcoind.py
```diff
@@ -0,0 +1,242 @@
...
+        ),
+    )
+
+    if mining:
+        miner_loop(my_bitcoind, config_obj["DATA_FOLDER"], mining_every_x_seconds, echo)
```
ben-kaufman, line 193:
```suggestion
        miner_loop(my_bitcoind, config_obj["SPECTER_DATA_FOLDER"], mining_every_x_seconds, echo)
```

src/cryptoadvance/specter/server_endpoints/settings.py
```diff
@@ -45,6 +45,9 @@ def bitcoin_core():
     if not current_user.is_admin:
         flash("Only an admin is allowed to access this page.", "error")
         return redirect("")
+    # The node might have been down but is now up again
+    # (and the checker did not realized yet) and the user clicked "Configure Node"
+    app.specter.check()
```
ben-kaufman, line 50: Is this needed even if the RPC is set? I know it's good to check anyway but this means potentially significant slowdown for loading the settings page. So if this is needed only if Specter thinks no node is connected maybe we should check if RPC is None and only if it is do it?

src/cryptoadvance/specter/specter.py
k9ert: Good spots! As a result of this, i decided to do more tests especially on the cli_server side which is more customer-facing.

src/cryptoadvance/specter/server_endpoints/settings.py
k9ert: Ok, i've added a check of app.specter.rpc being None beofre doing this.

k9ert:
Ok, so i did some more refactoring with 7c514ed and leveraged the app.config-object a bit more. The idea is to decrease the use of variables (e.g. "port") and use the app.config object to store the configuration. The config-object is also quite handsome to load stuff from the env-vars and so we can completely remove that aspect in the cli_server.py.

src/cryptoadvance/specter/cli/cli_bitcoind.py
ben-kaufman: This is also still needed no? Getting `Key error` crash without this

ben-kaufman:
Looks good! For some reason `--no-mining` seems to just crash for me. Are you seeing this issue too?

ben-kaufman (approved):
Looks good!

PR #723, opened 2020-11-30
Create specterd.service

k3tan172:
specterd.service file for those running specter-desktop on a linux server. Ensures specterd starts on boot directly after bitcoind. uses host 0.0.0.0 for all IP's to connect to it and caters for https to ensure users are able to use webcams for QR code scanning in the browser.

### DIFF
--- a/docs/specterd.service
+++ b/docs/specterd.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=SpecterDesktop
+Wants=bitcoind.service
+After=bitcoind.service
+
+[Service]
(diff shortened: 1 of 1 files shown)


k9ert:
Mostly fine but would be nice to have fitting documentation.
Can you review https://github.com/cryptoadvance/specter-desktop/blob/master/docs/daemon.md and adjust it accordingly?

k3tan172:
> Mostly fine but would be nice to have fitting documentation.
> Can you review https://github.com/cryptoadvance/specter-desktop/blob/master/docs/daemon.md and adjust it accordingly?

Disregard PR, I was trying to look for the service file but I didn't look here. This documentation is fine as is!

PR #713, opened 2020-11-28
Included API end points

pxsocs:
Suggestion for API routes on Specter Server. For now, all API functions are "GET" and return json data that can be easily used by other applications.  

### /api/specter ["GET"]
Basic Specter info from app.specter class

### /api/full_txlist/ ["GET"]
Full Transaction list for all wallets

### /api/wallet_info/<wallet_alias>/ ["GET"]
More detailed wallet info and transactions from the app.specter.wallet_manager class

PR #726, opened 2020-12-01
Break the controller into blueprints

ben-kaufman:
This breaks the huge Controller file into multiple enpoints:
```
app.register_blueprint(auth_endpoint, url_prefix="/auth")
app.register_blueprint(devices_endpoint, url_prefix="/devices")
app.register_blueprint(price_endpoint, url_prefix="/price")
app.register_blueprint(settings_endpoint, url_prefix="/settings")
app.register_blueprint(wallets_endpoint, url_prefix="/wallets")
```
Also moved filters to their own file.
One more thing I didn't do yet (because I think it would make sense to do when adding the API with #713) is to move a few global endpoints (generate mnemonic + a few calls for RPC) to a separate API. I added a TODO for that in the code above the 4 endponts I would like to move to an API endpoint.

This also includes a small bugfix for validating address history on Specter DIY via QR code, the index was wrong...


k9ert:
Nice! As soon as the 3 issues are solved, i'll try to run the cypress tests against this!

k9ert:
I ran the cypress tests and also did some manual testing. LGTM, I'll merge today evening if tests are green and no other issues.

Issue #694, opened 2020-11-25
2 of 2 multisig - Multiple addresses fail device verification


hzoman:
Created a 2 of 2 multisig wallet with 2 KeepKey devices.

Using ‘show on device’ option with (default) Address #0:
The Keepkey displayed address does not match the Specter wallet address.
Red notification box pops up: “Addresses did not match! Be careful! Please contact support and beware of using the displayed address!”

Address # 1, # 2, etc, confirm properly as expected.

Additionally, a 2 of 3 multisig wallet with the same 2 KeepKeys plus an Electrum wallet: address #0 displays and confirms correctly.

stepansnigirev:
What is your Bitcoin Core version?
Can you check further addresses, ideally up to number 10?

Does 2-of-2 address the same as in Specter if you create the same wallet in Electrum?

hzoman:
Bitcoin Core version v0.20.1

Tested Addr #0 -10:
0 - fail
1 - pass
2 - failed, then passed on retries
3 - pass
4 - fail
5 - fail
6 - pass
7 - fail
8 - fail
9 - fail
10 - fail

All addresses are the same in Specter vs Electrum wallet creation.

starsoccer:
Same issue as OP

Bitcoin Core version v0.20.1

Tested Addr #0 -10:
0 - fail
1 - fail
2 - pass
3 - fail - quite sure the address specter showed here is what the device showed on attempt 0
4 - fail
5 - pass - at this point it refused to generate new addresses for me without restarting the whole app between each address. 
6 - fail
7 - Stopped working at this point and refused to generate new addresses would just sit with the loading top bar. Gave it 15 mins to generate a new address and tried restarting it multiple times

stepansnigirev:
Also keepkey devices?

starsoccer:
No mine was a trezor model one. I can also try with a trezor T and ledger S if that would help you

stepansnigirev:
I was able to find a problem, working on a fix now. I'll test tomorrow when I get to my hardware wallets and make a pull request.
Probably we will do a v0.10.2 this week with this fix merged.

starsoccer:
Cool thanks, if I can test/assist with anything let me know

starsoccer:
new release has fixed this for me, but going to make a new issue for my ledger nano S not working

PR #733, opened 2020-12-04
fix multisig verification

stepansnigirev:
Close https://github.com/cryptoadvance/specter-desktop/issues/694
As soon as the issue is fixed in HWI we can revert it, a workaround for now.


ben-kaufman (approved):
Tested, works :)

Issue #707, opened 2020-11-27
Use descriptor wallets for Bitcoin Core v0.21.0


Sjors:
Coming soon(tm). I have some ugly WIP code to support it, but anyone should feel free to make a PR earlier.

I suggest checking the Bitcoin Core version and if it's 0.21 or later, leverage the following:

* `createwallet` takes a new `descriptors` argument, set it to `true`
* use `importdescriptors` instead of `importmulti`. It's the same syntax except:
  * you can drop `keypool`
   * you have to set `"active": True` for both receive and change
   * you have to set `"internal": True` for change
   * Nice benefit: the New Address button in Bitcoin Core GUI will work
* you can probably ditch all the keypool refill logic, as Bitcoin Core will just expand the descriptors as needed
* use `balance = self.rpc.getbalances()["mine"]` (instead of "watch-only")

stepansnigirev:
What's the default address gap in core?

Sjors:
The default gap is 1000, and can be configured with `-keypool` or (1 off) by setting a higher range during the descriptor import. If you set it to 1 million things get unacceptably slow, at least the last time I tried.

ben-kaufman:
Closed with #737

PR #737, opened 2020-12-04
Use descriptor wallet for Bitcoin Core >= v0.21.0

Sjors:
First part of #707.

This PR checks if Bitcoin Core is v0.21.0 or above. If so it creates a descriptor wallet.

We then check for each wallet if it's a descriptor wallet (since the node may have been upgraded after a wallet was created), using `getwalletinfo`.

Descriptor wallets use `importdescriptors` instead of `importmulti`. This means the `getnewaddress` RPC works as well as the Receive tab in the Bitcoin Core GUI.

Balance is checked with `mine` rather than `watchonly` for descriptor wallets.

For descriptor wallets we no longer refill the keypool. There's a bunch of other address management code that can probably be skipped, but I don't know how it works. I bypassed the `self._addresses.add` stuff in `keypoolrefill`, which seems to work.

### DIFF
--- a/src/cryptoadvance/specter/wallet.py
+++ b/src/cryptoadvance/specter/wallet.py
@@ -25,7 +25,7 @@ LISTTRANSACTIONS_BATCH_SIZE = 1000
 class Wallet:
     # if the wallet is old we import 300 addresses
     IMPORT_KEYPOOL = 300
-    # a gap of 20 addresses is what many wallets do
+    # a gap of 20 addresses is what many wallets do (not used with descriptor wallets)
     GAP_LIMIT = 20
     # minimal fee rate is slightly above 1 sat/vbyte
     # to avoid rounding errors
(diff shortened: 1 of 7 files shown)


stepansnigirev:
Thanks a lot for the PR! Looks good.
I would like to test it a bit before merging, but I am really excited about this PR and how much we can simplify with Core 21!

`self._addresses` are currently used to get the derivation path and derive descriptors for receiving addresses without rpc calls to Core, but I think we can slightly refactor that and only store labels there. We can derive descriptors for receiving addresses from the index, without lookups (that makes more sense).

I'll have a look over the weekend what we need to change for that and also test your PR :)

Sjors:
There's also the `deriveaddressess` RPC call. It might be best if you add some commits to this PR to deal with that stuff.

k9ert:
For the reference, i'm adding this link which seems to be the best source in order to understand what this PR is about:
https://github.com/bitcoin/bitcoin/pull/15764

Sjors:
I rebased on master, so there's no need for the merge commit...

ben-kaufman:
Looks great, thanks! The only thing which is an issue currently is that without the keypool management from Specter the addresses caching in CSV stops working, and we currently rely on that for quite a bit. We'll have to figure out first how to keep the addresses caching up to date properly without relying on keypool refill events on Specter.

ben-kaufman:
Hi @Sjors, I've worked on this PR now to fix the address caching issue, and also a coin control issue and support the new `add_inputs` flag in `walletcreatefundedpsbt`.

Would it be fine if I just push directly over here to this PR?

Sjors:
@ben-kaufman that's fine by me

ben-kaufman:
Great, just pushed it, tested with Core 0.21 and 0.20, and everything looks good to me now

Sjors:
tACK c8dd881

Tested on Signet with Bitcoin Core v0.21 by creating a new 2-of-2 wallet, receiving coins on it and then sending them.

stepansnigirev (approved):

src/cryptoadvance/specter/wallet.py
```diff
@@ -189,6 +189,45 @@ def fetch_transactions(self):
...
+            while (
+                len(
+                    [
+                        tx
+                        for tx in self._transactions
+                        if self._transactions[tx]["category"] != "send"
+                        and not self._transactions[tx]["address"]
+                    ]
+                )
+                != 0
```
k9ert, line 202: In #1316 i'm currently migrating to v0.21.1 and the tests are hanging in this endless loop.
How does this loop is suppose to end, @Sjors ?

Sjors: @stepansnigirev a Python strangle?

PR #752, opened 2020-12-06
Fix issue with failing to load a wallet

ben-kaufman:


src/cryptoadvance/specter/wallet_manager.py
```diff
@@ -219,8 +226,11 @@ def create_wallet(self, name, sigs_required, key_type, keys, devices):
         if self.working_folder is not None:
```
stepansnigirev, line 226: As you are checking `w` below makes sense to check here as well...

```suggestion
        if w and self.working_folder is not None:
```

PR #758, opened 2020-12-07
Export CSV data

ben-kaufman:
This PR adds the option to export as CSV:
- Wallet transaction history
- Wallet current UTXO
- Combined transaction history of all wallets
-  Combined UTXO list of all wallets
- Addresses history

The export file will also include price at time of the transaction if Bitstamp (or some of the Spotbit providers - some seem to allow that, others don't) is set as the price provider.

Supports both BTC and sats as a unit for the report - based on the unit Specter is configured with in the settings.


k9ert:
I was not able to have a filled csv-file. I only get a file with a header but no data. In this case i only have a coinbase-x and a tx to the same wallet but that shouldn't matter, right?

ben-kaufman:
I think I'll rework this one after #760 is merged since there are probably a ton of merge conflicts...

k9ert:
Awesome PR. I'm not sure whether there are any tables left which are not exportable. Might be tricky to find any left!
LGTM, i'll merge when travis is green!

PR #760, opened 2020-12-08
Refactor History Tab

### LINKED ISSUE #656: Transaction history shows all the UTXOs not only the ones owned by the wallet
I created a new wallet pointing to the Wasabi wallet I use.
After a CoinJoin, the transactions history in Specter shows ALL the outputs of that transaction instead of just the ones actually owned by the wallet.
So with an anonymity set of 95 you see 95 rows + N other rows from the same transaction.

ben-kaufman:
- [x]  Label addresses from everywhere, change label without refresh (see `address-label` below)
- [x] Refactor the UTXO tab to look like the tx history tab.
- [x] Fix transaction caching to track all transactions
- [x] Use the cache instead of `listtransactions` to get the history tab data
- [x] Show batch transactions as a single row (should fix #656)
- [x] Refactor transaction list to use web components 
- [x] Add search and sort transactions

This also adds a new `address-label` web-component, which helps to improve the functionality, reusability, and performance of labeling. Address label component allows to:
- Copy address/ open link to block explorer (if one is configured).
- Specify the current address label or the component will automatically fetch in the background.
- Edit label (with background call instead of full refresh)
- Updating a label will update all `address-label`s on the screen with that same address to the new label


src/cryptoadvance/specter/server_endpoints/wallets.py
```diff
@@ -1000,3 +991,33 @@ def rescan_progress(wallet_alias):
...
+    try:
+        wallet = app.specter.wallet_manager.get_by_alias(wallet_alias)
+        address = request.form.get("address", "")
+        label = wallet.getlabel(address)
+        return {
+            "address": address,
+            "label": label,
+        }
+    except SpecterError as se:
+        app.logger.error("SpecterError while get label: %s" % se)
```
k9ert, line 1008: What about flashing the error to the user? I thought about the SpecterError  as something which is meaningful to the User. We can and imho should raise these kind of errors with that in mind?
Unfortunately I just checked and can't find a single place in the code where this is the case.

src/cryptoadvance/specter/server_endpoints/wallets.py
```diff
@@ -1000,3 +961,178 @@ def rescan_progress(wallet_alias):
...
+    try:
+        wallet = app.specter.wallet_manager.get_by_alias(wallet_alias)
+        address = request.form.get("address", "")
+        label = wallet.getlabel(address)
+        return {
+            "address": address,
+            "label": label,
+        }
+    except Exception as e:
+        flash("Error while get label: %s" % e, "error")
```
k9ert, line 978: Flashing to the user doesn't make sense in API-calls. What about:
```suggestion
        logging.exception(e)
        return "Error while get_label: %s" % e, 500
```
Imho that's valid in other cases as well. Will reference it as "see above".

src/cryptoadvance/specter/server_endpoints/wallets.py
```diff
@@ -1000,3 +961,178 @@ def rescan_progress(wallet_alias):
...
+@login_required
+def set_label(wallet_alias):
+    try:
+        wallet = app.specter.wallet_manager.get_by_alias(wallet_alias)
+        address = request.form["address"]
+        label = request.form["label"]
+        wallet.setlabel(address, label)
+        return {"success": True}
+    except Exception as e:
+        flash("Error while get label: %s" % e, "error")
```
k9ert, line 993: see above

src/cryptoadvance/specter/server_endpoints/wallets.py
```diff
@@ -1000,3 +961,178 @@ def rescan_progress(wallet_alias):
...
+            validate_merkle_proofs=app.specter.config.get(
+                "validate_merkle_proofs", False
+            ),
+            current_blockheight=app.specter.info["blocks"],
+        )
+        return process_txlist(
+            txlist, idx=idx, limit=limit, search=search, sortby=sortby, sortdir=sortdir
+        )
+    except Exception as e:
+        flash("Error while get txlist: %s" % e, "error")
```
k9ert, line 1019: see above

src/cryptoadvance/specter/server_endpoints/wallets.py
```diff
@@ -1000,3 +961,178 @@ def rescan_progress(wallet_alias):
...
+        sortdir = request.form.get("sortdir", "asc")
+        txlist = wallet.utxo
+        for tx in txlist:
+            if not tx.get("label", None):
+                tx["label"] = wallet.getlabel(tx["address"])
+        return process_txlist(
+            txlist, idx=idx, limit=limit, search=search, sortby=sortby, sortdir=sortdir
+        )
+    except Exception as e:
+        flash("Error while get txlist: %s" % e, "error")
```
k9ert, line 1041: see above ... you get the idea, won't mark it anymore.

src/cryptoadvance/specter/templates/includes/tx-table.html
```diff
@@ -0,0 +1,445 @@
...
+        <td></td>
+        <td class="optional hidden blockhash-empty"></td>
+    </tr>
+	<tbody class="tx-tbody">
+	</tbody>
+    </table>
+</template>
+
+<script type="text/javascript">
+    class TxTableElement extends HTMLElement {
```
k9ert, line 144: Would be good to have some comments on the code. Class-documentation and maybe each function or so?

src/cryptoadvance/specter/wallet_manager.py
```diff
@@ -285,6 +290,24 @@ def full_txlist(self, page, validate_merkle_proofs=False):
                 result.append(tx)
         return list(reversed(sorted(result, key=lambda tx: tx["time"])))
 
+    def full_utxo(self):
```
k9ert, line 293: Function-docstrings, please!

src/cryptoadvance/specter/wallet.py
```diff
@@ -465,107 +470,192 @@ def save_pending_psbt(self, psbt):
             )
         self.save_to_file()
 
-    def txlist(self, page, limit=100, validate_merkle_proofs=False):
+    def txlist(
```
k9ert, line 473: docstring, please!

k9ert:
Reviewed some code and tested. Would appreciate more documentation. Made a tiny fix. LGTM.

src/cryptoadvance/specter/templates/includes/tx-table.html
k9ert: Awesome! So much better, really!

src/cryptoadvance/specter/server_endpoints/wallets.py
k9ert: For the reference: Flashing doesn't make sense for api-calls returning json. So instead of returning empty {}, the approach should be to return a reasonable error,essage and 500 code (or better suited code depending on what you know about the rerror).
Then the client is supposed to deal as best as he can with the situation.

Issue #720, opened 2020-11-29
UI: Transactions>History>Address


moritzwietersheim:
"Label" instead of "Address" would be probably better in the overview:

![image](https://user-images.githubusercontent.com/47259243/100552687-f3abdd80-3288-11eb-8752-b7c5a48b4e12.png)

ben-kaufman:
Closed by #764

PR #764, opened 2020-12-09
Switched Address to Label

figgyfigs:
"Label" does make more sense. Quick UI change. Should close issue #720.


stepansnigirev:
Makes sense, thank you!

PR #765, opened 2020-12-09
Tor settings

### LINKED ISSUE #786: hwi bridge settings should be available for the user
currently non-admin users can access it only by editing the url to /settings/hwi , menu button is not there

ben-kaufman:
Make Tor proxy URL and Control port configurable in the settings.
Also, add a Tor only mode to use only Tor if/ when making external calls.
Also added test connection button to the tor panel at the top right corner.
We should probably re-think how to organize that though, I think the difference between simply using tor, and the Tor hidden service, is confusing, and the top right panel looks a bit messy, but I'm not too great at planning the UX...

Also added here option to specify block explorer for UTXO rescan on pruned node. It'll have the same explorer as previously as default, but the user will be able to edit that, so that the Block explorer setting from the general settings page will again be considered as only for links, not external calls.

Also fixes #786


src/cryptoadvance/specter/rpc.py
```diff
@@ -240,6 +253,8 @@ def wallet(self, name=""):
             path="{}/wallet/{}".format(self.path, name),
             timeout=self.timeout,
             session=self.session,
+            proxy_url=self.proxy_url,
```
k9ert, line 256: inject the procy_url into the wallet? Can't we avoid that? The wallet-implementation should imho be a bit more independent of the underlying network-technology.

src/cryptoadvance/specter/hwi_server.py
```diff
@@ -93,12 +93,10 @@ def api():
...
             )
         data["forwarded_request"] = True
-        requests_session = requests.Session()
+        requests_session = app.specter.requests_session(
+            force_tor=".onion/" in app.specter.hwi_bridge_url
+        )
         requests_session.headers.update({"origin": request.environ["HTTP_ORIGIN"]})
-        if ".onion/" in app.specter.hwi_bridge_url:
-            requests_session.proxies = {}
-            requests_session.proxies["http"] = "socks5h://localhost:9050"
```
k9ert, line 100: Looks like default-configuration. Can we put that in config.py ?

src/cryptoadvance/specter/rpc.py
```diff
@@ -152,7 +158,9 @@ def autodetect_rpc_confs(datadir=get_default_datadir(), port=None):
     available_conf_arr = []
     if len(conf_arr) > 0:
         for conf in conf_arr:
-            rpc = BitcoinRPC(**conf)
+            rpc = BitcoinRPC(
+                **conf, proxy_url="socks5h://localhost:9050", only_tor=False
```
k9ert, line 162: config.py ?

src/cryptoadvance/specter/rpc.py
```diff
@@ -209,6 +217,8 @@ def __init__(
         path="",
         timeout=None,
         session=None,
+        proxy_url="socks5h://localhost:9050",
```
k9ert, line 220: config.py ?

src/cryptoadvance/specter/server_endpoints/settings.py
```diff
@@ -233,6 +234,87 @@ def general():
     )
 
 
+@settings_endpoint.route("/tor", methods=["GET", "POST"])
+@login_required
+def tor():
```
k9ert, line 239: Some documentation maybe something like:
'''
controls the tor-behaviour
POST only
param action might be "save", "test_tor" or "toggle_hidden_service"
param proxy_url usually something like socks5h://localhost:9050
param only_tor "on" or something else (off)
...
'''

src/cryptoadvance/specter/templates/settings/tor_settings.jinja
```diff
@@ -0,0 +1,76 @@
...
+				<button type="submit" class="btn" name="action" value="save">Save</button>
+			</div><br><br>
+            {% if not debug %}
+                <h2>Tor Hidden Service</h2>
+                <p>Running Specter behind a Tor hidden service allows you to acccess Specter over Tor from anywhere.<br>
+                </p>
+                {% if tor_service_id %}
+                    {% if specter.config.auth == "none" %}
+                        <p class="warning" style="max-width:700px;">
+                        ⚠️ Warning!<br>
```
k9ert, line 51: Warning is imho too weak. Can we prevent that entirely?

src/cryptoadvance/specter/templates/settings/tor_settings.jinja
```diff
@@ -0,0 +1,76 @@
...
+			<br>
+            <p class="note">
+			Default for Tor daemon: socks5://127.0.0.1:9050<br>
+			Default for Tor Browser: socks5://127.0.0.1:9150
+			</p>
+			Tor control port:
+			<p class="note">
+			Restart Specter for change of the control port to take effect.
+			</p>
+			<input name="tor_control_port" value="{{ tor_control_port }}" type="text" placeholder='Set "default" for default or specify port to use.' />
```
k9ert, line 23: The placeholder text is not visible if the field is pre-populated with "default". And default is also a very strange value. 
Maybe we can we keep it empty and do the placeholder like:
"Set to custom port here, if kept empty the default (9051) is used"

src/cryptoadvance/specter/user.py
```diff
@@ -100,6 +100,18 @@ def set_explorer(self, specter, explorer):
         self.config["explorers"][specter.chain] = explorer
         self.save_info(specter)
 
+    def set_proxy_url(self, specter, proxy_url):
```
k9ert, line 103: Every user can have his own tor-configuration? Does that really make sense? Isn't that exactly what the admin should be responsible for?

src/cryptoadvance/specter/util/tor.py
```diff
@@ -40,6 +40,24 @@ def start_hidden_service(app):
             f.write("%s.onion" % app.tor_service_id)
     app.tor_service_id = app.tor_service_id
     app.tor_enabled = True
+    if app.specter.config.get("auth", "none") == "none":
```
k9ert, line 43: Throw a SpecterError("Sorry, Tor is not supported with no Authentication-method! Please activate one before activating Tor!")
Ideally there is somewhere up the call-stack the endpoint which is flashing it to the user (or an API-call which returns a 500 with that message).

src/cryptoadvance/specter/util/version.py
```diff
@@ -12,12 +12,13 @@
 
 
 class VersionChecker:
-    def __init__(self, name="cryptoadvance.specter"):
+    def __init__(self, name="cryptoadvance.specter", specter=None):
```
k9ert, line 15: This is a good example: We're imho complicating here the specter-object even though this does not need to be necessarily tied to the specter-object. Maybe it make sense to have a "connection-object" which handles all sorts of connections. It can be passed to e.g. this VersionChecker and it would also potentially prevent cyclic dependencies which we already imho have in this case.

k9ert:
There are definitely some architectural question-marks with this PR. I have the impression that the tor-implementation-detail is leaking too much in effectively all aspects of specter. Ideally i think the "network-layer" should be separated a lot more from business-code.
What do you think?

src/cryptoadvance/specter/templates/settings/tor_settings.jinja
ben-kaufman: We prevent it from the UI, the warning is only for those who do it with the CLI, so it's not entirely impossible, but not something the average user can do "accidentally"

src/cryptoadvance/specter/util/tor.py
ben-kaufman: But I think from the command line we can allow this, and just block it from the UI, so that advanced users can decide for themselves.

src/cryptoadvance/specter/rpc.py
k9ert: Bullshit. That's not the wallet-impl.

k9ert:
After removing the tor-options for users, this no longer looks that scary. As this is suppose to reduce support effort, let's merge and ship quick.

PR #781, opened 2020-12-15
Add addresses tab allowing users to navigate through derived addresses

jleo84:
This PR adds a new tab called "addresses" that allows users to see a list of the derived addresses (from receive descriptor or from change descriptor).

This is a common feature available in some good wallets (like Electrum), but it's missing within Specter experience.
It also provides an overall view of the wallet.

The screen show not just the addresses but also:
. If the address is already used or not
. The line with used addresses is visually different from unused ones
. The HD Key Path
. How many UTXO are associate with each address
. The amount of each address

Although I have tested during the development, more tests are recommended.
image


ben-kaufman:
Hi, first, thanks a lot for the PR!

The issue I have is that it is very intensive with calls to Bitcoin Core. We are now trying to work on minimizing calls as far as possible, since many users are running their Bitcoin Core remotely, and often over Tor, which makes calls to Core very slow in these cases.

We already have a table of addresses in the Receive tab, which takes information from caching, so if you'd like you could use this code to see how to use the caching for such address table.
If you'd like to do that and have questions or need some help let me know and I'll be happy to assist.

jleo84:
Thank you for the review. I truly appreciate it.

I agree that it is very intensive with calls to Bitcoin Core.
The first version of this code just called one function: self.rpc.listtransactions() (the same used in the method fetch_transactions() of wallet.py) and so, it could identify the used addresses.

But the result of this method does not show the change addresses that were used. The “sent” operations just show the recipient address. So this approach worked for receive addresses but not for the change addresses.

The wallets that have the feature proposed in this PR are Electrum Server based, that has transactions and the addresses already  indexed.

If there’s a way to retrieve all the transactions - including the change address associated - in a single call, I can update this PR with a more efficient code.

ben-kaufman:
Hi, you can see here how we're getting addresses out of the caching and displaying them in a table with info:
https://github.com/cryptoadvance/specter-desktop/blob/119a70e1e61d7dc0d6bced592f898722c5d268a7/src/cryptoadvance/specter/server_endpoints/wallets.py#L481

https://github.com/cryptoadvance/specter-desktop/blob/119a70e1e61d7dc0d6bced592f898722c5d268a7/src/cryptoadvance/specter/templates/wallet/receive/wallet_receive.jinja#L103

jleo84:
I've updated the code to use the cache mechanism.
I've done some tests and it apparently worked well.

However two RPC calls still persists:

1. self.rpc.listunspent(0) - to retrieve information about UTXO and amount. It is just called once.

2. self.rpc.getaddressinfo(addr.address)  - to retrieve HD Key Path. It is called at each address. One possible solution to minimize this call is to store this information in the cache or simply remove this information from addresses list. But some users can consider it important.

jleo84:
New commit. I've made some more improvements.
Now the code use RPC Multi for get HD Key Path.
```python
rpc_multi_getaddressinfo = self.rpc.multi([("getaddressinfo", addr.address) for addr in addresses_cache])
```
Thus the call `getaddressinfo` is no longer in each iteration of the loop. It's just called once through multi RPC call.

There's no others RPC calls inside the address list code.  It uses the same code used in transactions screen:
```python
# update balances in the wallet
app.specter.check_blockheight()
wallet.get_balance()
wallet.check_utxo()
```

jleo84:
There's an other issue, but apparently it's not related to this functionality.
The change addresses haven't been used sequentially. This information come from `wallet._addresses`

If I run the code below, I receive the result in the image below from command line.

```python
addresses_cache = [v for _, v in self._addresses.items() if v.change == is_change]

for a in addresses_cache:
    print('-------')
    print(a.index)
    print(a.address)
    print(a.used)
```

image

The change addresses screen also shows this issue:

image

ben-kaufman:
Thanks! Looks much better now :)

The biggest problem I see now is that this breaks on multisig wallets, since they don't return hdkeypath (since there might be multiple...). I'd suggest just not have it in the table, or at least not for multisig wallets. But I think the index is enough to be honest, the derivation is pretty much a constant with the index at its end...

It'll also be nice to have a "label" column there like we have in the master branch on the transactions screen. (Using the `<address-label>` element. Having a verify on device option would be great too but might be harder to add. We have it on the address history tab already so you could copy from there, but the issue would be with the time it takes to generate descriptors for each address. The best solution would probably be to make a call from the JS (using `fetch`) to an endpoint on the server (like we do for `get_label` and `set_label` for example). This might be a bit complicated, so not necessary for this PR.

ben-kaufman:
Regarding the change used non sequentially, was the wallet created and used solely from Specter? Or did you just import an existing one into Specter or maybe used it from another software as well previously?

jleo84:
Ok. I've removed the HD Key Path (it is really null in multisign) and added the label column.

I also removed the multi RPC call to `getaddressinfo`, but it returns the descriptor of each address. For example:
```
'desc': "wsh(multi(2,[42a7aba0/48'/1'/0'/2'/0/0]02d36b4da76c2627388b496cd823b1789a022f5c35892bcfa198586b79b0705fcd,[9b5a6abe/48'/1'/0'/2'/0/0]033fb2d9a2aeb713c80aef7a1597fcd942722dc8c500c02bc36e83fbf00c2042ae))#q7jrxc2p"`
```

It works for both single and multisig.
Since the descriptor can't be changed, maybe there's not need for asynchronous call.

I'll check how it is done on the address history in master branch.

Maybe it can be used a modal screen called "Address Info" when user clicks on address, similarly it's done for "Transaction Details" when user clicks on transaction ID.

image

jleo84:
_Regarding the change used non sequentially, was the wallet created and used solely from Specter?_
Yes. It was a single key  (hot)  wallet create from Specter with Bitcoin Core Testnet v 0.20.0 .

I also created a new 2-of-2 multisign wallet (using two devices - both hot wallet) and the result was the same (image below).


image

ben-kaufman:
Hi @jleo84, really sorry for the late response!

Looks really good now! Can you please just rebase/ merge master? There's a conflict marked but it should be fine to just leave both changes as they seem to not really conflict.

stepansnigirev:
I like it, looks really good!
Would it make sense to reverse the order of the addresses so the new ones come first?

There are a few things we could improve later (probably in the next PRs):
- add pagination, too much to scroll if you have more than 100 addresses
- allow export as csv from this screen as well, and export only recv or change addresses depending on the currently active tab

jleo84:
Hi @ben-kaufman @stepansnigirev thank you for the reviews.

No problem, I reversed the order of the addresses, so it can be evaluate if this is the better way to show them.
I had implemented from used addresses to the unused ones just because it is the Electrum approach.

When everything is OK, I can create the others PRs for the pagination and CSV.

ben-kaufman (approved):
Looks good, thanks!

I'm not sure whatever it's best to have it ordered like that or first to last, but I guess we could add sorting like in the history tab later on.

In another PR, we could probably also add here a verify on device option, but since you need the descriptor for that and the calls take some time it might be better to fetch it in the background individually on a button click. Let me know if that's something you'd like to do as well or if I should do that.

jleo84:
Yes, I would like to do that.

I am also working in a new version of address tab, that is asynchronous, similar to history/transactions.
This will enable improvements such as sorting and pagination.

I can fork this branch or from master after this one is merged.

ben-kaufman:
Sounds good, thanks! Merging this now so you can pull and work from master.

PR #790, opened 2020-12-18
Fixed wallet name bug in json backup download

### LINKED ISSUE #684: Backup files naming: use wallet current name
I think it would be better when you export a backup file to name it with the current name instead of the directory's original name given (explicitly or not) at wallet creation time.

For example in my case the following are the tests wallet I am using. I didn't change the default name of some of them at creation time. But I did it later on from the wallet settings. 
So now the directories don't match the Wallet names shown in Specter and the backup files are named after these directories. It is confusing and you have to manually change them at backup time.

image

rajarshimaitra:
Found the bug. 
Fixes #684


rajarshimaitra:
Pushed a minor Doc nit for `launch.json` for python debugging.

ben-kaufman (approved):
Looks good, thanks!

PR #802, opened 2020-12-24
Fixes fiat symbol

figgyfigs:
USD and GBP precedes amount. Euro symbol will come after the amount. 

Black Python Linter should pass now.

### DIFF
--- a/src/cryptoadvance/specter/server_endpoints/filters.py
+++ b/src/cryptoadvance/specter/server_endpoints/filters.py
@@ -59,7 +59,7 @@ def btcunitamount(context, value):
 @filters_bp.app_template_filter("altunit")
 def altunit(context, value):
     if app.specter.price_check and (app.specter.alt_rate and app.specter.alt_symbol):
-       rate = (
+        rate = (
             "{:,.2f}".format(float(value) * float(app.specter.alt_rate))
             .rstrip("0")
             .rstrip(".")
--- a/src/cryptoadvance/specter/server_endpoints/filters.py
+++ b/src/cryptoadvance/specter/server_endpoints/filters.py
@@ -59,17 +59,15 @@ def btcunitamount(context, value):
 @filters_bp.app_template_filter("altunit")
 def altunit(context, value):
     if app.specter.price_check and (app.specter.alt_rate and app.specter.alt_symbol):
-        if app.specter.alt_symbol == "€":
-            return (
-                "{:,.2f}".format(float(value) * float(app.specter.alt_rate))
-                .rstrip("0")
-                .rstrip(".")
-                + app.specter.alt_symbol
-            )
+       rate = (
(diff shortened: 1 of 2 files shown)


src/cryptoadvance/specter/templates/components/price_bar.jinja
```diff
@@ -27,7 +27,11 @@
 </style>
 <a id="price-bar-btn" type="button" style="cursor: pointer;" class="settings-bar-btn left-toolbar" onclick="showPageOverlay('price-popup')">
     {% if specter.price_check %}
-        <span>{{ ("%.2f"|format(specter.alt_rate|float)).rstrip('0').rstrip('.') }}{{ specter.alt_symbol }}</span>
+        {% if specter.alt_symbol == "€" %}
+            <span>{{ ("%.2f"|format(specter.alt_rate|float)).rstrip('0').rstrip('.') }}{{ specter.alt_symbol }}</span>
+        {% else %}
+            <span>{{ specter.alt_symbol }}{{ ("%.2f"|format(specter.alt_rate|float)).rstrip('0').rstrip('.') }}</span>
+        {% endif %}
```
ben-kaufman, line 34:
```suggestion
        <span>{{ 1 | altunit}}</span>
```
Would actually be cleaner like this.

src/cryptoadvance/specter/server_endpoints/filters.py
```diff
@@ -59,12 +59,17 @@ def btcunitamount(context, value):
...
+            return (
+                "{:,.2f}".format(float(value) * float(app.specter.alt_rate))
+                .rstrip("0")
+                .rstrip(".")
+                + app.specter.alt_symbol
+            )
+        else:
+            return app.specter.alt_symbol + "{:,.2f}".format(
+                float(value) * float(app.specter.alt_rate)
+            ).rstrip("0").rstrip(".")
```
ben-kaufman, line 72:
```suggestion
       rate = (
            "{:,.2f}".format(float(value) * float(app.specter.alt_rate))
            .rstrip("0")
            .rstrip(".")
        )
        if app.specter.alt_symbol in ["$", "£"]:
            return app.specter.alt_symbol + rate
        else:
            return rate + app.specter.alt_symbol
```
I'd suggest doing it like this. Both cleaner and because I think the default should be to the right, since if you use the manual price settings for stuff which don't have symbol (So you put usually something like ticker symbol, I think they should be on the right. (I.e. 1 USD, 10 AUX etc.). Does that make sense?

figgyfigs: Agreed, I did not think about that! Makes sense.

ben-kaufman (approved):

ben-kaufman:
Looks good, thanks! Merging.

PR #837, opened 2021-01-03
Allow uppercase bech32 addresses

ben-kaufman:
Allow input uppercase bech32 addresses on send.

### DIFF
--- a/src/cryptoadvance/specter/server_endpoints/wallets.py
+++ b/src/cryptoadvance/specter/server_endpoints/wallets.py
@@ -559,6 +559,12 @@ def send_new(wallet_alias):
                         amounts.append(float(output.split(",")[1].strip()) / 1e8)
                     else:
                         amounts.append(float(output.split(",")[1].strip()))
+            addresses = [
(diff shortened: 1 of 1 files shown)


src/cryptoadvance/specter/server_endpoints/wallets.py
```diff
@@ -559,6 +559,10 @@ def send_new(wallet_alias):
                         amounts.append(float(output.split(",")[1].strip()) / 1e8)
                     else:
                         amounts.append(float(output.split(",")[1].strip()))
+            addresses = [
+                address.lower() if address.startswith("bc1") else address
```
stepansnigirev, line 563: This code will never execute. If it's upper case it never start with lower bc1.
We need to check that it starts with one of these:
BC1, TB1, BCRT1
Or better check if all chars are in capitalized bech32 alphabet.

ben-kaufman: I just check the prefixes, seems shorter and I don't see a case where that won't work right?

PR #843, opened 2021-01-05
Consistent rescan labels

justinmoon:
There are two rescan buttons in settings. One says "Abort" and the other says "Abort rescan". This PR makes them both say "Abort rescan".


ben-kaufman:
Makes sense. Thanks

PR #844, opened 2021-01-05
Improve messaging with concurrent utxo & blockchain rescans

justinmoon:
Since blockchain and utxo rescans can happen concurrently, it's probably better to allow transactions page to show messages for them independently.

### DIFF
--- a/src/cryptoadvance/specter/templates/wallet/history/components/total_wallet_balances.jinja
+++ b/src/cryptoadvance/specter/templates/wallet/history/components/total_wallet_balances.jinja
@@ -23,7 +23,12 @@
     </h1>
     {% if wallet.rescan_progress or specter.utxorescanwallet == wallet.alias %}
         <span id="wallet_rescan_data" style="display: contents;">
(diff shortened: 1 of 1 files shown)


stepansnigirev:
This change makes two elements with the same id, that will break javascript updating these values.

Here is the code that uses these elements:

https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/templates/includes/sidebar/sidebar.jinja#L74

https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/templates/includes/sidebar/components/sidebar_wallet_list_item.jinja#L44

@ben-kaufman I think it's confusing that elements with ids are separated from the javascript code managing them.
We could use signals, and these elements can listen to the signals and update their values accordingly (similar to our messageboxes). Or at least use classes instead of ids.

k9ert:
@justinmoon do you still consider to work on this? I'm closing that for now but feel free to simply reopen it.

PR #848, opened 2021-01-07
Added persistent tor setting.

### LINKED ISSUE #710: Save tor-enabled setting in config
Currently if you enable Tor from the UI and restart Specter it will not start with tor enabled automatically.
I think it makes sense to store "tor_enabled" flag in config so Specter knows that on launch it should enable tor.

rajarshimaitra:
Fixes #710 

Tried to do it minimally. 

- adds a `tor_status` flag (default `false`) to `app.specter.config`. 
- `Start` button in tor settings page toggles this flag.
- app reads the flag at launch and starts tor if the flag is set `true` in the last session.

Problem:

- Doesn't work on the first init run (no clue why, need help)

debugged so far:

- In the first run when the `Start` tor button is clicked in tor seeting, `toggle_tor_status` is hit.
- Status is read as `false 
- `config.json` file updates. 
- But the status flag doesn't change state. It remains `false`... weird. 

Then It works as expected from the subsequent runs without any issue. So the user will see expected behaviour when they run the app second time and starts tor again. Then the flag changes and tor will run automatically for next runs until stopped.

### DIFF
--- a/src/cryptoadvance/specter/cli/cli_server.py
+++ b/src/cryptoadvance/specter/cli/cli_server.py
@@ -222,7 +222,7 @@ def server(
                       Starting in production mode instead."
                 )
                 debug = False
-            if tor or os.getenv("CONNECT_TOR") == "True":
+            if tor or os.getenv("CONNECT_TOR") == "True" or app.specter.config["tor_status"] == True:
                 try:
                     app.tor_enabled = True
                     start_hidden_service(app)
--- a/src/cryptoadvance/specter/cli/cli_server.py
+++ b/src/cryptoadvance/specter/cli/cli_server.py
@@ -230,6 +230,8 @@ def server(
                 try:
                     app.tor_enabled = True
                     start_hidden_service(app)
+                    if app.specter.config["tor_status"] == False:
+                        app.specter.toggle_tor_status()
                 except Exception as e:
                     print(f" * Failed to start Tor hidden service: {e}")
                     print(" * Continuing process with Tor disabled")
--- a/src/cryptoadvance/specter/cli/cli_server.py
+++ b/src/cryptoadvance/specter/cli/cli_server.py
@@ -222,7 +222,11 @@ def server(
                       Starting in production mode instead."
                 )
                 debug = False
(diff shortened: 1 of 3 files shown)


rajarshimaitra:
Updated with lint corrections

src/cryptoadvance/specter/specter.py
```diff
@@ -119,6 +119,7 @@ def __init__(self, data_folder="./data", config={}):
             "proxy_url": "socks5h://localhost:9050",  # Tor proxy URL
             "only_tor": False,
             "tor_control_port": "",
+            "tor_status": "false",
```
ben-kaufman, line 122:
```suggestion
            "tor_status": False,
```
This will fix the first init issue. The problem is that you made it a string on init, which evaluates to true (if not empty) when checked as boolean. So instead of starting as false it started as string evaluating to False when first flipping it in  the `toggle_tor_status `

rajarshimaitra: Yes that was it. silly me. :)

rajarshimaitra:
Updated.

src/cryptoadvance/specter/cli/cli_server.py
```diff
@@ -222,7 +222,11 @@ def run(debug=debug):
                       Starting in production mode instead."
                 )
                 debug = False
-            if tor or os.getenv("CONNECT_TOR") == "True":
+            if (
+                tor
+                or os.getenv("CONNECT_TOR") == "True"
+                or app.specter.config["tor_status"] == True
+            ):
```
ben-kaufman, line 229: If I use the --tor flag there can be a situation where I then turn the tor service off from the app, then the toggle happens incorrectly...
Could add here a check that if the toggle is false then toggle here on start

rajarshimaitra: Hmm makes sense. Will update shortly..

rajarshimaitra:
Updated with tor status toggling in init.

ben-kaufman:
Thanks! Looks good, merging.

PR #849, opened 2021-01-07
[ADD] SLIP-132 switch for PDF backup key format

maxidev:
Added the SLIP-132 switch on the backup PDF view so the document can properly reflect master key format.
image

### DIFF
--- a/src/cryptoadvance/specter/templates/wallet/receive/wallet_receive.jinja
+++ b/src/cryptoadvance/specter/templates/wallet/receive/wallet_receive.jinja
@@ -186,6 +186,14 @@
 			</div>
 			{% include 'wallet/components/wallet_pdf.jinja' %}
 		</div>
+		<div style="margin: 0 auto;" class="center">
+			<label style="font-size: 1em;"">Use SLIP-132:&nbsp;</label>
+			<label class="switch">
+				<input type="checkbox" id="use_slip_132_pdf" style="margin: auto;" onchange="toggleKeyDisplay()" checked>
+				<span class="slider"></span>
+			</label><br>
+		</div>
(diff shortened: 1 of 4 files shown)


ben-kaufman:
Hi, thanks, looks good for existing wallets, but breaks the popup of the new wallet since there's no element with that id there.
https://github.com/cryptoadvance/specter-desktop/blob/cf6a2b9444ff29c45d02b1ad1e3e93a9f6f0a269/src/cryptoadvance/specter/templates/wallet/receive/wallet_receive.jinja#L179

I guess you can either add the same toggle option there or just put it on false by default if that complicates the UI

maxidev:
PR updated to fix the new wallet creation view, also adding the switch.

src/cryptoadvance/specter/templates/wallet/receive/wallet_receive.jinja
```diff
@@ -217,6 +225,11 @@
 				btn.innerText = 'Hide past addresses';
 			}
 		}
+		function toggleKeyDisplay() {
+			let slip132Checkbox = document.getElementById('use_slip_132_pdf');
+			let xpubs = document.getElementsByClassName('slip132_xpub');
+			let zpubs = document.getElementsByClassName('normal_xpub');
+		}
```
ben-kaufman, line 232: What is this for? Looks like it doesn't do anything...

maxidev: Done, commented out the selectors and left a minimal detail on why they are not completely removed.

ben-kaufman (approved):
Looks good, thanks!

PR #852, opened 2021-01-09
add rate limiting and registration link expiry

djpnewton:
-add rate limiting to login and register endpoints
-change otp codes to be harder to guess

### DIFF
--- a/src/cryptoadvance/specter/server_endpoints/settings.py
+++ b/src/cryptoadvance/specter/server_endpoints/settings.py
@@ -422,7 +422,7 @@ def auth():
                     expiry = 0
                     expiry_desc = ""
                 app.specter.add_new_user_otp(
-                    {"otp": new_otp, "created_at": now, "expiry": now}
+                    {"otp": new_otp, "created_at": now, "expiry": expiry}
                 )
                 flash(
                     "New user link generated{}: {}auth/register?otp={}".format(
--- a/src/cryptoadvance/specter/server_endpoints/settings.py
+++ b/src/cryptoadvance/specter/server_endpoints/settings.py
@@ -390,9 +390,7 @@ def auth():
                     current_user.password = hash_password(specter_password)
                 current_user.save_info(app.specter)
             if current_user.is_admin:
-                app.specter.update_auth(
-                    method, rate_limit, registration_link_timeout
-                )
+                app.specter.update_auth(method, rate_limit, registration_link_timeout)
                 if method == "rpcpasswordaspin" or method == "usernamepassword":
                     if method == "usernamepassword":
                         users = [
--- a/src/cryptoadvance/specter/server_endpoints/settings.py
+++ b/src/cryptoadvance/specter/server_endpoints/settings.py
@@ -331,6 +331,7 @@ def auth():
     current_version = notify_upgrade(app, flash)
     auth = app.specter.config["auth"]
     auth_rate_limit = app.specter.config["auth_rate_limit"]
+    registration_link_timeout = app.specter.config["registration_link_timeout"]
     users = None
     if current_user.is_admin and auth == "usernamepassword":
         users = [user for user in app.specter.user_manager.users if not user.is_admin]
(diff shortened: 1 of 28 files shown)


djpnewton:
- do we want to show (and make editable) the rate limiting on the `/settings/auth` page?
- I could add expiration to the registrations links too
  - would we want the expiration time editable?

ben-kaufman:
Thanks! I think it'd be nice to have if you'd like to implement that!

djpnewton:
added:
-  allow user to view/change login and registration rate limiting value
- add registration link expiry
- make it harder for attackers to discover usernames
- disallow empty usernames
- disallow short passwords
- move "generate registration link" button to the bottom to try and make it clearer that it wont update settings

It seems like the config should have an "auth" subsection now like "rpc"

ben-kaufman (approved):

ben-kaufman:
That'd make sense, would you like to move the auth data to its own section in this PR?
If not I think it's good to merge.

djpnewton:
Created "auth" subsection in config to hold new auth settings (rate limiting, registration expiry, password chars) and changed "auth" setting to "method" in the "auth" subsection

migrate old config files

src/cryptoadvance/specter/server_endpoints/settings.py
```diff
@@ -385,13 +408,25 @@ def auth():
...
+                    if timeout > 1:
+                        expiry_desc = " (expires in {} hours)".format(timeout)
+                    else:
+                        expiry_desc = " (expires in 1 hour)"
+                else:
+                    expiry = 0
+                    expiry_desc = ""
                 app.specter.add_new_user_otp(
-                    {"otp": new_otp, "created_at": time.time()}
+                    {"otp": new_otp, "created_at": now, "expiry": now}
```
ben-kaufman, line 425:
```suggestion
                    {"otp": new_otp, "created_at": now, "expiry": expiry}
```

Looks like here it should be expiry instead of now

src/cryptoadvance/specter/specter.py
```diff
@@ -549,13 +560,26 @@ def add_new_user_otp(self, otp_dict):
...
-    def burn_new_user_otp(self, otp):
-        """ validates an OTP for user registration and removes it if valid"""
+    def validate_new_user_otp(self, otp):
+        """ validates an OTP for user registration and removes it if expired"""
+        if "new_user_otps" not in self.config:
+            return False
+        now = time.time()
+        for i, otp_dict in enumerate(self.config["new_user_otps"]):
+            if otp_dict["otp"] == otp:
+                if "expiry" in otp_dict and otp_dict["expiry"] > now:
```
ben-kaufman, line 570:
```suggestion
                if "expiry" in otp_dict and otp_dict["expiry"] < now and otp_dict["expiry"] > 0:
```
The condition here seems to be reversed. Also lacking the case where expiry is 0.

ben-kaufman (commented):
Looks really good, thanks!

Seems like there are 2 small mistakes in the expiration logic, I've suggested fixes there.

src/cryptoadvance/specter/server_endpoints/settings.py
djpnewton: done

src/cryptoadvance/specter/specter.py
djpnewton: done

djpnewton:
thanks for the review

src/cryptoadvance/specter/server_endpoints/settings.py
```diff
@@ -280,7 +280,7 @@ def tor():
                 tor_connectable = False
         elif action == "toggle_hidden_service":
             if not app.config["DEBUG"]:
-                if app.specter.config.get("auth", "none") == "none":
+                if app.specter.config.auth.get("method", "none") == "none":
```
ben-kaufman, line 283:
```suggestion
                if app.specter.config["auth"].get("method", "none") == "none":
```

ben-kaufman:
One more issue I just noticed is that we have a few places still comparing `specter.config.auth` to `none` where it now needs to be `specter.config.auth.method` (in the Jinja files, tor_settings.jinja and base.jinja, you can see by searching them for that) and in the settings.py there's another fix needed I marked as well.

src/cryptoadvance/specter/server_endpoints/settings.py
djpnewton: done

djpnewton:
good catch, updated

ben-kaufman (approved):
Looks good, thanks!

PR #859, opened 2021-01-10
Fix no block height with Bitcoin Core v0.19

### LINKED ISSUE #851: 2-of-3 multisig history all showing "Unconfirmed" after 1.0
History had previously shown correct confirmation counts prior to v1.0. Tried a wallet rescan via the UI but still shows everything as "Unconfirmed". Individual transactions are all listed as expected (aside from being displayed as unconfirmed). If I click into a transaction it says "Mine at block: undefined" but all other info is correct (tx id, size, Block hash, Block time, Inputs/Outputs).

Wallet is in an isolated user login within Specter.

Seeing this same issue in all my 2/3 multisig wallets.

Connection to my node running on the same machine has no complaints. Full node, no pruning, indexing enabled.

ben-kaufman:
Bitcoin Core v0.19 doesn't return the `blockheight` as part of the `gettransaction` command, which currently breaks Specter blockheight caching and tx details display.

Adding here a fix to calculate the blockheight if the property doesn't exist, but this should be removed once we drop support for Bitcoin Core v0.19.

Fix #851

### DIFF
--- a/src/cryptoadvance/specter/server_endpoints/wallets.py
+++ b/src/cryptoadvance/specter/server_endpoints/wallets.py
@@ -949,6 +949,14 @@ def decoderawtx(wallet_alias):
         txid = request.form.get("txid", "")
         if txid:
             tx = wallet.rpc.gettransaction(txid)
+            # This is a fix for Bitcoin Core versions < v0.20
+            # These do not return the blockheight as part of the `gettransaction` command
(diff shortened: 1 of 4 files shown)


stepansnigirev (approved):

stepansnigirev:
Tested, works. I think we can merge it.

PR #860, opened 2021-01-10
Auth improvements

ben-kaufman:
- Add password protection auth option
- Disable RPC as password if autodetect is on so we still can log in even if the node is offline


src/cryptoadvance/specter/server_endpoints/settings.py
```diff
@@ -370,7 +370,16 @@ def auth():
                 current_user.save_info(app.specter)
             if current_user.is_admin:
                 app.specter.update_auth(auth)
-                if auth == "rpcpasswordaspin" or auth == "usernamepassword":
+                if (
+                    auth == "rpcpasswordaspin"
+                    or auth == "passwordonly"
+                    or auth == "usernamepassword"
+                ):
```
stepansnigirev, line 377:
```suggestion
                if auth in ["rpcpasswordaspin", "passwordonly", "usernamepassword"]:
```

k9ert:
tested, LGTM, ready to merge!

PR #861, opened 2021-01-10
Display slashes and parentheses correctly in tx labels

caldon:
Replace slashes and parentheses in tx data by corresponding HTML5 entities (and not all by `&apos;`).

Without this PR, Specter (e.g. v1.0) displays all slashes and parentheses in tx labels as apostrophes (in both History and UTXO views).


ben-kaufman (approved):
Thanks! Good catch

PR #865, opened 2021-01-12
Fix tx info showing wrong input address

### LINKED ISSUE #864: Wrong input address in PSBT preview
I noticed just now that an unsigned transaction for a multi-sig wallet in Specter showed the wrong input address.

The transaction I created consisted of 1 input and 1 output.
The input correctly showed that output #0 will be used for the transaction, but the shown address was actually from output #1.
The created PSBT was correctly shown in the hardware devices with the address of output #0.

So this seems to just be something visual in Specter.

ben-kaufman:
Fix #864 

Also removes unnecessary Label field since we now have the <address-label> object there.

### DIFF
--- a/src/cryptoadvance/specter/templates/wallet/send/sign/wallet_send_sign_psbt.jinja
+++ b/src/cryptoadvance/specter/templates/wallet/send/sign/wallet_send_sign_psbt.jinja
@@ -124,12 +124,11 @@
 						<b>Transaction id:</b><br>
 						{{ explorer_link('tx', input['txid'], input['txid'], specter.explorer) }}
 						(Output #{{ input['vout'] }})<br>
-						{% set tx = wallet.gettransaction(input['txid']) %}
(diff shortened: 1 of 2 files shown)


stepansnigirev:
With existing PSBT got an error:
`jinja2.exceptions.UndefinedError: No first item, sequence was empty.`

Transaction had 2 recipients.

Came from here (`wallet_send_sign_psbt.jinja2` -> `tx_data|first`):
```py
{% if tx_data %}
	<b>Address:</b>
	<address-label data-address="{{ (tx_data|first).address }}" data-wallet="{{ wallet_alias }}"></address-label>
	<br>
{% endif %}
```

ben-kaufman:
Strange because 1. It should never be empty. 2. I do check for emptyness before taking the first element `{% if tx_data %}`
I even checked when intentionally setting it to be empty it still doesn't error because the check for emptiness catches it...

stepansnigirev (approved):

PR #866, opened 2021-01-12
[FIX] #784 URL encode to prevent breaking characters on RPC connection

### LINKED ISSUE #784: Can't connect to the node with certain symbols in the password
For example if % is in the password

maxidev:
This fixes #784 by doing url encode on both user and password and thus, parsing characters that might break the URL format.

### DIFF
--- a/src/cryptoadvance/specter/rpc.py
+++ b/src/cryptoadvance/specter/rpc.py
@@ -2,6 +2,7 @@ import logging
 import requests, json, os
 import os, sys, errno
 from .helpers import is_ip_private
+import urllib.parse
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


ben-kaufman:
Looks good, thanks! Maybe @brmdbr or @Clamenda could also test this and see if that solves the issue for them?

stepansnigirev:
Unfortunately this approach doesn't work if `:` is in the username or password.
I suggest slightly different approach - remove credentials from the url and use `auth=(self.user, self.password)` in `requests.post()` method (it is used only in `rpc.multi`, so it's one place change):
```py
        r = self.session.post(
            url, data=json.dumps(payload), headers=headers, timeout=timeout,
            auth=(self.user, self.password)
        )
```

maxidev:
> Unfortunately this approach doesn't work if `:` is in the username or password.
> I suggest slightly different approach - remove credentials from the url and use `auth=(self.user, self.password)` in `requests.post()` method (it is used only in `rpc.multi`, so it's one place change):
> 
> ```python
>         r = self.session.post(
>             url, data=json.dumps(payload), headers=headers, timeout=timeout,
>             auth=(self.user, self.password)
>         )
> ```
Hi Ben, it works form me 🤔 
image

maxidev:
Also tested with : on the user field like so:
image
And still works for me.

stepansnigirev:
Just checked again with the same credential as last time, all good, works fine. Not sure why it didn't work last time...
Sorry for confusion. Looks good to me.

stepansnigirev (approved):

PR #871, opened 2021-01-14
add notes about built in authentication methods

### LINKED ISSUE #862: Update reverse-proxy.md to mention specters user/password auth system
It seems like the document was written before specter had a user/password auth system?

At the moment the document just talks about using basic authentication via the nginx proxy

djpnewton:
also mention rate limiting

fixes #862


k9ert:
Yes, good pointl! Thank you very much.

PR #883, opened 2021-01-17
Specify that hwi is not compatible with Python 3.9

D4nte:


k9ert:
It is specified here:
https://github.com/cryptoadvance/specter-desktop/blob/master/setup.py#L40

What's missing?

k9ert:
Should be catched in
https://github.com/cryptoadvance/specter-desktop/blob/master/setup.py#L40

but that's a valuable addition. Thanks for the PR!

D4nte:
> Should be catched in
> [https://github.com/cryptoadvance/specter-desktop/blob/master/setup.py#L40](https://github.com/cryptoadvance/specter-desktop/blob/master/setup.py?rgh-link-date=2021-01-18T11%3A53%3A33Z#L40)
> 
> but that's a valuable addition. Thanks for the PR!

[`setup.py`](https://github.com/cryptoadvance/specter-desktop/blob/master/setup.py#L40) is not referred in [`./DEVELOPMENT.md`](https://github.com/cryptoadvance/specter-desktop/blob/master/DEVELOPMENT.md)

Should a developer use `setup.py` instead of the current described method?

k9ert:
> Should a developer use setup.py instead of the current described method?

That's not necessary because if you do a `pip3 install -e .` then the `setup.py` is respected as well.

D4nte:
> > Should a developer use setup.py instead of the current described method?
> 
> That's not necessary because if you do a `pip3 install -e .` then the `setup.py` is respected as well.

The issue is that if one only has `python3.9` installed then they cannot go past the `pip3 install -r requirements.txt --require-hashes` step as pip would refuse to install `hwi 1.2.0` due to aforementioned incompatibility.

I could not reach the `pip3 install -e` step without install `python3.8` first and it took me a good 10 min to understand why I could not get `hwi` installed.

k9ert:
I agree, but that's imho not solvable by using setup.py one way or another but just by your PR and adding some documentation --> this PR. Thanks again!

Issue #881, opened 2021-01-17
UX Feedback: "Add Keys" and "Purpose"


D4nte:
I am trying to add and name new account for ledger nano s.

1. Click on "Devices > Ledger Nano S" in side bar
2. Click "Add more keys"
3. Click "Edit"
4. Click "Add custom derivation"
5. Enter derivation path and "purpose"
6. Click "Add"
7. Click "get via USB" to import the xpubs
8. Click Done (see screenshot below)
9. At this point it looks good. I see no way to "finalize" the work so I click back to "Devices > Ledger Nano S".
10. expected: new key is visible. Actual: All changes were reverted (see screenshot)

Edit: I now understand that on step #9 I should click on "Add Keys". This is really unclear (I went trough the process 10 times yesterday and today before realizing that).

I suggest to either change the colour of "Add Keys" to Blue so that it is highlighted as a finalise button.
Or, you could number the steps: 1. edit, 2. Get via USB, 3. Add keys

Also, the "Purpose" is a free text field that can be edited by the user. I tried to use this text as a way to label my keys.
But actually, "Purpose" is the type of addresses to be used with this account (segwit native vs segwit nested). This is also very confusing.

I would suggest to rename it to "Address Type".
Also, I would suggest to not make it an editable free text field as you do not let the user choose it when using ledger nano s.
If there are situations where the user can select what address type they want then put a dropdown list instead.

Looks good once account is added:
![specter-upload-keys](https://user-images.githubusercontent.com/300805/104856289-fad89900-5965-11eb-9f35-f0b135f90fef.png)

Actually no, it disappeared:
![ledger nano s](https://user-images.githubusercontent.com/300805/104856286-f90ed580-5965-11eb-9c43-c5c81484e53f.png)

D4nte:
After more thoughts on the matter, I'd like to propose some improvements.

To import new xpub keys from a device, the user must follow this sequence:
- Specify derivation paths
- Import extended public keys from ledger
- Save the public keys against the device to later create wallets from them

Currently, there are a number of confusing elements with the current screen.

In the device landing screen I can see 5 keys. From this screen I can delete individual keys.
It also shows the wallets using this device. However, it does not show which key is used for which wallet. Knowing that a wallet uses a unique derivation path, it would make sense to put the wallet names and links directly in the table. Today, if I click `remove` I am uncertain whether I am removing a path related to an existing wallet.

[...]

PR #884, opened 2021-01-17
Make `Add Keys` more obvious

D4nte:
It is current not clear that "Add Keys" is the last step of the wizard.

I wonder if there is another issue where the logic to hide/display `wizard-next` and `wizard-submit` is incorrect.
Indeed, should the `Add Keys` button only be displayed once the the xpub keys are loaded?
Looking at the code, it seems that the wizard does not apply to Ledger.

Helps with #881

Before:
specter_desktop_before

After:
specter_after_change


k9ert:
It's a bit difficult to judge what this PR changes effectively. Can you create a screenshot? I gues the current state is visible in 881.
Also pinging @moritzwietersheim to get some input on this.

D4nte:
> It's a bit difficult to judge what this PR changes effectively. Can you create a screenshot? I gues the current state is visible in 881.
> Also pinging @moritzwietersheim to get some input on this.

Sorry! I took a screenshot and forgot to attach it. Now done.

k9ert:
Nice, i like it! I'm wondering whether this is a convention we're using elsewhere as well or we don't (and should consider to document/do that).
CC @GBKS

GBKS:
Definitely a common pattern to have primary and secondary button styles (often also a tertiary style). Primary buttons are usually much easier to pick out (often via a stronger color) and there's just one per page so it's really clear what to do next (like the green "Comment" button under this commenting form).

As far as documenting, style guides are helpful as they allow anyone to survey all UI elements, understand how they should be used and pick the right ones for their needs. For example, see the one for [Bootstrap](https://getbootstrap.com/docs/4.0/components/buttons/) or in the [Wallet UI Kit Figma file](https://www.figma.com/file/VB3GQdAnhl8yta44DY3PSV/Bitcoin-Wallet-UI-Kit?node-id=96%3A2639) I'm working on. For web-based style guides, [Storybook](https://storybook.js.org) is a pretty good solution.

D4nte:
> Nice, i like it! I'm wondering whether this is a convention we're using elsewhere as well or we don't (and should consider to document/do that).
> CC @GBKS

Great! Yes, `btn centered action` is using in other places, that's how I came up with it. I did not want to introduce something new.

k9ert:
Ok, LGTM, will merge soon ...

PR #888, opened 2021-01-19
improve hints for not working connections

k9ert:
Instead of some random Stackoverflow-article have our own article if a connection to a node is broken.
Also have more consistent UI and some obvious more verbose hints if a connection to core does not work.


ben-kaufman:
I've recently added a `<tool-tip>` custom element, would be much nicer to use that instead.
Like this: https://github.com/cryptoadvance/specter-desktop/blob/18b2ad3f3cc69ce56c58a861dfff45c5d89b012e/src/cryptoadvance/specter/templates/settings/tor_settings.jinja#L26

djpnewton:
yeah that stackoverflow article is not very clear

ben-kaufman (approved):

PR #894, opened 2021-01-20
Fixed deprecation warnings

paeet:
This PR fixes several deprecation warnings from the pyTest summary.

### Invalid escape sequence
`specter-desktop/src/cryptoadvance/specter/helpers.py:310: DeprecationWarning: invalid escape sequence \.`

In helpers.is_ip_private, `\.` is considered as an invalid escape sequence. The fix is to parse the regular expression as a raw string, rather than a string literal.

On top of that, I added a unit test for the method.

### collections.Mapping alias will be removed in python 3.9
`specter-desktop/src/cryptoadvance/specter/helpers.py:77: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
    if not isinstance(dv, collections.Mapping):`

https://docs.python.org/3.7/library/collections.html

> Deprecated since version 3.3, will be removed in version 3.9: Moved Collections Abstract Base Classes to the collections.abc module. For backwards compatibility, they continue to be visible in this module through Python 3.8.

### logger.warn deprecation
https://docs.python.org/3.7/library/logging.html#logging.warning
>  There is an obsolete function warn which is functionally identical to warning. As warn is deprecated, please do not use it - use warning instead.


k9ert:
Thanks for the PR. Didn't knew that logger.warn is deprecated now. LGTM!

k9ert:
Awesome, thank you very much for this PR. I have requested whitelisting for your account at travis. Unfortunately our repo is suffering from "abuse-detection" and we need to whitelist every new committter.
However, LGTM.

Issue #896, opened 2021-01-21
fetch from blockexplorer fails


stepansnigirev:
requests_session fails to fetch data for some reason (at least for me and a few other people in the group).
I made an ugly fix here: https://github.com/stepansnigirev/specter-desktop/tree/fix-utxo-explorer

But I think it's time to refactor this thing and fetch blocks from peers instead of the block explorer. Will work on it...

PR #897, opened 2021-01-21
ugly fix utxo blockexplorer rescan

stepansnigirev:
if using requests_session fails - tries the same just with requests

Related: https://github.com/cryptoadvance/specter-desktop/issues/896


ben-kaufman (approved):
Not very pretty, but better ugly and working than pretty and broken :)

PR #898, opened 2021-01-21
Fix typo (puropse -> purpose)

satsie:
Fix typo on the word "purpose" in the 'Add Device' screen.

Screenshot of typo as it is right now:

Screen Shot 2021-01-21 at 2 15 22 PM


ben-kaufman:
Thanks!

PR #899, opened 2021-01-22
UIUX Fixes - TX table toolbar alignment, Network label alignment

paeet:
### TX Table - Page limit & Search bar alignment
The vertical alignment of the page limit and search bar was fixed. Also the horizontal alignment of the search bar seemed to be off center.

On top of that, I fragmentized the page limit control since it was used in multiple places (tx-table.html, address-table.html).

#### Before
Bildschirmfoto 2021-01-22 um 11 41 00

#### After
Bildschirmfoto 2021-01-22 um 12 11 17

### Device Overview, New Wallet - Network label alignment
The vertical alignment of the network label (BTC Main, BTC Test) was fixed. 

Also, since this label is uses in multiple places (device.jinja, new_wallet_keys.jinja), I fragmentized this into a separate component. 

#### Before
Bildschirmfoto 2021-01-22 um 11 54 47

#### After
Bildschirmfoto 2021-01-22 um 11 55 01

Developed and tested on Chromium, Firefox, Safari.


ben-kaufman (approved):
Thanks, looks great!

PR #901, opened 2021-01-22
Removing high level consulting request

chmac:
In discussion with @k9ert, it seems this info is out of date.


k9ert:
Ups, Ben was maybe a bit fast to merge it. I think the text about the strategy can/should be changed because we have a clear strategy on how to be browser-compatible. Webcomponents are compatible with most browsers so, together with templates and the fetch-API, it's no longer [2016](https://hackernoon.com/how-it-feels-to-learn-javascript-in-2016-d3a717dd577f) all that frontend-complexity is gone. The frontend-world is nice again, right?!
image

PR #912, opened 2021-01-27
Fix #605 - Display UI notification after btc core connection test

### LINKED ISSUE #605: Test results: Connectable
I think the language around the connection test results could be made much clearer.

```
Test results: 

Connectable  X
```

Looking at that could be interpreted in several ways. 

1. There were no results since after the : is blank.
2. Test result came back as "connectable"

You really have to decipher that, oh, that X graphic means it's not connectable. 

Something more explicit like:

"Test results: Connection failed."
"Test results: Connection succeeded" 

And if it failed, display the error message(s), like "Process finished with code -1Error message: Failed to connect!:

SPECTE~2

paeet:
The result of a BTC core connection test was not 100% obvious before. You had to scroll down to check the results.

This PR introduces a simple UI notification that gives initial feedback about the test, while keeping the full fledged test results at the bottom of the page.

Bildschirmfoto 2021-01-27 um 11 17 03
Bildschirmfoto 2021-01-27 um 11 16 19

### DIFF
--- a/src/cryptoadvance/specter/server_endpoints/settings.py
+++ b/src/cryptoadvance/specter/server_endpoints/settings.py
@@ -93,6 +93,12 @@ def bitcoin_core():
                 autodetect=autodetect,
                 datadir=datadir,
             )
+
+            if "tests" in test:
+                if not test["tests"]["connectable"]:
+                    flash(f"Test failed: {test['err']}", "error")
+                else:
+                    flash("Test passed", "info")
         elif action == "save":
             if current_user.is_admin:
                 success = app.specter.update_rpc(
--- a/src/cryptoadvance/specter/server_endpoints/settings.py
+++ b/src/cryptoadvance/specter/server_endpoints/settings.py
@@ -95,7 +95,8 @@ def bitcoin_core():
             )
 
             if "tests" in test:
-                if not test["tests"]["connectable"]:
+                # If any test has failed, we notify the user that the test has not passed
+                if False in list(test["tests"].values()):
                     flash(f"Test failed: {test['err']}", "error")
                 else:
                     flash("Test passed", "info")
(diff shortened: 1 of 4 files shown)


src/cryptoadvance/specter/templates/settings/bitcoin_core_settings.jinja
```diff
@@ -31,7 +31,7 @@
 				Bitcoin Core data directory path:<br><input type="text" id="datadir" name="datadir" type="text" value="{{ datadir }}">
 			</span>
 			<br><br>
-			<div id="rpc_settings" style="padding: 10px; border-radius: 10px;">
```
paeet, line 34: Removed the padding to be consistent with the rest of the bitcoin core settings form, incl. the other setting pages. Border radius not needed, since no border exists for the rpc_settings div.

k9ert (approved):

src/cryptoadvance/specter/server_endpoints/settings.py
```diff
@@ -93,6 +93,12 @@ def bitcoin_core():
                 autodetect=autodetect,
                 datadir=datadir,
             )
+
+            if "tests" in test:
+                if not test["tests"]["connectable"]:
```
ben-kaufman, line 98: Why test only connectable? I think it'd make sense to test all... Or maybe have an error message for each stage.

paeet: Good catch! Fixed the issue by keeping the user notification binary, but while considering all the sub tests. Fixed this with the latest commit.

https://github.com/cryptoadvance/specter-desktop/pull/912/files#diff-3a59cc1aa05ae505db526d6d6861af0634ffee79409099c09b97d97a95cbfdeeR99

ben-kaufman (approved):
Looks good, thanks

PR #915, opened 2021-01-28
Support Cobo single key files

### LINKED ISSUE #910: Support Cobo Vault import xpub with microSD
Seems like the format Cobo uses isn't compatible with what we currently support.

Here are example files from Cobo to test with:
[Cobo export xpub.zip](https://github.com/cryptoadvance/specter-desktop/files/5878901/Cobo.export.xpub.zip)

ben-kaufman:
Fix #910


k9ert:
Can't test as i don't have a cobo (yet) but looks good to me!

k9ert (approved):

PR #925, opened 2021-01-31
Fix key initial format in wallet info

### LINKED ISSUE #924: Zpubs are shown, but SLIP132 toggle is off
multisig wallet -> settings -> wallet info -> scroll to bottom click "Show key details" 

on the first open, by default, Specter is showing Zpubs, but the SLIP-132 toggle button above is defaulted to off

so either that toggle should be turned on by default, or Specter should display xpubs instead of Zpubs

ben-kaufman:
Fix #924


djpnewton:
tested ACK

PR #927, opened 2021-01-31
Fix a minor bug that always shows addresses as used

jleo84:
In the line 39 of the file templates/includes/address-data.html , the variable _this.used_ is set as string.
So the conditional in line 85 always returns true and so, any address will be shown as used.


ben-kaufman (approved):
Good catch, thanks!

PR #932, opened 2021-02-02
Add reason for why device is disabled in new wallet screen

### LINKED ISSUE #930: SD 1.1.0: Missing words on "create wallet" page, explanation bottom of page
There seem to be some missing words on the "create wallet" page in SD 1.1.0. "Here we only show and General keys. Follow the standards."
Pls. add missing words, maybe explain (link?) which standards should be followed.
Specter_flaw_Wallet_creation_2021-02-01_20-27-05

### LINKED ISSUE #931: in multisig quorum page, give reason why devices unselectable
took me a couple minutes of scanning around to realize that some devices had been added under regtest, and one on mainnet, so they are showing but greyed out. This should be made more obvious with a tooltip or something.

ben-kaufman:
Add reason for why device is disabled in new wallet screen - fix #931 

Removes broken message which doesn't really seem to be very useful - fix #930


instagibbs:
ACK on words :+1:

PR #938, opened 2021-02-08
Preserve form status when creating a transaction

### LINKED ISSUE #889: preserve form status when creating a new tx
When creating a new tx (https://github.com/cryptoadvance/specter-desktop/blob/2a4450501e899743028f6692e6dca8837cca6607/src/cryptoadvance/specter/server_endpoints/wallets.py#L520) and an error is returned (eg 'invalid address') then the form inputs are returned to their default cleared state. It would be nice if the form status is preserved (recipient, amount, coin selection, fee etc)

We could pass the form state into the final render_template function 

Also we could move the "openpsbt"/"deletepsbt" code to the "send/pending" endpoint and the "importpsbt" code to the "send/import" endpoint to reduce the size of this function (and I think be a bit more logically consistent)

If this seems reasonable I can try and code up a PR

djpnewton:
fixes #889


ben-kaufman:
Thanks! Works really well! Would it be nice if you could just also preserve the unit, at list in the txt option as there it doesn't convert to the right amount of the unit is changed it seems, but would be nice anyway. If you don't have time or something though that's fine, I think it's a great improvement already!

djpnewton:
added preservation of the recipient amount unit

ben-kaufman (approved):
Looks great, thanks!

PR #945, opened 2021-02-12
pass result of createpsbt call back to calculateEstimatedFee

djpnewton:
show the user why the fee estimation failed


ben-kaufman:
Thanks @djpnewton 
There's just a conflict to solve, although I'm not sure why is that change needed?

djpnewton:
fixed the conflict.

the amount sanitation is needed because I was getting the exception string of an invalid "amount" back to the "calculate estimated fee" field. Although that is less likely to occur after a later commit where I use the form validation before calling the fee estimation endpoint

PR #956, opened 2021-02-17
Freeze UTXO and select UTXO for new transaction from the UTXO tab

ben-kaufman:
Adds the ability to freeze and unfreeze any UTXO and do coin selection directly from the UTXO table.
Added a select option on the rows of the wallet UTXO table. Once at least one UTXO was selected there'll appear the 2 buttons for the actions.

Screen Shot 2021-02-17 at 16 41 34
Screen Shot 2021-02-17 at 16 41 18


k9ert (approved):

donnlee:
@ben-kaufman , is this the same functionality as Electrum’s freeze?
“When you freeze an address, the funds in that address will not be used for sending bitcoins. You cannot send bitcoins if you don’t have enough funds in the non-frozen addresses.”

i assume a use-case is to freeze dust utxo so it doesn’t get spent.

ben-kaufman:
Yes that's the same

PR #958, opened 2021-02-18
Use the request session API for authentification

### LINKED ISSUE #929: Specter Desktop 1.1.0: RPC password working only without "special" characters (" ", "_")
V.1.1.0 does not connect to a node with a RPC password containing blancs (reproducible) or underscores ("_"; tested once). V.1.0.0 does. Node is "connectable". Error: "Process finished with code 401Error message: RPC authentication failed!"

juergenhoetzel:
Instead of manually encoding user/password. Fixes #929


k9ert (approved):
Great fix!

PR #968, opened 2021-02-20
Bugfix on missing csrf in js fetch calls, pending send bug

kdmukai:
* JS `fetch` POSTs are failing the `flask_wtf` csrf checks on `combine` and `broadcast`. Adds missing `csrf_token` to the payload.
* Pending psbt signing going nowhere on submit because of the form `action="?"` (but works fine for new psbts). Fixed to explicitly point to the `send/new` endpoint.
* Oddity where HWI signing a pending psbt (KeepKey) was getting misrouted with "/history" improperly appended to url. Did not occur when signing new psbts. Did not occur when signing with bitcoin core hot wallet. Fixed to explicitly point to the `history` endpoint.


ben-kaufman:
Thanks!

PR #989, opened 2021-03-01
show inputs and ouputs that belong to the users wallet

### LINKED ISSUE #955: UI suggestion: Highlight my addresses in Transaction details pane
When viewing the Transaction Details pane, it would be nice if the UI created some sort of visual distinction of which items in the list belong to my wallet. Right now, they all blend in with each other. Originally I thought, "why not just put all of mine at the top of the list" but perhaps there's a good reason why you would want to maintain the default ordering. So instead, perhaps a better solution is a different colored background or some additional icon to make my UTXOs stand out.

Also, along the same lines, if I've created a custom label, the address is no longer visible. It would be nice if both the address and the custom label were displayed.

image

(Not suggesting this is a good background color! It's ugly, I just tweaked the HTML to a random hex for effect.)

djpnewton:
fixes #955

### DIFF
--- a/src/cryptoadvance/specter/server_endpoints/wallets.py
+++ b/src/cryptoadvance/specter/server_endpoints/wallets.py
@@ -1256,6 +1256,7 @@ def addressinfo(wallet_alias):
                 "address": address,
                 "descriptor": descriptor,
                 "walletName": wallet.name,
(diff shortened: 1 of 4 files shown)


ben-kaufman (approved):
Tested, works great, thanks!

PR #991, opened 2021-03-02
Add "Abandon transaction" option for low fee txs that have been purged from the mempool

### LINKED ISSUE #951: Pending Txn is "Stuck" - Can't Spend UTXO
Hey team, having an issue with a transaction from over a month ago. 

- Fee was too low so didn't get confirmed
- Cleared from the mempool (see blockstream.info, no txn)
- Specter hasn't cleared it yet so I can't recreate the txn with the same utxo
- RBF "Speed Up" doesn't work

Version:
- Currently 0.10.1 but this was happening on prev version as well
- Running on MyNode Raspi4, viewing on macOS

Any thoughts on how to fix this?

Screen Shot 2021-02-15 at 5 09 28 PM
Screen Shot 2021-02-15 at 5 11 09 PM
Screen Shot 2021-01-19 at 9 45 16 AM

kdmukai:
Fixes #951

abandon_tx_screenshot

Changes/Rationale:
`DEVELOPMENT.md`:
* Additional notes for running bitcoind manually.
* minor reorg.

`bitcoind.py`:
* Now takes optional `extra_args` when starting bitcoind to allow for setting `maxmempool` or other future test case needs.
* Implement stop_bitcoind for BitcoindPlainController. Running the tests against local bitcoind was leaving the process running.

`bitcoin_core.py`:
* Adds `sign_raw_tx` helper for the test case. Parallels `sign_psbt`.

`wallets.py`:
* Adds `is_purged` check when loading the tx details popup.
* 'Abandon transaction' button posts to `history` endpoint to keep the flow simple (no Ajax back to the tx details popup).

`tx-data.html`:
* The "Abandon transaction" text tries to strike a balance between being informative for less savvy users and guide them on when to properly use the feature AND be as brief as possible! I did my best!

~~`wallet-history.jinja`:~~
~~* I needed the `csrf_token` so I can do the POST on the "Abandon transaction" click.~~

`wallet.py`:
* The additional `if "address" in tx` check was necessary to get through the test case. I think the bitcoin core method of generating large transactions makes them nonstandard (also why we pass in `acceptnonstdtxn=1` when starting bitcoind)?

`test/bitcoin_core/*`
* See the README added there.

`conftest.py`:
* Factored `instantiate_bitcoind_controller` out of `bitcoin_regtest` so it could be re-used with different `extra_args` as needed for this test and potentially future tests.
* Pass `rpcport` into `BitcoindPlainController` so it can differentiate between the default port and the one-off instance used in the test.

`test_wallet_manager`:
* First two tests were failing against a local bitcoind but passing with the docker version. Changed the tests to use the new `instantiate_bitcoind_controller` so they can run on a completely fresh, isolated bitcoind.

HOW TO TEST ON MAINNET:
Restart your node. The mempool should reinitialize `mempoolminfee` to 0.00001000 (check via `bitcoin-cli getmempoolinfo`). Create a transaction with 1sat/vB fee and broadcast it. If there are enough txs already in other nodes' mempools, odds are that they will ignore your tiny transaction for violating their `mempoolminfee`. Hit a public block explorer to confirm that your txid never landed. But the tx will still be in your node's mempool.

Restart your node again but this time pass in `-maxmempool=5` to severely restrict the size of your mempool. Watch the `mempoolminfee` creep up as the node fills its limited mempool with just the highest-fee txs. Anything above 0.00001000 should prompt your node to purge your tiny tx. Specter should now show you the "Abandon transaction" option.

Click to abandon the tx to clear it out of your wallet. Restart your node with normal settings. Send a new test tx using the same coins but with a reasonable fee for the current mempool. The abandoned tx should no longer appear in your history, just the new tx. This final tx should land a confirmation eventually and get updated in Specter accordingly.

FOLLOW-UP WORK TODO:
The abandoned tx still shows up on the history page. A different icon or status indicator would be useful in the list as well as in the `decoderawtx` details popup. The abandoned tx disappears once its coins are spent in a new tx.


src/cryptoadvance/specter/templates/includes/tx-data.html
```diff
@@ -67,7 +67,26 @@ <h2>Transaction details</h2><br>
...
+                if(tx.is_purged) {
+                    rawtxHTML += `
+                    <div class="warning">
+                        <p><b>This transaction (aka "tx") could not be found in your node's mempool!</b></p>
+                        <p style="text-align:left;">If the bitcoin network is very busy, nodes will start purging the pending txs with the lowest fees. Txs older than two weeks are also purged.</p>
+                        <p style="text-align:left;">A purged tx will never complete. Abandoning this tx will clear it from your wallet, making those funds spendable once again.</p>
+                        <p style="text-align:left;">But first verify that other nodes have also purged your tx! Enter the tx id into a public block explorer (warning: slight privacy leak). If the tx cannot be found, it is safe to abandon this tx.</p>
+
+                        <br/><br/>
+                        <form method="POST" action="${walletLink}history">
```
ben-kaufman, line 80:
```suggestion
                        <form method="POST" action="${walletLink}history/">
```
This seems to return an error page without the trailing /

src/cryptoadvance/specter/templates/includes/tx-data.html
```diff
@@ -67,7 +67,26 @@ <h2>Transaction details</h2><br>
...
+                            <input type="hidden" class="csrf-token" name="csrf_token" value="${csrf_token}"/>
```
ben-kaufman, line 81: Doesn't really matter, but are you sure it's not accessible simply via `"{{ csrf_token() }}"` and have to be added to the `wallet_history.jinja`?

ben-kaufman:
Looks really good! Needs a small rebase and left 2 small comments, but still need to go through it a bit more thoroughly.

src/cryptoadvance/specter/templates/includes/tx-data.html
kdmukai: Ah, I should have tried the obvious. I had assumed the naming convention implied that the *.html were being handled as static (as opposed to the *.jinja templates) but didn't notice that there were of course template injections happening in there already.

src/cryptoadvance/specter/templates/includes/tx-data.html
kdmukai: Passes either way on Brave/MacOS, but updated in latest commit.

k9ert:
@kdmukai thank you very much for this PR. You really uped the testing-play here and this is a really useful feature avoiding a lot of confusion!

PR #997, opened 2021-03-03
Cypress travis

k9ert:
This PR:
* Moves the the OS which is used by Travis-CI from xenial (16.04LTS) to focal (20.04LTS) 
* Runs cypress-tests in travis after the pytest
* Improves and speed -up the bitcoin-core-dependency mechanism

An average build is now taking 7 mins compared to 5 mins build before. A build where the bitcoin-core-version is changed (pytest.ini) takes now 17 mins (compiling berkleyDB and core itself).

### DIFF
--- a/tests/install_bitcoind.sh
+++ b/tests/install_bitcoind.sh
@@ -7,17 +7,19 @@ bitcoind_setup_needed=false
 if [ ! -d "./bitcoin/.git" ]; then
     echo "    --> cloning bitcoin"
     git clone https://github.com/bitcoin/bitcoin.git
-    cd bitcoin
+
     bitcoind_setup_needed=true
-else
-    cd bitcoin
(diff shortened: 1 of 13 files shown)


tests/install_bitcoind.sh
```diff
@@ -7,17 +7,19 @@ bitcoind_setup_needed=false
...
+
     bitcoind_setup_needed=true
-else
-    cd bitcoin
-    echo "    --> fetching bitcoin"
-    git fetch
+#else
+    # do not unnecessarily fetch bitcoin as this affects the cache
+    #echo "    --> fetching bitcoin"
+    #git fetch
```
ben-kaufman, line 15: Maybe delete if not needed?

ben-kaufman (approved):

PR #998, opened 2021-03-05
Add full edit transaction for RBF

ben-kaufman:
Add an option to do full edit of an RBF transaction. This will open the transaction editor with the original transaction details and the new selected fee from the popup and will allow user to customize the transaction completely. There are currently no checks in the PR on the transaction before attempting to broadcast, which means it could potentially fail (for example due to absolute fee too low), or be edited to not be rbf (if user unchecks all the inputs for example).

Also fixed typo in parts of the code (settigns -> settings)


k9ert:
:+1:  great feature!

PR #1002, opened 2021-03-06
Keep local dev .specter data separate

kdmukai:
I use my laptop for local dev as well as real-world Specter/bridge mode use. Local dev is often set up to connect to a local bitcoind. Real-world use connects to my node on a different machine. Adding this to make it easier to keep my worlds separate.


kdmukai:
Don't know why the Black formatting failed. Closing this and resubmitting as #1003.

PR #1003, opened 2021-03-06
Keep local dev .specter data separate

kdmukai:
_(originally tried to submit this as a simple patch PR just using the github UI. Dunno why the Black formatting pass failed. Trying again here)_

I use my laptop for local dev as well as real-world Specter/bridge mode use. Local dev is often set up to connect to a local bitcoind. Real-world use connects to my node on a different machine. Adding this to make it easier to keep my worlds separate.


k9ert:
Thanks, very reasonable.

PR #1021, opened 2021-03-17
Add currencies to the price provider

ben-kaufman:
- Add more fiat currencies to the price provider
- Add precious metals
- Add weight unit selection for precious metals
- Fix direction of symbol


k9ert:
If the list would grow constantly from here, i'd probably request a different implementation for the price-providers reducing these tedious if/else-statements. But i think we're fine here for the release and can postpone it for after the release.

PR #1032, opened 2021-03-20
Update DEVELOPMENT.md

kdmukai:
On a brand new Raspberry Pi OS image `libffi-dev` and `libssl-dev` need to be installed otherwise pip installing `cffi` and `cryptography` will fail.


k9ert:
cool, thanks!

PR #1044, opened 2021-03-22
Default bitcoind timeout to 60s for all platforms

kdmukai:
Timeout was expanded specifically for under-powered Raspberry Pis but just ran a test with a 2010 Macbook Air that also needs more than 30s to spin up bitcoind and respond. With my local code updated to 60s the Macbook Air was able to get bitcoind to respond in time.

I don't see any reason not to just default to 60s for everyone.

_Note/future TODO: If you successfully install bitcoin core through the UI but the `wait_for_bitcoind` times out, the wizard does not know to resume from the `start_bitcoind` call and instead starts doing the QuickSync download again. Current workaround is to go to settings and turn on the built-in node. But if the device failed the `wait_for_bitcoind` timeout once, I'd guess it'll keep failing until the timeout is expanded. Might need a "try again" option that extends the timeout even further._

### DIFF
--- a/src/cryptoadvance/specter/bitcoind.py
+++ b/src/cryptoadvance/specter/bitcoind.py
@@ -97,7 +97,7 @@ class BitcoindController:
         cleanup_hard=False,
         datadir=None,
         extra_args=[],
-        timeout=30,
+        timeout=60,
     ):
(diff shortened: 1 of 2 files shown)


k9ert:
Thanks, merging!

PR #1055, opened 2021-03-25
Pre release minor fixes

ben-kaufman:
- Raise error to ensure Specter stops after cleanup
- Stop wallet manager `is_loading` flag if update thread wasn't called
- Improve hot wallet creation error handling

### DIFF
--- a/src/cryptoadvance/specter/server_endpoints/devices.py
+++ b/src/cryptoadvance/specter/server_endpoints/devices.py
@@ -16,6 +16,7 @@ from mnemonic import Mnemonic
 from ..helpers import is_testnet, generate_mnemonic
 from ..key import Key
 from ..wallet_manager import purposes
+from ..specter_error import handle_exception
 
 rand = random.randint(0, 1e32)  # to force style refresh
 
@@ -142,6 +143,7 @@ def new_device():
                         + "?newdevice=true"
                     )
                 except Exception as e:
+                    handle_exception(e)
                     flash(f"Failed to setup hot wallet. Error: {e}", "error")
                     app.specter.device_manager.remove_device(
                         device,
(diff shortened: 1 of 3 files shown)


src/cryptoadvance/specter/server_endpoints/devices.py
```diff
@@ -124,22 +124,31 @@ def new_device():
...
+                        keys_range=[range_start, range_end],
+                        keys_purposes=keys_purposes,
+                    )
+                    flash("{} was added successfully!".format(device_name))
+                    return redirect(
+                        url_for("devices_endpoint.device", device_alias=device.alias)
+                        + "?newdevice=true"
+                    )
+                except Exception as e:
+                    flash(f"Failed to setup hot wallet. Error: {e}", "error")
```
k9ert, line 145: I think the rule should be: If you catch a SpecterError, do that flash and move on. However if you catch an Exception, then you don't know what the real cause of that issue is, so you should make sure the stacktrace ends up in the log.
```suggestion
                    handle_exception(e)
                    flash(f"Failed to setup hot wallet. Error: {e}", "error")
```
(And import it)

src/cryptoadvance/specter/server_endpoints/devices.py
```diff
@@ -209,19 +218,28 @@ def new_device_manual():
...
+                        file_password,
+                        app.specter.wallet_manager,
+                        is_testnet(app.specter.chain),
+                        keys_range=[range_start, range_end],
+                    )
+                    return redirect(
+                        url_for("devices_endpoint.device", device_alias=device.alias)
+                    )
+                except Exception as e:
+                    flash(f"Failed to setup hot wallet. Error: {e}", "error")
```
k9ert, line 236: Same here, please.

src/cryptoadvance/specter/specter.py
```diff
@@ -234,6 +234,7 @@ def cleanup_on_exit(self, signum=0, frame=0):
             self._bitcoind.stop_bitcoind()
 
         logger.info("Closing Specter after cleanup")
+        raise SpecterError("Specter closed")
```
k9ert, line 237: What's the purpose of this? Maybe you can add a comment here, looks very unconventional.

PR #1062, opened 2021-03-28
Pre download binaries

ben-kaufman:


k9ert:
to be checked with the next pre-release.

PR #1080, opened 2021-03-31
Use Bitcoin icon set

### LINKED ISSUE #1051: Consider using the Bitcoin icon set
I am working on the Bitcoin icon set project ([Github](https://github.com/Bosch-0/Bitcoin-Icons), [Figma](https://www.figma.com/community/file/948545404023677970/Bitcoin-icon-set)), which is an open-source/open-design effort to create a collection of icons for use in Bitcoin-related applications. It's a very young project that is just getting into a usable state (for a larger project overview, view [this video](https://youtu.be/hgZsRffc52Q)). It would be fantastic if we could evaluate whether it's a good fit for Specter.

To facilitate this, I mapped all the current icon in Specter against the icon set and filled in any missing ones. [Here's a short video](https://youtu.be/W4h59fTExzc) with an overview, and [the Figma file](https://www.figma.com/file/hooBxH1hB5pbHMM1V09frt/Specter-icon-map?node-id=2%3A1812) from the video.

Switching is ideally as easy as copy/pasting the SVG icons from the [repo](https://github.com/Bosch-0/Bitcoin-Icons/tree/main/svg) and updating a few file names. But I think in reality there's more development work to fine-tune the sizing, colors and positioning of icons in different components. Since this would be the first implementation of the icons, I can also imagine that we come across some minor surprises. But since they are just icons, the complexity should still be straightforward.

If this is something you're interested in, I'm happy to get hands-on and help sort out the details. Just let me know.

ben-kaufman:
Close #1051 

I've switched icons we currently have with the icons from https://github.com/BitcoinDesign/Bitcoin-Icons according to the Figma: https://www.figma.com/file/hooBxH1hB5pbHMM1V09frt/Specter-icon-map?node-id=0%3A1

(All except for the warning sign icon, which would be slightly more complicated to change properly)


djpnewton:
there are some other parts of the UI that use the colors of the current receive and send icons.

the colors are: #131a24 (send) and #925d07 (receive)

perhaps we should refactor them into some constants so that they can be easily kept in sync with other UI changes

k9ert:
I've tested this and i like the new icon-set. So i'd suggest to merge this. 
@djpnewton yes i agree to your point as well but we could do this in another PR and i already had to resolve merge-conflicts here.

I will merge tomorrow if there are no other comments here.

djpnewton:
ah actually testing this and there are css classes to make the icons the same color as the previous

looks good

k9ert:
According to https://www.githubstatus.com/ GithubActions is currently "degraded". But i don't have many doubts that the build would be completely green as we only had merges recently.
Merging!

PR #1081, opened 2021-03-31
Renderer POC

k9ert:
This is a small POC and WIP to explore the use of renderers for devices which should simplify the frontend-logic.

```
pytest tpytest tests/test_devices_renderer.pyests/test_devices_renderer.py
```
This will fail but only due to `assert False` (which i like to use in order to see output).

The statement in `devices.jinja` won't work as the Device-Class is in `cryptoadvance.specter` and not in `cryptoadvance.specter.devices`. 
Should be moved one level lower.

I've chosen the FunctionLoader as this was easier in the short-term. In probably make sense to store them eventually in files.

### DIFF
--- a/src/cryptoadvance/specter/device.py
+++ b/src/cryptoadvance/specter/device.py
@@ -32,6 +32,30 @@ class Device:
         self.keys = keys
         self.fullpath = fullpath
         self.manager = manager
+        self.renderer = self.get_renderer()
+
+    def get_renderer(self):
+        renderer_package = (
+            "cryptoadvance.specter.devices.renderer." + self.__class__.name + "Renderer"
+        )
+        components = renderer_package.split(".")
+        mod = __import__(components[0])
+        try:
+            for comp in components[1:]:
+                mod = getattr(mod, comp)
+            return mod
+        except AttributeError:
+            # ToDo: make a more sophisticated Lookup-mechanism
+            renderer_package = "cryptoadvance.specter.devices.renderer.DeviceRenderer"
+            components = renderer_package.split(".")
+            mod = __import__(components[0])
+            for comp in components[1:]:
+                mod = getattr(mod, comp)
+            return mod(self)
+
+    def render(self, name):
+        print(self.renderer)
+        return self.renderer.render(name)
 
     def create_psbts(self, base64_psbt, wallet):
         """
--- a/src/cryptoadvance/specter/devices/__init__.py
+++ b/src/cryptoadvance/specter/devices/__init__.py
@@ -9,6 +9,7 @@ from .cobo import Cobo
 from .generic import GenericDevice
 from .electrum import Electrum
 from .bitcoin_core import BitcoinCore
+from .renderer import DeviceRenderer
 
 # all device types
 __all__ = [
--- a/src/cryptoadvance/specter/devices/bitcoin_core.py
+++ b/src/cryptoadvance/specter/devices/bitcoin_core.py
@@ -1,7 +1,7 @@
 import os
 import shutil
 from embit import bip39, bip32, networks
-from ..device import Device
+from .device import Device
 from ..helpers import alias
 from ..util.descriptor import AddChecksum
 from ..util.base58 import encode_base58_checksum, decode_base58
--- a/src/cryptoadvance/specter/devices/device.py
+++ b/src/cryptoadvance/specter/devices/device.py
@@ -1,8 +1,8 @@
 import json
-from .key import Key
-from .persistence import read_json_file, write_json_file
+from ..key import Key
+from ..persistence import read_json_file, write_json_file
 import logging
-from .helpers import is_testnet
+from ..helpers import is_testnet
 
 logger = logging.getLogger()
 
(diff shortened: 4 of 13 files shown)


k9ert:
It didn't looked that bad actually until some point. Here is a summary.
So with the Device-instantiation you look up a specific renderer for that Device. If the lookup-mechanism is clever (not implemented) then you could follow the hierarchy of the device and take the best renderer available if one specific for your device-type is not available. The renderer is initialized with a reference back to the device.
Then you implement a render-method in your Device which defers the rendering to its renderer:
```
    def render(self, name):
        return self.renderer.render(name)
```
which itself uses a jinja2-environment which only inject the device as variable.
You have some naming conventions for `name` which makes clear what's now rendered, e.g.:
```
    def template_multi_add_keys(self):
        return """
			<button id="add_keys" type="submit" name="action" value="add_keys" class="btn centered">Add more keys</button>
        """
```

Depending on the capabilities of the device, you can either not implement certain functions (which returns "") or maybe even check for device-capabilities and then return something (or something else or nothing).

Potentially you can push that stuff to files and render it from there to have syntax-highlighting, proper-parsing and the like.

So far so good but then you have e.g. a snippet like this:

```
			<tbody>
				{% for key in device.keys %}
					<tr>
						<td>
						    {% include "components/network_label.jinja" %}
						</td>
						<td>{{ key.purpose }}</td>
						<td class="optional">{{ key.derivation }}</td>
						<td width="15" onclick="showPageOverlay('key_export_qr');document.getElementById('key_selected').value='{{loop.index0}}';toggleKeyDisplay();" title="Show key QR code" style="cursor: pointer;">
							<img style="margin-top: 5px;" width="15" src="{{ url_for('static', filename='img/qr_tiny.svg') }}"/>
						</td>
						<td class="xpub scroll optional">
							<span class="explorer-link" onclick="copyText('{{ key }}', 'Copied key: {{ key }}')">{{ key }}</span>
						</td>
						<td width="80px" class="optional">
							<form action="./" method="POST">
								<input type="hidden" class="csrf-token" name="csrf_token" value="{{ csrf_token() }}"/>
								<input type="hidden" name="key" value="{{ key.original }}">
								<button type="submit" name="action" value="delete_key" class="btn danger hovering">Delete</button>
							</form>
						</td>
					</tr>
				{% endfor %}
			</tbody>
```
So here, you have at least one, probably two issues at hand:
* the `csrf_token` is only valid in the "big-flask-environment"
* the include-statement is also for now available in the big-flask-env

So maybe the include statement might be solvable by a [ChoiceLoader](https://jinja.palletsprojects.com/en/2.11.x/api/#loaders) but the `csrf_token` is creating headache. How should that value get injected into the template?
Neither a custom [filter](https://flask.palletsprojects.com/en/1.1.x/templating/#registering-filters) nor a [context-processor](https://flask.palletsprojects.com/en/1.1.x/templating/#context-processors) seem to help here.

That ended quickly. Any feedback (also with very different ideas than the one hers) very much appreciated.

ben-kaufman:
Looks like you completely deleted the toggle passphrase button and now going to trezor shows 2 sign message buttons instead...

kdmukai:
I don't think I'm fully appreciating the goal here. The if/else branching in the template seems fine as-is without pushing the device-specific template code into python (which I am strongly against) or into some separate template snippet it might import (better). Some of my python--and especially django--projects get so micro-sliced up that it becomes a huge pain to track down where a tiny bit of functionality actually lives.

I don't find the front-end logic messy enough. One area that IS absolutely killing me is the Add Device wizard code (getting lost in JS messiness is basically the last hurdle preventing me from finishing off the hot wallet restoration PR). By comparison the templates' device if/else checks seem straightforward and a non-issue.

But again, I'm not sure I fully understand the pain point you're hoping to solve here.

k9ert:
The reason for this is probably, that we had that rendering-idea as a potential solution to problems like "complicated business-logic-containing templates".

I also agree to your critics about micro-slicing, although it's also partially a trade-off-decision. The complexity of ever growing new features need to go somewhere. 

Anyway, most valuable is imho your hint where the code is most confusing. I've also chose the device-topic as i can only refactor what i'm understanding. That's why places sometimes gets avoided although it would be most beneficial to target exactly those places.

Point taken, i'll probably will have a look at the wizard trying to find ways in order to clean that up a bit better. Let's still keep this open for some more time for potential more feedback (and at least until i have another one open targeting [code-smells](https://en.wikipedia.org/wiki/Code_smell))

stepansnigirev:
Text templates in the code is what I really don't like in this PR. I would like to avoid this kind of things at all costs, because if the jinja code is in python it becomes a pain to find proper place in the code, also code highlighting doesn't work so it's easy to make mistakes there.

Is there a way to put jinja templates in the `devices/` folder instead? If flask doesn't allow loading templates from other folders we could place them in `templates/devices` or read the template string from file in the python code instead of hardcoding it.

kdmukai:
@k9ert yeah the problem with the wizard (and really we should move this part of the discussion into its own issue) is that it uses the DOM state to keep track of which steps are required (some devices need, say, 2 steps, some need 3) and which step you're on. It then switches UI components on and off accordingly. Updating a device has slightly different DOM state implications than Create.

It'd be 1000x more straightforward to refactor each step as its own really simple url endpoint that passes state from request to request. Each endpoint can then handle the logic to figure out the next step in the wizard to route to.

For example: Need to deal with the mnemonic for a hot wallet? Get routed to wallets/new/hotwallet/mnemonic or whatever and resume the flow from that POST.

For now I'm just drowning in DOM state and struggling to manage elements that are hidden or not hidden based on that state. I worked through most of the issues but don't have high confidence in it. I'm more at the "guess-and-check and, oh, look, I think it's doing the right thing" stage as opposed to the "I've got this figured out and it all makes sense" stage. Since we're not writing a JS SPA, atomize it all into distinct request/response cycles to isolate simple pieces of functionality!

k9ert:
@kdmukai yes, very good points. i think Ben is working on the wizard in #1111 and i'm currently focusing on #1106 when i'm not trying to fix the CI-system. As it got worse and worse, i spend the last days to move over to cirrus which imho works much better now.

k9ert:
Closing. This was a nice exercise but i don't think it's a good way to go.

PR #1085, opened 2021-04-01
Add unconfirmed tx fees data 

### LINKED ISSUE #1038: No fee rate in transaction detail (incoming transaction)
When I view an outgoing transaction detail, there is the "fee" and "fee rate".  Example: 15.7 sat/vbyte.
However, when I view an incoming transaction detail (in the same way by clicking the transaction id), the fee and fee rate are missing.
If this is easy to add, I would be willing to create a PR.  I just want to confirm it is a known behavior and understand what steps would have to be added (eg. if additional RPC requests about the inputs need to be sent to bitcoind).

My Specter: v1.1 linux .appimage

ben-kaufman:
Fix #1038 (for unconfirmed transactions, but that's all the fees are actually relevant for anyway)


src/cryptoadvance/specter/server_endpoints/wallets.py
```diff
@@ -1067,6 +1067,15 @@ def decoderawtx(wallet_alias):
 
             if tx["confirmations"] == 0:
                 tx["is_purged"] = wallet.is_tx_purged(txid)
+                try:
+                    if wallet._transactions[txid].get("category", "") == "receive":
+                        tx["fee"] = (
+                            wallet.rpc.getmempoolentry(txid)["fees"]["modified"] * -1
+                        )
+                except Exception as e:
+                    app.logger.warning(
```
k9ert, line 1076: can you either do one of:
* Catch a more specific Exception (`RpcError`)
* `handle_exception(e)` 
* `logger.exception(e)`

This combination is imho suboptimal.

ben-kaufman: Added handle_exception(e)

PR #1086, opened 2021-04-01
Fix crash if estimatesmartfee fails

ben-kaufman:
Use 1 sat/vbyte if fee estimation fails completely. 

Not sure that's the best behaviour though... Maybe force the user to enter manual fee would make more sense?


k9ert:
I think we need to flash that issue to the user. If we simply swallow the issue then this has some potential in the future for a lot of wasted time.

PR #1087, opened 2021-04-02
Show user-friendly error messages when session expires

ben-kaufman:
Show user-friendly error flash message asking to refresh the page instead of scary CSRF/ 405 error pages if the session CSRF token or authentication token have expired/ changed/ not properly sent.


src/cryptoadvance/specter/server_endpoints/wallets.py
```diff
@@ -1100,13 +1100,19 @@ def rescan_progress(wallet_alias):
...
-    }
+    try:
+        wallet = app.specter.wallet_manager.get_by_alias(wallet_alias)
+        address = request.form.get("address", "")
+        label = wallet.getlabel(address)
+        return {
+            "address": address,
+            "label": label,
+        }
+    except Exception as e:
```
k9ert, line 1111: Can you please log the exception with e.g. handle_exception? Rule of thumb should imho be: If you threw the exception yourself, you probably don't neeed a stacktrace in the logs. But if you catch an exception, you probably don't know much about the error, so please then make sure to report it in the logs.

k9ert:
Tested, LGTM apart from that one comment. After that ready to merge.

src/cryptoadvance/specter/server_endpoints/wallets.py
ben-kaufman: Added

PR #1088, opened 2021-04-02
Add full address next to label in signing screen

### LINKED ISSUE #987: UI Problem - Labeled Address not Displayed During Confirmation
If you have already labeled your address prior to initiating a transaction with it, when the application pops the modal to confirm the transaction on the hardware device, the address is no longer displayed anywhere, making it difficult to confirm that the address on the device matches the address in Specter (see screenshot).

I suggest displaying both the label and the address and ensuring that the device interface modal does not obscure it, or display the address directly in the device interface modal.

Specter Desktop - No Display of Target Address for Verification - Screenshot 2021-02-25 142334

ben-kaufman:
Fix #987

Show the full address itself in addition with the label in the PSBT sign screen


k9ert:
Tested a LGTM. Ready to merge!
image

djpnewton:
huh, tested just now too

PR #1090, opened 2021-04-03
Fix BitBox02 timeout issue

ben-kaufman:
Fix issue where the request to the BitBox02 throws error due to the timeout.


ben-kaufman:
Note that I couldn't reproduce the issue myself, but got one user so far confirming this works.

djpnewton:
does the timeout happen while unlocking the bitbox02? I might have hit this issue too, so I could test this fix

ben-kaufman:
@djpnewton , not exactly, the issue is that it seems to happen to some users even after unlocking the BitBox02. Could not reproduce myself, but if you can it'll be very helpful!

k9ert:
I've created https://github.com/cryptoadvance/specter-desktop/releases/tag/v1.3.1-pre2 in order to test this PR.

PR #1093, opened 2021-04-04
Use RPCAuth instead of username + password

ben-kaufman:
Use `rpcauth` instead of deprecated `rpcuser` + `rpcpassword` for the internal node setup.

I'm thinking whatever it would make sense to force change `config["internal_node"]["autodetect"]` to be `False` in existing internal_node setups? If it won't make problems, and I see no reason it would, I guess it would make sense. Is there a reason not to?


k9ert:
Tested, LGTM. Ready to merge.

PR #1095, opened 2021-04-04
Fix truncated PDF

### LINKED ISSUE #1035: Save Backup PDF for multisig : truncated Master public key
The PDF backup of a multisig wallet could be unusable when a master public key is truncated.

This split (in wallet_pdf.jinja) for master public key use 40 which is too large :
`.split(/(.{40})/g);`

I see two solutions :
- use a monospace font to get a constant and testable length.
- keep a sans-serif font but calibrate the split value with the widest letter (I ofen use M or W for that purpose).

If you're not aware of the sans-serif character length variance, look here : https://stackoverflow.com/questions/3949422/which-letter-of-the-english-alphabet-takes-up-most-pixels

In the case of my unusuable backup, a split of 38 with the current font should work (two characters are truncated with 40), but in the worst case with only W, it's certainly a smaller number.

ben-kaufman:
Fix #1035

Using Roboto Mono font with line breaking adjustments.


src/cryptoadvance/specter/templates/wallet/components/wallet_pdf.jinja
```diff
@@ -6,6 +6,9 @@
         const doc = new jsPDF();
         const pageHeight = 297;
         const pageWidth = 210;
+        doc.addFont("{{ url_for('static', filename='fonts/RobotoMono-Regular.ttf') }}", "Roboto", "normal");
```
djpnewton, line 9: does this embed the actual font in the resultant PDF document? or link?

k9ert: It seems it's embedding it. see https://github.com/MrRio/jsPDF#use-of-unicode-characters--utf-8 or https://stackoverflow.com/questions/26908266/custom-font-faces-in-jspdf

k9ert:
Tested it and LGTM. Ready to merge. Let's wait 2 more days for a potential comment from @djpnewton and then merge.

src/cryptoadvance/specter/templates/wallet/components/wallet_pdf.jinja
djpnewton: cool thats good

PR #1102, opened 2021-04-09
fix typo/exception

djpnewton:
>UnboundLocalError: local variable 'se' referenced before assignment


djpnewton:
I also updated `server_rpc_error` so it actually shows the user the error message

k9ert:
Yes, good point! Thank you for the PR. Looks good to me. 

Unfortunately we're suffering of permanent abuse-suspection. The repo has code in here which triggers travis's abuse-detection-code. That's the reason why the tests are not running for your PRs. People can be whitelisted and i just wrote a mail to travis-support to do that for you.

Let's give them some time and when the tests are green, let's merge this!

djpnewton:
> People can be whitelisted and i just wrote a mail to travis-support to do that for you.

Awesome thankyou

k9ert:
It's horrible. I just discovered that i already ask for whitelisting of @djpnewton back in february. However, the flagged you again. And me as well! I asked again today and got the same answer with a question which i don't know how often i already answered:

> Thanks for the email and apologies for the issue.
> 
> This is a part of our abuse prevention system, which blocks a repo if it looks like someone is trying to mine cryptocurrency. It is an evolving system, as we work to ensure our service is reliable and performant for the community and our customers. We make every attempt to avoid false positives, but they do happen, and we know this can cause disruptions in your work. Could I ask how you're using Travis for this repository, and how you're using cryptocurrency in relation to Travis?
> 
> Look forward to hearing from you.

k9ert:
Thank you and sorry it took so long for such a reasonable Changerequest. Will be faster next time.

PR #1106, opened 2021-04-09
refactor check_methods out of specter

k9ert:
This is a refactoring PR, WIP.
The goal is to make the specter-class much slimmer and move business-code to the edges. That means that we create more managers like the `WalletManager`, the `DeviceManager` and the `UserManager` which are hopefully quite distinct from one another and the `Specter` class ideally becomes a stupid bracket-object which makes them accessable for other parts of the code (mostly the controllers).
So now i did the following:
* Removed the otp-logic (just three messages) out from specter and in it's own class: `OtpManager`
* `OtpManager` is derived from `GenericDataManager` which represents a manager with a `load()` and a `save()` method which is managing data in a json-file. The data is available in the data-property. Perfect to be derived from more specific classes like `OtpManager`
* One of the biggest functionality is also dealing with the config.json file. So in parallel to the above `OtpManager`, i've created a `ConfigManager` derived from `GenericDataManager` and moved all these update-methods over to that manager. 
* However, i think it's a good idea to not make the managers dependent from one another or at least avoid that as much as possible. So right now, the `ConfigManager` is not dependent on the specter-object or any other Manager.
* Other then the `OtpManager`, i did not modified the places where it's called. So right now i left all the methods in there but deferred the logic to the `ConfigManager` like:
```
    @property
    def config(self):
        """A convenience property simply redirecting to the config_manager"""
        return self.config_manager.data

    def update_fee_estimator(self, fee_estimator, custom_url, user):
        """ update the fee estimator option and its url if custom """
        self.config_manager.update_fee_estimator(fee_estimator, custom_url, user)

# many more update_ ... methods
```
In some cases, this pattern only works partially. As the ConfigManager is free of any dependencies but some methods had side-effects which needs specter-members, some method got split up between `Specter` and `ConfigManager`. E.g.:

```
    def update_rpc(self, **kwargs):
        need_update = self.config_manager.update_rpc(**kwargs)
        if need_update:
            self.rpc = get_rpc(
                self.config_manager.rpc_conf,
                None,
                proxy_url=self.proxy_url,
                only_tor=self.only_tor,
            )
            self._save()
            self.check(check_all=True)
        return self.rpc is not None
```

Even though the logic has now been mostly removed from `Specter` and `ConfigManager` has now 340 lines of code, we're still not in an optimal position. IMHO there are 2-3 options going forward:
* We're removing the method-signatures from `Specter` and putting all the logic directly into the managers. This would mean that we need to carefully manage the dependencies between the different managers. Should the `WalletManager` be dependent on the `ConfigManager` or the other way around? We should avoid circular dependencies under all circumstances!
* We keep it like this which might double the method-signatures e.g. in `Specter` and `ConfigManager` but it also makes it easier to deal with the dependencies and still makes `Specter` class much slimmer. 
* Additional to this, we could potentially use [Mixin-classes](https://www.residentmar.io/2019/07/07/python-mixins.html) to separate the different aspects from one another. So each Manager __might__ get a `SpecterSomethingMixin` which contains most of the methods of the Manager but mostly deferring to the Manager. Sometimes it can do stuff which needs other Managers, though.

To me, the last option looks the most charming. What do you think?


ben-kaufman (commented):
- I think it would make sense to move also the old managers to the manages folder.
- Why leave all the methods of the ConfigManager also in the specter and not call directly to configmanager?

src/cryptoadvance/specter/managers/config_manager.py
```diff
@@ -0,0 +1,338 @@
...
+                explorer_data["url"] += "/"
+        else:
+            chain_name = (
+                ""
+                if (chain == "main" or chain == "regtest")
+                else ("signet/" if self.chain == "signet" else "testnet/")
+            )
+            explorer_data["url"] += chain_name
+        # update the urls in the app config
+        if user.id == "admin":
```
stepansnigirev, line 189:
```suggestion
        if user.is_admin:
```

src/cryptoadvance/specter/managers/config_manager.py
```diff
@@ -0,0 +1,338 @@
...
+    def update_fee_estimator(self, fee_estimator, custom_url, user):
+        """ update the fee estimator option and its url if custom """
+        if isinstance(user, str):
+            raise Exception("Please pass a real user, not a string-user")
+        fee_estimator_options = ["mempool", "bitcoin_core", "custom"]
+
+        if fee_estimator not in fee_estimator_options:
+            raise SpecterError("Invalid fee estimator option specified.")
+
+        if user.id == "admin":
```
stepansnigirev, line 205:
```suggestion
        if user.is_admin:
```

stepansnigirev (commented):
just for consistency, we should use user.is_admin everywhere instead of checking the user.id

src/cryptoadvance/specter/managers/config_manager.py
```diff
@@ -0,0 +1,338 @@
+import logging
+import os
+import random
+import secrets
+import threading
+import time
+from urllib.parse import urlparse
+
+from mock import Mock
```
stepansnigirev, line 9:
```suggestion
```

k9ert:
>     * I think it would make sense to move also the old managers to the manages folder.
> 
>     * Why leave all the methods of the ConfigManager also in the specter and not call directly to configmanager?

Both point pretty valid. Will do that as another PR.

PR #1109, opened 2021-04-11
[WIP] Restore hot wallets; some device type, descriptor refactoring

### LINKED ISSUE #948: Data Backup does not restore Bitcoin Core hot wallet.
Using a brand new Specter `1.0.0` (via getumbrel.com), I loaded the data backup files from my other node and it appeared to successfully restore all devices and wallet onto the node.

When attempting a simple roundtrip transaction using a 2-of-3, I discovered the following issue.

When signing with the newly restored hot wallet, the following error was surfaced:

```
Failed to sign PSBT: Request error: Requested wallet does not exist or is not loaded.
```

My best guess is that Specter restores the metadata, but does not check for hot wallets within, and therefor does not reinitialize those wallets inside Bitcoin Core. This causes the wallet to feel restored, but otherwords be orphaned.

kdmukai:
Fixes #948

This is a work-in-progress! The basic hot wallet restore code and its new test case work. But the UI integration is still buggy. But would like feedback on the backend changes and overall "watch only" device type approach.

* Explicitly creates a "watch only" device type to avoid confusion over "not-yet-hot restored 'hot' wallets".
* A (formerly) hot wallet that is read in from a wallet backup file will initially be configured as a "watch only" device.
* Adds a `convert_to_hot_wallet` flow for "watch only" devices.

Misc refactoring:
* Moves string-based device identifiers to a new `DeviceTypes` constants class (haven't updated all references yet but you see where it's headed).
* `wallet.py` account map now includes device types and labels to assist with restoring backed up wallets. (see discussion: https://github.com/cryptoadvance/specter-desktop/issues/948#issuecomment-792299094)
* Factored logic out of `server_endpoints/wallets.py` and into `utils/descriptor.py` so it can be reusable, specifically for the new test case (and is a good structural practice regardless). Note line 480 where including the device type pays off and we can treat the imported hot wallet as "watch only".
* `device_manager.py`: Adding trivial `get_by_name` so that external callers don't need to know the details of internal DeviceManager fields.

TODO:
* Need to check if an imported hot wallet is actually already hot in the node instead of assuming we have to treat it as watch only.
* `convert_to_hot_wallet` last step UI fix, confirm that end-to-end flow works (should already be proven via the test case though).
* See if there are device cleanup steps on disk left behind when the watch-only is converted to hot.
* Remove debugging statements.


moneymanolis:
Will move to an issue

PR #1110, opened 2021-04-12
Travis abuse detection

k9ert:
This adds a document on how to efficiently deal with Travis-CI abuse-detection.
I'll continue to manage that but i also wanted to have a proper place on how to do that which in the project.


k9ert:
Let's fuggedaboutit as since this morning we're no longer using travis-ci.

PR #1122, opened 2021-04-19
Improve tx-table and addresses-table render performance

willcosgrove:
I was having trouble getting the wallet overview page to load on my iPhone. The page would crash after trying to load a full list of 50 transactions. My desktop didn't seem to have a problem loading the list. It did take a little bit of time, but it wouldn't crash.  But if I changed the view to load all transactions I could get Chrome to give me a warning about the website being unresponsive.

I decided to profile it to see if I could tell what was going on. Eventually, I noticed this:

```js
self.tbody.innerHTML += txRow;
```

This line is causing a lot more computation than I think was expected. The TxRow element does a lot of DOM manipulation on construction, and setting the innerHTML like this will cause TxRow elements to get created and destroyed every time it gets set.

To fix this I switched to using the [ParentNode.append()](http://developer.mozilla.org/en-US/docs/Web/API/ParentNode/append) function. That function takes a Node to append, so I still needed to convert the DOMString into parsed nodes, so for that, I'm using [Range.createContextualFragment()](http://developer.mozilla.org/en-US/docs/Web/API/Range/createContextualFragment) which seems to be the current best practice for parsing strings of DOM.

And finally, since everyone likes numbers, here's my chrome profile of rendering a list of 605 transactions before this change:

CleanShot 2021-04-19 at 15 02 25@2x

_total blocking time: 29,930 ms_

And after this change:

CleanShot 2021-04-19 at 15 03 45@2x

_total blocking time: 1,050 ms_

Almost a 30x speedup in my case. And my phone no longer crashes when I look at my wallet overview 🎉 

---

While working on this I noticed the same pattern being used in the addresses-table, and so I made the same improvement there.

### DIFF
--- a/src/cryptoadvance/specter/templates/includes/addresses-table.html
+++ b/src/cryptoadvance/specter/templates/includes/addresses-table.html
@@ -454,7 +454,7 @@
 
                     for (let addr of addressesList) {
 
-                        let addrRow = `
+                        let addrRow = document.createRange().createContextualFragment(`
                         <address-row 
                             data-verify-qr="{{ supports_qr_code_verify != [] }}"
                             data-verify-hwi="{{ supports_hwi != [] and (supports_hwi_multisig_display_address != [] or not wallet.is_multisig) }}"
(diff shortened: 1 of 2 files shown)


k9ert:
Very cool! Thank you, that's a great contribution. Tested it and i guess i can even feel it with 50 txs.
LGTM, ready to merge.

k9ert (approved):

ben-kaufman:
Thanks, works really great!
Merging.

PR #1132, opened 2021-04-24
initial elements compilation

k9ert:
The goal of this PR is to build the foundation for including elements regtest to the testing setup. The current state is that elements gets compiled and cached and a very simple test checks whether it's booting up and available via RPC. 

I'm not yet sure at which point we should merge. Merging rather sooner than later is imho a good idea in order to provide the changes which i did (refactoring the bitcoind-controller to node_controller etc.) to other branches.


k9ert:
Current issue with cirrus:
https://cirrus-ci.com/task/6358475853791232?logs=cypress#L787
It's the `wallets-loading-done-refresh` issue.
Re-Running ...

PR #1140, opened 2021-04-30
Liquid class separation

stepansnigirev:
This is a very small PR with my idea of separation of the chain-specific wallet logic into different classes:
- `Wallet` class in `src/wallet.py` should only care about Bitcoin
- `LWallet` class in `src/liquid/wallet.py` should be derived from `Wallet` and be aware of liquid stuff - and code changes for liquid are there
- `WalletManager` can use different wallet classes depending on the `chain` argument in the constructor.

Also fixes a few things in `helpers.py` that are broken in liquid.

If you think this approach is reasonable I will continue with extra liquid-specific classes like `LAddressList` and `LTxList` classes for csv caching.

### DIFF
--- a/src/cryptoadvance/specter/helpers.py
+++ b/src/cryptoadvance/specter/helpers.py
@@ -11,8 +11,11 @@ import subprocess
 import sys
 from collections import OrderedDict
 from mnemonic import Mnemonic
-from hwilib.tx import CTransaction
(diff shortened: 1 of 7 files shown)


ben-kaufman (approved):
LGTM, just a few fixes building on it here https://github.com/stepansnigirev/specter-desktop/pull/1

PR #1141, opened 2021-05-01
Update hidapi to 0.10.1

leon-costa:
The currently locked version of hidapi (0.7.99.post21) is quite old and fails to install on some systems. The dependencies no longer require this old version ([HWI](https://github.com/bitcoin-core/HWI/commit/eeadbd6a680601cd7ca3d1e5519012fc36049aef), [bitbox02](https://github.com/digitalbitbox/bitbox02-firmware/commit/8473dabea915dfa0aea14b3e6dcd73da1be34175)), so can it be updated?

I only ran `pip-compile --generate-hashes --upgrade-package hidapi requirements.in`.


k9ert:
The CI is stuck due to:

> "User is too new to use Community clusters! Please check with support!"

I just did that and hope to get some test-results soon!

leon-costa:
HWI and bitbox2 have both already updated to a more recent hidapi (see the links in my initial comment). `pip-compile --upgrade-package` installed the most recent hidapi that is still compatible with the dependencies.

k9ert:
Yes, you're right. Other than us, they are (no longer) pinning to exact versions. As a result, their dependencies doesn't get reflected automatically in ours. Thanks, was not aware of this.

k9ert:
I've checked the hashes and upgraded myself, found no difference. Merging.

PR #1146, opened 2021-05-04
Node manager

ben-kaufman:
- [ ] Add node and node manager classes
- [ ] Allow adding, removing, and switching between nodes
- [ ] Fix internal node functionality
- [ ] Remove old Bitcoin Core settings screen

### DIFF
--- a/src/cryptoadvance/specter/managers/node_manager.py
+++ b/src/cryptoadvance/specter/managers/node_manager.py
@@ -1,0 +1,129 @@
+import os
+import logging
+
+from ..rpc import get_default_datadir
+from ..specter_error import SpecterError
+from ..persistence import write_node, delete_file
+from ..helpers import alias, load_jsons
+from ..node import Node
+
+logger = logging.getLogger(__name__)
+
+
+class NodeManager:
+    # chain is required to manage wallets when bitcoind is not running
+    def __init__(
+        self,
+        proxy_url="socks5h://localhost:9050",
+        only_tor=False,
+        active_node="default",
+        data_folder="",
+    ):
+        self.data_folder = data_folder
+        self._active_node = active_node
+        self.proxy_url = proxy_url
+        self.only_tor = only_tor
+        self.update(data_folder)
+
+    def update(self, data_folder=None):
+        if data_folder is not None:
+            self.data_folder = data_folder
+            if data_folder.startswith("~"):
+                data_folder = os.path.expanduser(data_folder)
+            # creating folders if they don't exist
+            if not os.path.isdir(data_folder):
+                os.mkdir(data_folder)
+        nodes = {}
+        nodes_files = load_jsons(self.data_folder, key="name")
+        for node_alias in nodes_files:
+            fullpath = os.path.join(self.data_folder, "%s.json" % node_alias)
+            nodes[nodes_files[node_alias]["name"]] = Node.from_json(
+                nodes_files[node_alias],
+                self,
+                default_alias=node_alias,
+                default_fullpath=fullpath,
+            )
+        if not nodes:
+            self.add_node(
+                name="Bitcoin Core",
+                autodetect=True,
+                datadir=get_default_datadir(),
+                user="",
+                password="",
+                port=8332,
+                host="localhost",
+                protocol="http",
+                external_node=True,
+                default_alias="default",
+            )
+        else:
+            self.nodes = nodes
+
+    @property
+    def active_node(self):
+        return self.get_by_alias(self._active_node)
+
+    @property
+    def nodes_names(self):
+        return sorted(self.nodes.keys())
+
+    def switch_node(self, node_alias):
+        # this will throw an error if the node doesn't exist
+        self._active_node = self.get_by_alias(node_alias).alias
+
+    def get_by_alias(self, alias):
+        for node_name in self.nodes:
+            if self.nodes[node_name] and self.nodes[node_name].alias == alias:
+                return self.nodes[node_name]
+        raise SpecterError("Node %s does not exist!" % alias)
+
+    def add_node(
+        self,
+        name,
+        autodetect,
+        datadir,
+        user,
+        password,
+        port,
+        host,
+        protocol,
+        external_node,
+        default_alias=None,
+    ):
+        if not default_alias:
+            node_alias = alias(name)
+        else:
+            node_alias = default_alias
+        fullpath = os.path.join(self.data_folder, "%s.json" % node_alias)
+        i = 2
+        while os.path.isfile(fullpath):
+            node_alias = alias("%s %d" % (name, i))
+            fullpath = os.path.join(self.data_folder, "%s.json" % node_alias)
+            i += 1
+
+        node = Node(
+            name,
+            node_alias,
+            autodetect,
+            datadir,
+            user,
+            password,
+            port,
+            host,
+            protocol,
+            external_node,
+            fullpath,
+            self,
+        )
+        write_node(node, fullpath)
+        self.update()  # reload files
+        return node
+
+    def delete_node(self, node, specter):
+        logger.info("Deleting {}".format(node.alias))
+        # Delete files
+        delete_file(node.fullpath)
+        del self.nodes[node.name]
+        if self._active_node == node.alias:
+            specter.update_active_node(next(iter(self.nodes.values())).alias)
+        self.update()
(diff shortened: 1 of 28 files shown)


k9ert:
Looks good to me and works pretty much as expected.
The Logging button "see bitcoind logs" for the internal node doesn't point to the new directory, though.

src/cryptoadvance/specter/managers/node_manager.py
```diff
@@ -0,0 +1,189 @@
...
+            password,
+            port,
+            host,
+            protocol,
+            external_node,
+            fullpath,
+            self,
+        )
+        write_node(node, fullpath)
+        self.update()  # reload files
```
k9ert, line 133: Can we have some logger.info here?!

src/cryptoadvance/specter/managers/node_manager.py
```diff
@@ -0,0 +1,189 @@
...
+            "http",
+            fullpath,
+            self,
+            self.bitcoind_path,
+            "mainnet",
+            self.internal_bitcoind_version,
+        )
+        write_node(node, fullpath)
+        self.update()  # reload files
+        return node
```
k9ert, line 170: Here as well. Always good to have some info in the logs if persistence data is changing.

src/cryptoadvance/specter/managers/node_manager.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def delete_node(self, node, specter):
+        logger.info("Deleting {}".format(node.alias))
+        # Delete files
+        delete_file(node.fullpath)
+        del self.nodes[node.name]
+        if self._active_node == node.alias:
+            specter.update_active_node(next(iter(self.nodes.values())).alias)
+        self.update()
+
```
k9ert, line 180: And here, please.

src/cryptoadvance/specter/node.py
```diff
@@ -0,0 +1,320 @@
...
+        self.password = password
+        self.port = port
+        self.host = host
+        self.protocol = protocol
+        self.external_node = external_node
+        self.fullpath = fullpath
+        self.manager = manager
+        self.proxy_url = manager.proxy_url
+        self.only_tor = manager.only_tor
+        self.rpc = None
```
k9ert, line 48: Delete this line?!

src/cryptoadvance/specter/node.py
```diff
@@ -0,0 +1,320 @@
...
+    detect_rpc_confs,
+    get_default_datadir,
+)
+from .persistence import write_node
+
+logger = logging.getLogger(__name__)
+
+
+class Node:
+    def __init__(
```
k9ert, line 19: Can we have some documentation here? Most stuff is pretty obvious but some not, e.g. what data-type is "external_node"? What's the difference between "name" and "alias". We have that distinction somewhere else as well but haven't documented it there either. Let's start with that here.

src/cryptoadvance/specter/server_endpoints/controller.py
```diff
@@ -74,9 +76,9 @@ def server_specter_error(se):
 def server_error(e):
     """Unspecific Exceptions get a 500 Error-Page"""
     # if rpc is not available
-    if app.specter.rpc is None or not app.specter.rpc.test_connection():
-        # make sure specter knows that rpc is not there
-        app.specter.check()
+    # if app.specter.rpc is None or not app.specter.rpc.test_connection():
```
k9ert, line 79: Should'nt that logic reestablished again checking with the current node?

src/cryptoadvance/specter/server_endpoints/nodes.py
```diff
@@ -0,0 +1,333 @@
...
+nodes_endpoint = Blueprint("nodes_endpoint", __name__)
+
+
+@nodes_endpoint.route(
+    "new_node/", defaults={"node_alias": None}, methods=["GET", "POST"]
+)
+@nodes_endpoint.route("node/<node_alias>/", methods=["GET", "POST"])
+@login_required
+def node_settings(node_alias):
+    err = None
```
k9ert, line 32: Unused variable.

src/cryptoadvance/specter/server_endpoints/nodes.py
```diff
@@ -0,0 +1,333 @@
...
+
+    if not current_user.is_admin:
+        flash("Only an admin is allowed to access this page.", "error")
+        return redirect("")
+    # The node might have been down but is now up again
+    # (and the checker did not realized yet) and the user clicked "Configure Node"
+    if node.rpc is None and node_alias:
+        node.update_rpc()
+
+    err = None
```
k9ert, line 71: Let's delete this line.

src/cryptoadvance/specter/server_endpoints/nodes.py
```diff
@@ -0,0 +1,333 @@
...
+
+    if not current_user.is_admin:
+        flash("Only an admin is allowed to access this page.", "error")
+        return redirect("")
+    # The node might have been down but is now up again
+    # (and the checker did not realized yet) and the user clicked "Configure Node"
+    if node.rpc is None:
+        node.update_rpc()
+
+    err = None
```
k9ert, line 237: Same here

src/cryptoadvance/specter/server_endpoints/nodes.py
```diff
@@ -0,0 +1,333 @@
...
+                flash(
+                    "Failed to deleted node. Specter must have at least one node configured",
+                    "error",
+                )
+        elif action == "stopbitcoind":
+            try:
+                node.stop()
+                time.sleep(5)
+                flash("Specter stopped Bitcoin Core successfully")
+            except Exception:
```
k9ert, line 275: Why do you catch Exception here? Can you either remove it or catch a more specific Exception?
If that's difficult for some reason, at least add a `logger.exception(e)`

src/cryptoadvance/specter/server_endpoints/nodes.py
```diff
@@ -0,0 +1,333 @@
...
+                try:
+                    flash("Stopping Bitcoin Core, this might take a few moments.")
+                    node.rpc.stop()
+                except Exception as e:
```
k9ert, line 279: Basically same arguments here.

src/cryptoadvance/specter/server_endpoints/nodes.py
```diff
@@ -0,0 +1,333 @@
...
+        specter=app.specter,
+        rand=rand,
+    )
+
+
+@nodes_endpoint.route("/internal_node_logs/<node_alias>/", methods=["GET"])
+@login_required
+def internal_node_logs(node_alias):
+    node = app.specter.node_manager.get_by_alias(node_alias)
+    logfile_location = os.path.join(node.datadir, "debug.log")
```
k9ert, line 319: Ahh, i guess this needs a change as it points to the old location.

k9ert:
There is imho one more question: How do we migrate the data on the user-side already using internal nodes? There are two possibilities:
* Do it automatically somehow. In that case we should try to get a more generic mechanism in place which executes specific funtions based on specific configurations (like an array of executed migration-functions).
* Or we simply don't do that and let the people recreate their internal node and this needs to be documented doemhow.

What do you think? I tend to the second solution. The first one is pretty huge and should be a PR of itself. And maybe we want to still defer that mechanism until we might change persistence-technology.

src/cryptoadvance/specter/server_endpoints/nodes.py
ben-kaufman: Not sure what you mean here? This now takes the node datadir argument, so it's the right location

ben-kaufman:
This should already work with migrating an existing internal node, I also tested and it worked well. The code responsible for this is at https://github.com/cryptoadvance/specter-desktop/pull/1146/files#diff-88c6a6cabbeff77aafe908b914ca9b08c2c0f5942b54b10b782fbc9dded8c6a6R569

src/cryptoadvance/specter/server_endpoints/nodes.py
k9ert: Hmmm, it didn't worked for me and i haven't checked this line good enough and simply assumed it's wrong. Doesn't look like this. Looks good, my fault!

k9ert:
There are known issues with the tor-test in cirrus, that's why this test is failing. Locally, all tests are green. That's why i'll merge this now nevertheless.

PR #1155, opened 2021-05-10
changed text from 'payed' to 'paid' in fee_selection file

chris-pollard:
'paid' should be used for the use of financial transactions, 'payed' used for nautical uses. see an example explanation https://english.stackexchange.com/questions/92547/paid-vs-payed


ben-kaufman:
Thanks!

PR #1165, opened 2021-05-17
Multiple internal nodes

ben-kaufman:
Allows setting up multiple internal nodes with different networks.
Currently, this will work only for one node per network, since the port will be already occupied, but the architecture already supports running multiple nodes of the same network on different ports, which is one of the 2 missing pieces we need in order to allow Quicksync option to have a background full sync (the second piece needed is way to migrate wallets between nodes).


k9ert:
LGTM but as long as we can't practically test that with e.g. testnet, signet or elements, this is imho not practical to merge.

ben-kaufman:
@k9ert added a test for this now

src/cryptoadvance/specter/managers/node_manager.py
```diff
@@ -137,6 +137,8 @@ def add_node(
     def add_internal_node(
         self,
         name,
+        network="mainnet",
```
stepansnigirev, line 140: Why not `"main"`? Looks like we have now main and mainnet that are basically the same.

ben-kaufman: Changed

stepansnigirev:
If I try to "forget the node" with built-in node I face certain issues:
- Specter doesn't stop it, so I have to kill it manually
- Next time I create a node they appear again but I can't edit them anymore

failed-to-delete

PR #1170, opened 2021-05-19
Hide sensitive info mode

ben-kaufman:
Add toolbar button to hide/ show "sensitive" information - like balances and data of txs and addresses.


relativisticelectron:
I do not understand the reason why this hiding is useful. What thread model does it protect against? 
- Sparrow or electrum don't hide amounts
- if you want to hide amounts, then log out or close the application

moneymanolis:
> * new users go panicky if suddenly nothing works any more and they don't know which button to press to make it visible again. I saw this happening, and it is not intuitive for users. This https://github.com/cryptoadvance/specter-desktop/blob/1374f56f9514892546c0b13724bbcb339df7ca02/src/cryptoadvance/specter/managers/config_manager.py#L68

Let me look into that. I, personally, would also not enable it by default - at least without any proper notification on what is going on.

stepansnigirev:
I agree, should be never by default.

relativisticelectron:
> I do not understand the reason why this hiding is useful. What threat model does it protect against?

@ben-kaufman : can you please explain?

PR #1173, opened 2021-05-20
Upgrade internal bitcoind version

ben-kaufman:
Update the packaged `bitcoind` version to `0.21.1` and add an upgrade mechanism for this and future updates


pyinstaller/specterd.spec
```diff
@@ -33,16 +33,16 @@ elif platform.system() == 'Linux':
                 break
     packaged_software_datas = [
         ('./torbrowser/tor-browser-linux64-10.0.15_en-US.tar.xz', 'torbrowser'),
-        ('./bitcoind/bitcoin-0.21.0-x86_64-linux-gnu.tar.gz', 'bitcoind'),
-        ('./bitcoind/bitcoin-0.21.0-arm-linux-gnueabihf.tar.gz', 'bitcoind'),
+        ('./bitcoind/bitcoin-0.21.1-x86_64-linux-gnu.tar.gz', 'bitcoind'),
+        ('./bitcoind/bitcoin-0.21.1-arm-linux-gnueabihf.tar.gz', 'bitcoind'),
```
stepansnigirev, line 37: I think we could package only one version of bitcoin core here depending on `platform.processor()`

ben-kaufman: But then we'll have to make 2 different Linux builds on 2 different Linux machines.

stepansnigirev: But we need to do it anyway. Currently we release only for x86_64, for arm we need another build machine

ben-kaufman: Right, my bad. I'll update it

ben-kaufman: Updated

src/cryptoadvance/specter/specter.py
```diff
@@ -685,7 +685,7 @@ def migrate_old_node_format(self):
                 self,
                 self.bitcoind_path,
                 "mainnet",
-                self.internal_bitcoind_version,
+                "0.20.1",
```
stepansnigirev, line 688: Should be 21.1, no?
```suggestion
                "0.21.1",
```

PR #1183, opened 2021-05-24
Broadcast txs over block explorer via tor

ben-kaufman:
Add option to broadcast via mempool.space (in addition to existing option to broadcast via blockstream), and add option to broadcast via the hidden service of either blockstream or mempool
Screen Shot 2021-05-24 at 11 51 46
Screen Shot 2021-05-24 at 11 51 55


k9ert (approved):
LGTM

PR #1188, opened 2021-05-26
Bugfix: Fix getting keys in wrong network format when no node is used

ben-kaufman:
Fix xpubs being converted to wrong SLIP132 format when no node is connected in Specter.

### DIFF
--- a/src/cryptoadvance/specter/hwi_rpc.py
+++ b/src/cryptoadvance/specter/hwi_rpc.py
@@ -235,22 +235,17 @@ class HWIBridge(JSONRPC):
             # Client will be configured for testnet if our Specter instance is
             #   currently connected to testnet. This will prevent us from
             #   getting mainnet xpubs unless we set is_testnet here:
-            if not chain or chain == "None":
-                try:
-                    client.chain = (
-                        Chain.TEST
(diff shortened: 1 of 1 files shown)


src/cryptoadvance/specter/hwi_rpc.py
```diff
@@ -235,19 +235,18 @@ def extract_xpub(
             # Client will be configured for testnet if our Specter instance is
             #   currently connected to testnet. This will prevent us from
             #   getting mainnet xpubs unless we set is_testnet here:
-            if not chain or chain == "None":
-                try:
-                    client.chain = (
-                        Chain.TEST
-                        if derivation.split("/")[2].startswith("1")
```
moneymanolis, line 242: Wouldn't it be better to be more specific here, setting the chain to main if derivation.split("/") startswith("0"), to test if startswith("1") and raising some error if sth. else (could also be by mistake) is provided?

moneymanolis:
Tested: 
(1) Using a regtest node, imported via USB and custom derivation a mainnet key, was successfully recognized as mainnet and not testnet key.
(2) Using a mainnet node, imported via USB and custom derivation a testnet key, was successfully recognized as testnet and not mainnet key.

PR #1193, opened 2021-05-27
Properly manage addresses caching using Core native descriptor wallets

ben-kaufman:
It fixes an issue where the address caching won't update so the wallet always had only 20 addresses showing.
The change should basically make sure that for wallets on Bitcoin Core 21 or higher, the addresses.csv caching is properly updated based on the needs of the keypool

### DIFF
--- a/src/cryptoadvance/specter/wallet.py
+++ b/src/cryptoadvance/specter/wallet.py
@@ -27,8 +27,6 @@ LISTTRANSACTIONS_BATCH_SIZE = 1000
 
 
 class Wallet:
-    # if the wallet is old we import 300 addresses
-    IMPORT_KEYPOOL = 300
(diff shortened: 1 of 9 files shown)


src/cryptoadvance/specter/wallet.py
```diff
@@ -1153,8 +1151,6 @@ def keypoolrefill(self, start, end=None, change=False):
         # Descriptor wallets were introduced in v0.21.0, but upgraded nodes may
         # still have legacy wallets. Use getwalletinfo to check the wallet type.
         # The "keypool" for descriptor wallets is automatically refilled
```
k9ert, line 1153: This comment is specific to the code which just got deleted, right? Please adjust the comment as well.

stepansnigirev (approved):

PR #1182, opened 2021-05-23
New sort function with regex for keys table addressing #1118

moneymanolis:
This replaces the sorting function of keys which was string based resulting in unintuitive orderings (account 12 came before account 2) as mentioned in #1118. The ordering is now: 1) chain/network 2) purpose 3) account. See screenshots for how it looks now:

grafik
grafik

PR #1195, opened 2021-05-27
Update to the new sort function for keys table

moneymanolis:
This is an update to #1182, changed ordering, now accounts before purpose, also made it Liquid-proof.


ben-kaufman:
Looks good, merging

PR #1200, opened 2021-05-29
Update Tor circuit every request

ben-kaufman:
Update Tor circuit for every new session used.

### DIFF
--- a/src/cryptoadvance/specter/specter.py
+++ b/src/cryptoadvance/specter/specter.py
@@ -568,8 +568,17 @@ class Specter:
     def requests_session(self, force_tor=False):
         requests_session = requests.Session()
         if self.only_tor or force_tor:
-            requests_session.proxies["http"] = self.proxy_url
-            requests_session.proxies["https"] = self.proxy_url
+            proxy_url = self.proxy_url
+            proxy_parsed_url = urlparse(self.proxy_url)
+            proxy_url = proxy_parsed_url._replace(
+                netloc="{}:{}@{}".format(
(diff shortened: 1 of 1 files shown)


src/cryptoadvance/specter/specter.py
```diff
@@ -568,8 +568,17 @@ def hide_sensitive_info(self):
     def requests_session(self, force_tor=False):
         requests_session = requests.Session()
         if self.only_tor or force_tor:
-            requests_session.proxies["http"] = self.proxy_url
-            requests_session.proxies["https"] = self.proxy_url
+            proxy_url = self.proxy_url
+            proxy_parsed_url = urlparse(self.proxy_url)
+            proxy_url = proxy_parsed_url._replace(
```
stepansnigirev, line 573: I think it's a bad idea to use private underscored method `_replace` outside of the class definition. Maybe there is a better way with public methods?

ben-kaufman: https://stackoverflow.com/questions/21628852/changing-hostname-in-a-url
It looks like in this case it's a common practice to use this `_replace` methond

ben-kaufman: Also https://docs.python.org/3/library/urllib.parse.html

k9ert: Yeah, strange but makes sense for a class which uses dynamic fields and don't want to conflict with any potential fieldname the user might come up with.

k9ert (approved):

PR #1202, opened 2021-05-29
Simplify emptiness checks in txlist

romanz:
Following https://docs.python.org/3/library/stdtypes.html#truth-value-testing


stepansnigirev:
Thank you!

PR #1204, opened 2021-05-30
return 0 if derivation is not like assumed

stepansnigirev:
Sorting fails if derivation path exists but doesn't match pattern.

For example, if Coldcard keys are imported including legacy `m/45h` the match will return `None` and break sorting.


k9ert:
image

k9ert:
Tests fail at tor. Will fix that soon. Merging nevertheless, Easy one!

moneymanolis:
That's good! This handles all sorts of special derivation paths / typos when using custom import.

PR #1212, opened 2021-06-02
Elements testing

k9ert:
Improving a lot of logging and hopefully fixed the flaky tests. Apart from that:
* configuring the elements-node
* creating a elements wallet
* funding that wallet
* Trying to create a TX as far as it gets (stopping with that unblinding issue)


src/cryptoadvance/specter/node.py
```diff
@@ -178,44 +189,57 @@ def update_rpc(
...
             self.protocol = protocol
             update_rpc = True
         if update_rpc:
             self.rpc = self.get_rpc()
-            write_node(self, self.fullpath)
+            if self.rpc and self.rpc.test_connection():
+                logger.info(f"persisting {self} in update_rpc")
+                write_node(self, self.fullpath)
+            else:
+                logger.error(f"not persisting broken {self.rpc} in update_rpc")
```
k9ert, line 228: I think this fixed the flaky test. `update_rpc()` is called way too often but the circumstances it'll change the persisted node was not specified good enough.

PR #1216, opened 2021-06-07
Liquid: asset support

stepansnigirev:
Adds asset support in Liquid wallets. Currently a random emoji is assigned to unknown asset (based on the asset hash).
Adds an `<asset-label>` custom element similar to `<address-label>` so users can edit the asset labels.
Asset labels are stored in the user config and can be changed from almost anywhere (except send screen).
Sending functionality is supported as well.

assets


stepansnigirev:
I think this PR is stable enough for review and merge.
Tests and other liquid-related fixes will come in other PRs - I don't want to blow up the size of this one too much.

ben-kaufman (approved):

PR #1225, opened 2021-06-11
Add SeedSigner device

stepansnigirev:
# Changes overview

Adds support for [SeedSigner](https://github.com/SeedSigner/seedsigner) device.

It is very similar to Generic device with a few exceptions:
- no SD card support
- QR codes frame rate is decreased to 2 frames per second for this device instead of default 3
- Added an icon. A bit too small but not sure how to deal with that.

Also this PR fixes a few minor bugs in QR code custom element, and adds support for the `frame-rate` and `animate` attributes for the QR code.
`frame-rate` is defined in frames per second, `animate` can set default animation mode - `auto / on / off`.
Also removes "Device PSBT - Full PSBT" tab bar in QR code signing - I don't think anybody ever used it.

# Minor

Removes redundant `__init__()` method that was basically just calling parent constructor.

# What it looks like

seedsigner


SeedSigner:
As I communicated privately, the workflow is very good, and I think the 2 frames per second is just right. The only other suggestion I would have is that b/c SeedSigner doesn't use static QRs, going directly to the animated QR sequence during the transaction signing process would save users the step of clicking to animate. Thank you again.

stepansnigirev:
> As I communicated privately, the workflow is very good, and I think the 2 frames per second is just right. The only other suggestion I would have is that b/c SeedSigner doesn't use static QRs, going directly to the animated QR sequence during the transaction signing process would save users the step of clicking to animate. Thank you again.

Updated. Now QR codes for SeedSigner are animated by default, and for electrum I disabled animation even for large QR codes.

src/cryptoadvance/specter/templates/includes/qr-code.html
```diff
@@ -323,20 +356,34 @@
...
+        }
+        // we have to use chunks
+        if(this.isQRtoolarge){
+          value = this.chunks[0];
+          this.animate = true;
+        }else{
+          // if this is bc-ur, remove checksum for static QR
+          value = this.maybe_remove_checksum(value);
+        }
+        if(this.mode == 'on'){
```
ben-kaufman, line 377: I guess this should also check that the QR gas more than 1 part?

stepansnigirev: Good catch! Fixed.

ben-kaufman (approved):
Looks good

PR #1226, opened 2021-06-13
Export device types and labels

### LINKED ISSUE #1176: Specter export does not contain hardware wallet type
While the export file (created from `Wallets > Settings > Export to Wallet Software` in Specter Desktop) contains sufficient information to reproduce a wallet as a watch only wallet, it does not contain information about the type of devices used to create it.

Practically this means that users need to reimport their hardware wallet after importing this file, since there is no way to tell what kind of interface to present to the user (for example when they want to sign a transaction). This is a common support request for Sparrow users.

As a suggestion, the JSON could contain an extra field which lists the devices in the order they appear in the descriptor:
```
{
  "label": "specter-multisig",
  "blockheight": 646346,
  "descriptor": "wsh(multi(3,[ffffffff/48'/0'/0'/2']xpub...,[eeeeeeee/48'/0'/0'/2']xpub...,[dddddddd/48'/0'/0'/2']xpub...,[cccccccc/48'/0'/0'/2']xpub...))",
  "devices": ["trezor", "ledger", "coldcard", "bitbox02"]
}
```

HWI already has a standard set of names which could be adopted.

kdmukai:
Fixes #1176

Makes the first half of #1109 redundant.

Screen Shot 2021-06-13 at 8 55 49 AM

Changes the Specter backup format to include more information about each signing device. As pictured above, even though each device is unknown to this Specter instance, the correct device types and names are applied in the import process. 

Import is still fully backwards compatible with previous Specter backups that do not have these new fields.

Changes:
* `helpers.py`: Pulls device types and labels from the new Specter backup file format as well as labels from Electrum backup files. Also fixes a bug when the Electrum file does not contain an "addresses" field. Changes wallet type determination to use the ~~derivation path~~ SLIP-132 xpub rather than address prefixes. Allows for detection of more wallet types.
* `wallets.py`: Moved functionality out of the endpoint and into `Descriptor.parse_signers` to make it easier to write test cases and just general better code cleanliness/separation of concerns. Escaping on the incoming wallet backup json was overly broad (e.g. the apostrophe in the wallet name was being escaped to "h"); now limited to just the "descriptor" field. Passes along any device types and names found in the backup file to the UI.
* `import_wallet.jinja`: Changes to use the device types and names when available.
* `wallet.py`: Changes `account_map` generation to use built-in json rather than relying on escaping strings. Adds the new "devices" field to the backup.
* `test_wallet_manager.py`: Two extensive new test cases to verify that single sig and multisig wallet backups can be restored with or without the new "devices" field.

Misc changes:
* Adds `DeviceTypes` (`device_types.py`) which is just a set of constants to avoid string-based device typing elsewhere in the code. Just a good habit to reference by constants instead of strings, especially as the list of device types keeps growing.
* `node_controller.py`: Adds additional time to the node timeout check if the node is still going through its initial launch steps. Only accounts for the two conditions I happened to hit ("Verifying blocks" and "Loading wallet") when running a node on my old MacBook Air. Likely that other messages from the node would signal that it's still working on spinning up. Also restored the `atexit` call with new exception handling to ensure that the internal node is properly shut down, especially when running the tests.
* `test_cli_bitcoind.py` and `test_node_controller.py`: Workaround fix to skip test when `elements` is not installed.
* `tests/utils.py`: Didn't mean to check this in, but contains an initial reusable function (that isn't being used yet) that will be useful for more complicated spending-related test cases.


src/cryptoadvance/specter/cli/cli_noded.py
```diff
@@ -264,7 +264,7 @@ def noded(
         )
         sys.exit(1)
     except Exception as e:
-        if e.startswith("There is already a node running!"):
+        if str(e).startswith("There is already a node running!"):
             echo(f"{e} please reset via:")
             echo(f"python3 -m cryptoadvacne.specter {node_impl}d --reset")
```
k9ert, line 269:
```suggestion
            echo(f"python3 -m cryptoadvance.specter {node_impl}d --reset")
```

tests/test_cli_bitcoind.py
```diff
@@ -27,12 +28,15 @@ def test_bitcoind(caplog):
...
 
 def test_elements(caplog):
-    # caplog.set_level(logging.DEBUG)
+    caplog.set_level(logging.DEBUG)
 
     runner = CliRunner()
     result = runner.invoke(elementsd, ["--no-mining", "--cleanuphard"])
     print(result.output)
     if result.exception != None:
+        if "Couldn't find executable elementsd" in str(result.exception):
```
k9ert, line 37: Well, that's a good point. In the mid-term, i think we should somehow create different labels for the tests and make it possible that people skip certain test-types. Did that e.g. here:
https://github.com/cryptoadvance/specter-desktop/blob/master/tests/test_wallet_manager.py#L13

Existing markers are defined in pytest-ini
https://github.com/cryptoadvance/specter-desktop/blob/master/pytest.ini#L5

You could then run the tests without the slow ones like this:
```
pytest -m "not slow"
```
Haven't figured out how devs could set their own default settings which would be a good thing to have, though.
What do you think about that approach?
And yes, i guess i need to do some better documentation about:
```
./tests/install_bitcoind.sh --elements compile
```

k9ert:
Great PR! Awesome work! Explicitly i like the detailed description of your changes and even more the added tests. Again, this is raising the bar for all of us! Thank you very much!

k9ert (approved):

src/cryptoadvance/specter/helpers.py
```diff
@@ -241,16 +241,32 @@ def generate_mnemonic(strength=256):
...
+    if simple_derivation.startswith("m/49"):
+        return "sh-wpkh"
+    elif simple_derivation.startswith("m/84"):
+        return "wpkh"
+    elif simple_derivation.startswith("m/48") and simple_derivation.endswith("/1"):
+        return "sh-wsh"
+    elif simple_derivation.startswith("m/48") and simple_derivation.endswith("/2"):
+        return "wsh"
+    else:
+        raise Exception(f"Unrecognized derivation: {derivation}")
```
stepansnigirev, line 255: This function assumes that any derivation that starts with `84h` or even non-hardened `84` is `wpkh`, and the same for all other script types. This is not true - bips define very specific length for the key derivations:
BIP84 must have at least 3 **hardened** indexes - `m / purpose' / coin_type' / account' `. The same applies for BIP49. Multisig `48h` derivations must have length of 4.
You can use [`embit.bip32.parse_path()`](https://github.com/diybitcoinhardware/embit/blob/master/src/embit/bip32.py#L259) function to convert the path to a list and then check the length, first index, second index (must be `0x80000000` or `0x80000001`), that third index is hardened ( `>=0x80000000` )

src/cryptoadvance/specter/server_endpoints/wallets.py
```diff
@@ -143,9 +143,17 @@ def new_wallet(wallet_type):
...
-                wallet_data = json.loads(
-                    request.form["wallet_data"].replace("\\'", "").replace("'", "h")
-                )
+                wallet_data = json.loads(request.form["wallet_data"])
+
+                # Only need to un-escape the descriptor, if there is one (Specter format
+                #   has it but Electrum backup does not.
+                if "descriptor" in wallet_data:
+                    wallet_data["descriptor"] = (
+                        wallet_data.get("descriptor")
```
stepansnigirev, line 152: what's the purpose of `get()` here? With the check above it should be safe to use `["descriptor"]`

stepansnigirev (changes requested):
I like the PR. Please have a look at my comment in `wallet_type_by_derivation` function. I think it needs more checks.

src/cryptoadvance/specter/helpers.py
```diff
@@ -260,29 +276,33 @@ def parse_wallet_data_import(wallet_data):
...
-            cosigners_types.append(d["hw_type"])
+            cosigners_types.append({"type": d["hw_type"], "label": d["label"]})
             i += 1
         xpubs = xpubs.rstrip(",")
-        if wallet_data["addresses"]["receiving"][0].startswith("bc") or wallet_data[
-            "addresses"
-        ]["receiving"][0].startswith("tb"):
-            wallet_type = "wsh"
+
+        if "derivation" in wallet_data["x1/"]:
```
stepansnigirev, line 283: I just checked with Electrum file format, looks like the most reliable way to detect script type is by looking at `xpub` field in the keys - they strictly use slip132:
- `wsh` uses `Zpub` (`Vpub` on testnet)
- `sh-wsh` uses `Ypub` (`Upub` on testnet)
- `sh` uses `xpub` (`tpub` on testnet).

For single-sig it's `zpub / vpub` for `wpkh`, `ypub / upub` for `sh-wpkh` and `xpub / tpub` for `pkh`.

Derivation is not a very reliable source of information as the seed generated on electrum itself has derivation `m/1'` (at least for `wsh`). It is also possible to import xpubs directly to Electrum without any derivation paths.

So I think the best way to properly parse electrum wallet type is to check prefix of the `xpub` field in the dictionary.

src/cryptoadvance/specter/util/descriptor.py
```diff
@@ -267,6 +268,21 @@ def parse(cls, desc, testnet=False):
...
+        if self.wpkh:
+            return "wpkh"
+        elif self.wsh:
+            return "wsh"
+        elif self.sh_wpkh:
+            return "sh-wpkh"
+        elif self.sh_wsh:
+            return "sh-wsh"
+        elif self.sh:
+            return "sh-wsh"
```
stepansnigirev, line 282: Looks like a typo
```suggestion
            return "sh"
```

src/cryptoadvance/specter/util/descriptor.py
```diff
@@ -415,6 +431,59 @@ def serialize(self):
 
         return AddChecksum(descriptor_open + base + descriptor_close)
 
+    def parse_signers(self, devices, cosigners_types):
+        from cryptoadvance.specter.key import Key
```
stepansnigirev, line 435: Why not `from ..key import Key`, and also why here and not at the top of the file?

kdmukai: @stepansnigirev Yeah, no reason for it not to be at the top of the file.

As for the dot import notation I find it great for sibling-level imports but I usually only use`..` when I'm working in something like Django where you're constantly drilling down into the identical directory structure copied over and over again for each child app. So you always know that, say, `..models` is still within that individual child app. 

I'm less of a fan when it reaches back to a more root level (e.g. `cryptoadvance.specter`). I'd rather be more explicit then. Rough analogy: If I'm inside an apartment, I don't have to specify my apartment number when I'm talking about my kitchen. But if I'm referring to something in the building's hallway or in someone else's apartment, then we should be explicit.

That being said, I can't really say I'm always super consistent one way or the other!

Given the above, since there was already another existing import with the full path notation, I left my import notation as-is but I did move it up.

src/cryptoadvance/specter/server_endpoints/wallets.py
kdmukai: It's mostly habit so that I don't ever try to retrieve something in an unsafe way that might not be there. But totally unnecessary here--and would fail anyway if the get() ever returned `None`.

Updating now.

src/cryptoadvance/specter/helpers.py
kdmukai: Awesome. That's much more straightforward. Update coming.

src/cryptoadvance/specter/util/descriptor.py
stepansnigirev: I understand your reasoning and agree with it.
Let's keep the import `from cryptoadvance.specter.key ...`, but move it up.

I like having all imports at the top of the file so it's easier to see the dependencies when moving / refactoring things, unless the dependency is used in a single function that is almost never called (for example during migration or in tests).

stepansnigirev:
I'm afraid you need to add a few more `DeviceTypes` to this PR as we just added SeedSigner and Jade ;)

src/cryptoadvance/specter/helpers.py
kdmukai: Now updated based on your comment further below.

tests/test_cli_bitcoind.py
kdmukai: The test suite runs in under 5 minutes on my old Macbook Air (that's with the two elements tests skipping themselves). So I'm not worried yet about getting finer control over the tests. I think the bigger issue is that we just need more tests! Stepan caught at least one bug in my code that should have ideally been caught by a test case.

stepansnigirev (approved):

src/cryptoadvance/specter/helpers.py
```diff
@@ -241,16 +241,38 @@ def generate_mnemonic(strength=256):
...
 
 
+def wallet_type_by_slip132_xpub(xpub):
+    """
+    see: https://github.com/satoshilabs/slips/blob/master/slip-0132.md
+    Electrum backups use SLIP-132 but note that other wallets don't make the same
+    guarantee.
+    """
+    if xpub.startswith("xpub") or xpub.startswith("tpub"):
+        return "sh"
```
stepansnigirev, line 251: It can be either `sh` or `pkh` depending if it's multisig or single-sig.
We don't support it, but I think this should work for that as well. Maybe we can return `legacy` and then handle it accordingly depending on the number of keys we detected below?

tests/test_cli_bitcoind.py
k9ert: Yeah, more tests are always better. 5 mins is to me already too long. Also it's the distribution. On my nuc:
* pytest -m "not slow" --> 81 tests in 10s
* pytest --> 89 tests in 3:17 mins

src/cryptoadvance/specter/helpers.py
kdmukai: @stepansnigirev Now updated with an `is_multisig` var to pass into that method.

stepansnigirev (approved):

ben-kaufman (approved):

PR #1234, opened 2021-06-15
Add Blockstream Jade support

ben-kaufman:
Add support for Blockstream Jade (single sig).
I had to make a few small changes/ hacks in the files I got from the HWI Jade PR to make this work, but tested with signet and everything works fine now for me.


src/cryptoadvance/specter/device.py
```diff
@@ -124,6 +124,8 @@ def key_types(self, network="main"):
 
     def has_key_types(self, wallet_type, network="main"):
         if wallet_type == "multisig":
+            if self.device_type == "jade":
```
stepansnigirev, line 127: I don't think it's a good idea to do jade- specific things in global device class.
You can overwrite `has_key_types` method in jade class.

stepansnigirev (changes requested):
Please move jade-specific code to jade device class.
Also don't we need to add udev rules for Linux?

https://github.com/JamieDriver/HWI/blob/blockstream_jade_support/hwilib/udev/55-usb-jade.rules

src/cryptoadvance/specter/device.py
ben-kaufman: Makes sense, moved it.

src/cryptoadvance/specter/devices/jade.py
```diff
@@ -0,0 +1,32 @@
...
+class Jade(HWIDevice):
+    device_type = "jade"
+    name = "Jade"
+    icon = "jade_icon.svg"
+
+    supports_hwi_toggle_passphrase = True
+    supports_hwi_multisig_display_address = True
+
+    def __init__(self, name, alias, keys, blinding_key, fullpath, manager):
+        HWIDevice.__init__(self, name, alias, keys, blinding_key, fullpath, manager)
```
stepansnigirev, line 14: If you are using exactly the same function as the parent class you can remove the whole `__init__` function declaration

src/cryptoadvance/specter/hwi_rpc.py
```diff
@@ -417,6 +418,9 @@ def _get_client(
                 if chain == "liquidv1":
                     chain = "main"
                 client.chain = Chain.argparse(chain)
+                if type(client) is JadeClient:
+                    if chain == "signet":
+                        client.chain = Chain.TEST
```
stepansnigirev, line 423: If Jade requires a replacement of signet to testnet, does it also apply to regtest or it works fine with regtest?

stepansnigirev (commented):
Thanks for the changes.
Is there a reason you deleted `udev/README.md` completely?

src/cryptoadvance/specter/hwi_rpc.py
ben-kaufman: No regtest works fine

ben-kaufman:
No it was by mistake, added it back now

stepansnigirev (approved):

Issue #1230, opened 2021-06-13
"Invalid hash of the non witness utxo" - When creating Coldcard PSBT transaction


ArdRay:
Hey folks,

I am receiving the following error when I am attempting to create a PSBT transaction based on Coldcard hardware wallet. I am running the latest 1.4.3 version, on MacOS 11.4.

<img width="729" alt="server_error" src="https://user-images.githubusercontent.com/4355130/121822227-3de13300-cc9e-11eb-998a-0381740db9ef.png">

Details:

```
Traceback (most recent call last):
  File "flask/app.py", line 1950, in full_dispatch_request
  File "flask/app.py", line 1936, in dispatch_request
  File "flask_login/utils.py", line 269, in decorated_view
  File "cryptoadvance/specter/server_endpoints/wallets.py", line 690, in send_new
  File "flask/templating.py", line 140, in render_template
  File "flask/templating.py", line 120, in _render
  File "jinja2/environment.py", line 1090, in render
  File "jinja2/environment.py", line 832, in handle_exception
  File "jinja2/_compat.py", line 28, in reraise
  File "/var/folders/fq/rfybkrcs4qs7xllgrh75lv1w0000gn/T/_MEIBW9Ohs/templates/wallet/send/sign/wallet_send_sign_psbt.jinja", line 2, in top-level template code
    {% set tab = 'send' %}
  File "/var/folders/fq/rfybkrcs4qs7xllgrh75lv1w0000gn/T/_MEIBW9Ohs/templates/wallet/components/wallet_tab.jinja", line 1, in top-level template code
    {% extends "base.jinja" %}
  File "/var/folders/fq/rfybkrcs4qs7xllgrh75lv1w0000gn/T/_MEIBW9Ohs/templates/base.jinja", line 62, in top-level template code
    {% block main %}
  File "/var/folders/fq/rfybkrcs4qs7xllgrh75lv1w0000gn/T/_MEIBW9Ohs/templates/wallet/components/wallet_tab.jinja", line 12, in block "main"
    {% block content %}
  File "/var/folders/fq/rfybkrcs4qs7xllgrh75lv1w0000gn/T/_MEIBW9Ohs/templates/wallet/send/sign/wallet_send_sign_psbt.jinja", line 10, in block "content"
    {{ sign_tx_method(wallet, device, psbt, specter.info.chain) }}
  File "jinja2/runtime.py", line 679, in _invoke
[...]

PR #1235, opened 2021-06-16
Fix `non_witness_utxo` hash mismatch problem

stepansnigirev:
I made prev-tx check optional in `embit` library, this will stop breaking Specter-Desktop.
Meanwhile I'll try to reproduce hash mismatch issue that people have and see what's wrong with the verification.

*edit*
So the issue was that we recently implemented `psbtv2` support in `embit`, and the API became slightly different that caused this side-effect in `wallet.fill_psbt()` method. I simplified the method there to utilize changes in psbt implementation, should be good now.


k9ert:
see #1230

k9ert (approved):

k9ert:
LGTM, tested and works (although i couldn't reproduce the issue on coldcard/HWI/regtest in the first place.

PR #1244, opened 2021-06-21
Chore: Wallet new refactoring

k9ert:
Created a `WalletImporter` which encapsulates all the logic needed to create a wallet from a json and refactored the wallet-endpoint new_wallet wallet accordingly.


ben-kaufman (approved):
LGTM

PR #1245, opened 2021-06-21
Redesign the Tor settings screen

ben-kaufman:
Redesign the Tor setup screen based on feedback from https://www.youtube.com/watch?v=Lq77qFFOBcE


k9ert:
Just tested this and the The test_tor action assumes that the type of the connection is via-specter. However if it's custom this will throw an exception:
```
logger.error(app.specter.tor_daemon.get_logs())
```
I had my difficulties with getting the logs anyway so maybe try/except it inside the get_logs-function and return a reasonable text is the most resilient fix for this?

Also, if the internal tor-daemon is running, you shouldn't be able to switch to custom as the user is assuming that if he clicks "Test Tor", he's testing the custom implementation. But if the custom-implementation is "accidentally" pointing to the internal tor-installation, the test is still "sucessfull".

The Tor-logs are only obtainable when the log-daemon is not running. So i'd recommend to also dump the logs by default when you're stopping the daemon.

Hope that makes sense.

src/cryptoadvance/specter/server_endpoints/settings.py
```diff
@@ -279,8 +283,9 @@ def tor():
                 tor_connectable = res.status_code == 200
                 if tor_connectable:
                     flash("Tor requests test completed successfully!", "info")
-                    logger.error("Tor-Logs:")
-                    logger.error(app.specter.tor_daemon.get_logs())
+                    if tor_type == "builtin":
+                        logger.error("Tor-Logs:")
+                        logger.error(app.specter.tor_daemon.get_logs())
```
k9ert, line 288: This won't work if tor is running properly. If it's not running properly, it's no wonder the test failed.
So maybe the best move here is stopping tor, getting the logs and start it again?!

k9ert:
Can you also put `logger.info("..")` commands in front of each action to make the logs more expressive?

k9ert (approved):

PR #1247, opened 2021-06-22
Babel integration

kdmukai:
Note: New pip requirement `pip install Flask-Babel` but I need info on how to generate a new `requirements.txt` that will include the hashes.

Overview: Babel will attempt to find a translation for a given wrapped string in the target language. If an entry is not found or is left blank, Babel will just return the English text. So even if there are large holes in the translation file for a target language, we won't have any broken functionality. The worst-case scenario is that the user gets English text, which is no worse than what they're seeing now.

See the new `babel/README.md` for most of the details.

Key changes:
* `config.py`: Defines the supported `LANGUAGES`.
* `server.py`: Defines a `SpecterFlask` class so we can add some locale-handling methods to the Flask app. Handles locale auto-detection. New endpoint to make explicit language selections. Initializes Babel to the app instance.
* `helpers.py`, `devices.py`: Multi-language support for BIP39 mnemonic wordlists. The library we're using already has all of the wordlists for other languages that are defined in BIP39. Wraps calls to `Mnemonic` so it can be initialized with the selected language.
* `babel/README.md`: docs to guide volunteer translators as well as explain to devs how to wrap strings in various situations and how to generate and compile the translation files.
* Tons and tons of string wrapping: Three main techniques used:
    ```
    from flask_babel import lazy_gettext as _
    before = "Hello, World!"
    after = _("Hello, World!")

    before = f"Hello, {user.name}!"
    after = _("Hello, {}!").format(user.name)

    # template code
    <div>Welcome!</div>
    <div>{{ _("Welcome!") }}</div>
    ```
    _Note: using `lazy_gettext` everywhere rather than deciding between it and `gettext`. There are some pitfalls when using `gettext` that are solved by `lazy_gettext` and I don't see a downside to just always using `lazy_gettext`._
* `language_select.jinja` and `language_js.jinja`: template includes to add the language selection droplist anywhere in the UI. Currently on the main landing page and in Settings.

Misc:
* `config.py`: Explicitly forcing `FLASK_ENV=development` when using the DevelopmentConfig
* `wallets.py`: Two routes were checking for POST when the route decorator could just be updated to limit to that method. `jsonify` used in place of manually constructed json responses to be more pythonic.

**This is still very much a draft as I'm rushing to get this PR out for everyone to take a look (i.e. not my ideal level of self-scrutiny). There are bound to be some sloppy text wrapping mistakes that still need to be caught.**


src/cryptoadvance/specter/config.py
```diff
@@ -112,6 +119,9 @@ class DevelopmentConfig(BaseConfig):
         os.getenv("SPECTER_DATA_FOLDER", "~/.specter_dev")
     )
 
+    # Env vars take priority over config settings so ensure that this is set
```
k9ert, line 122: Does this really work? I can loosely remember that i struggeled to define ENV-vars like this because it's already too late for that. But i might be wrong.

src/cryptoadvance/specter/server.py
```diff
@@ -26,6 +27,26 @@
 csrf = CSRFProtect()
 
 
+class SpecterFlask(Flask):
```
k9ert, line 30: That looks interesting. I think this is the very right thing to do. I18N is about the user-frontend. The specter-object is on a very different layer. I hope we'll be able to keep it that way. Might get challenging, though.
Great idea!

k9ert:
Amazing work, thank you so much, difficult to underapreciate this.

> Note: New pip requirement pip install Flask-Babel but I need info on how to generate 
> a new requirements.txt that will include the hashes.

it's [here](https://github.com/cryptoadvance/specter-desktop/blob/master/DEVELOPMENT.md#some-words-about-dependencies):
```
pip-compile --generate-hashes requirements.in
```
I've tested it and it seems to work even though that switching the language worked in the about-dialog but now in the settings-dialog.

kdmukai:
> it's [here](https://github.com/cryptoadvance/specter-desktop/blob/master/DEVELOPMENT.md#some-words-about-dependencies):
> 
> ```
> pip-compile --generate-hashes requirements.in
> ```

Got it! Just updated with new `requirements.in` and `requirements.txt`.

> I've tested it and it seems to work even though that switching the language worked in the about-dialog but now in the settings-dialog.

Ah, there was a dumb bug on Settings. That's now fixed. Also moved the language selection droplist on the About page so that it appears no matter which state the node is in.

src/cryptoadvance/specter/config.py
kdmukai: In my testing it didn't work if I just declared the var:
```
FLASK_ENV = "development"
```
But this `os.environ` approach has seemed to work for me.

kdmukai:
By the way, @k9ert @moritzwietersheim, notice that there's an empty Deutsch language file!

https://github.com/kdmukai/specter-desktop/blob/babel_integration/src/cryptoadvance/specter/translations/de/LC_MESSAGES/messages.po

You could try out the process of adding translations in Poedit (see the instructions for translators in the `babel/README.md`) and then compiling them into your local system (at the bottom of that README). Poedit isn't perfect nor is the process I laid out in the README. No matter how we do it, there's bound to be an unavoidable level of messiness.

The only other step you'd have to do is to add "de" to the `LANGUAGES` list in `config.py`; the UI droplist automatically populates from that list.

Another thought: I wonder if it might make sense for the language selection droplist to estimate a completion percentage for each language (e.g. "Español (traducciones completadas 35%)") to set expectations. Or maybe report it after a selection is made in a UI `flash` if the language still has a lot of missing strings.

babel/README.md
```diff
@@ -0,0 +1,139 @@
...
+### Create github account, install Github Desktop
+Install the [Github Desktop app](https://desktop.github.com/). This is an easy way to manage your changes to an open source project.
+
+When Github Desktop starts up it'll ask you to log in to your github account. If you don't already have an account, create one now.
+
+
+### Clone the Specter Desktop repo
+Now you're going to grab a copy of the full Specter Desktop source code, which includes the translation files.
+
+Open the Github Desktop app and add the Specter Desktop repo via "Clone a Repository". On the "URL" tab type in ~~"cryptoadvance/specter-desktop"~~ (**for now we have to point to my work-in-progress repo: "kdmukai/specter-desktop"**):
```
ben-kaufman, line 17: I guess just before we merge this PR this should be updated

ben-kaufman:
Seems like I get an error whenever trying to get a mnemonic in a language other than the English
Screen Shot 2021-06-24 at 12 17 43

kdmukai:
> Seems like I get an error whenever trying to get a mnemonic in a language other than the English

So it's the difference between `mnemonic==0.18` and `mnemonic==0.19`. Weird that at some point my local machine had 0.19 installed so it worked fine. Trying to update the `requirements.txt` now but running into some `pip-compile` complaints.

ben-kaufman:
Looks good with mnemonic 0.19

kdmukai:
Okay @ben-kaufman, I updated the `requirements.in` and `requirements.txt` to include `mnemonic==0.19`. Spanish and French mnemonics working again on my local system.

src/cryptoadvance/specter/templates/includes/language/language_select.jinja
```diff
@@ -0,0 +1,10 @@
+{#
+	Relies on `supported_languages=app.supported_languages` being injected into the context
+#}
+<select name="language_code" onchange="changeLanguage(this)">
+	{% for language_code, language in supported_languages.items() %}
+		<option value="{{ language_code }}" {% if session["language_code"] == language_code %}selected{% endif %}>{{ language }}</option>
+	{% endfor %}
+	<option value="de" disabled {% if session["language_code"] == "de" %}selected{% endif %}>Deutsch</option>
```
ben-kaufman, line 8: I guess these 2 should be removed since the languages are fetched autoomatically

babel/README.md
```diff
@@ -0,0 +1,139 @@
...
+
+Any newly wrapped text strings will be added to each `messages.po` file. Altered strings will be flagged as needing review to see if the existing translations can still be used.
+
+Once the next round of translations is complete, recompile the results:
+```
+pybabel compile -d src/cryptoadvance/specter/translations
+```
+
+
+### Adding support for another language
```
ben-kaufman, line 133: Here it should also mention you need to update the list of languages in the `config.py`

src/cryptoadvance/specter/templates/includes/language/language_select.jinja
kdmukai: Done!

babel/README.md
kdmukai: Good catch. Updated!

ben-kaufman (approved):
Looks great! @k9ert waiting for your review too before merging.

PR #1254, opened 2021-06-25
faq: more details about missing funds

k9ert:


docs/faq.md
```diff
@@ -413,8 +413,13 @@ Run it with the
...
+
 If you're running a pruned node, it's not possible to scan for the entire transaction history without doing a full re-download of the blockchain (IBD). Alternatively, we also support scanning for only the existing wallet balance (UTXO) which is very quick and supports both full and pruned nodes. However, for pruned nodes to support this feature, we must query some external data from an outside source (such as a block explorer, configurable by the user). This does not constitute a security risk, as the validity of the data can be verified against the hash existing on the pruned node itself, but can be a potential privacy risk, although it's possible to get the data over Tor to reduce the potential privacy leak. Yet, we still strongly recommend using a non-pruned-node (if possible) when dealing with older wallets.
 
+Also be aware that if you've use more than one "account" in your old wallet, you'd need to create the corresponding key in the device first and create a new wallet based on that account/key.
```
kdmukai, line 419: Suggested clarification:

...in your old wallet, you'll need to explicitly add the other account(s). Bitcoin accounts are actually specified at the device level. By default Specter imports the keys for just your first account. To import keys for additional accounts:
* click on your device to "Add more keys"
* click the "Edit" button at the top right
* click the "Add account" that will have now appeared
* Account 0 is the default that Specter imports on its own. Specify your target account number (accounts start at 0, so your second account is 1, your third account is 2, and so on). 
* click "Add account"
* retrieve keys from your device as usual (over USB, airgapped SD card, or QR code)

Now that the other account's keys have been imported, create a new wallet using the new key. After scanning this new wallet for funds, you should see your expected balance.

k9ert: Good point, added it.

PR #1275, opened 2021-07-06
Chinese (Simplified) translations

kdmukai:
Another @dudezoo commit that went to my fork that I'm moving over to merge into the main repo.

Already ran `update` and `compile`.


k9ert (approved):

k9ert:
Not sure which formal criteria's should go into these kind of PRs. I think the review are rather users which might complain about a specific language.

However, isn't their an entry in `config.py` missing?

PR #1276, opened 2021-07-06
Add RTL languages support

ben-kaufman:
Adding RTL support for languages like Hebrew


kdmukai:
Awesome! I was wondering how difficult this was going to be.

Also: I read "RTL" as [Ride-The-Lightning](https://github.com/Ride-The-Lightning/RTL).

ben-kaufman:
I probably missed quite a bit though. Still looking for more stuff that needs fixing.

src/cryptoadvance/specter/templates/base.jinja
```diff
@@ -10,6 +10,34 @@
 			img[src="{{ url_for('static', filename='img/devices/elementscore_icon.svg') }}"] {
 				transform: scale(2.2);
 			}
+			{% if session["language_code"] in config["RTL_LANGUAGES"] %}
```
k9ert, line 13: I expect that to be an idiom which might come up hundreds of times in the code:
```
session["language_code"] in config["RTL_LANGUAGES"]
```
Can we make that much shorter? something like `session["l10n_rtl"] `

stepansnigirev (approved):

PR #1282, opened 2021-07-08
improve logging

k9ert:
* We now have a `--tracerpc` switch which will dump each and every call to core/elements in the debug-logs like e.g.:
```
[  DEBUG] in             rpc: call(wallet/specter/mytrezor_2   ) payload:[{'method': 'listtransactions', 'params': ['*', 1000, 0, True], 'jsonrpc': '2.0', 'id': 0}]
[  DEBUG] in             rpc: call(wallet/specter/mytrezor_2   ) payload:[]
[  DEBUG] in             rpc: call(wallet/specter/mytrezor_2   ) payload:[{'method': 'getwalletinfo', 'params': [], 'jsonrpc': '2.0', 'id': 0}]
```
* We have a different logging-scheme for dev than for production (no timestamps and a bit more structure for dev)


src/cryptoadvance/specter/specter.py
```diff
@@ -138,7 +138,8 @@ def cleanup_on_exit(self, signum=0, frame=0):
 
         logger.info("Closing Specter after cleanup")
         # For some reason we need to explicitely exit here. Otherwise it will hang
-        exit(0)
+        # Doesn't seem to be the case anymore?!
```
stepansnigirev, line 141: Still hangs forever for me

stepansnigirev:
Looks good to me, except exit() thing above

src/cryptoadvance/specter/specter.py
ben-kaufman: For me as well

k9ert: ok, recreated it

stepansnigirev (approved):

PR #1283, opened 2021-07-08
Update messages.po

2a3dex:
Done with the 196 new strings.


kdmukai:
@ben-kaufman this one is now good to go.

PR #1290, opened 2021-07-09
Update messages.po

Pat4cryptoFR:
french translation again, 

to be continued, next days


kdmukai:
@ben-kaufman This is ready to go. I left out the compiled *.mo after a conversation with @k9ert. There's no need to keep committing changes to that binary file to the repo, which will cause unnecessary bloat. All of the *.mo files can just be generated as part of the `setup.py` and/or just before packaging the installers all at once with:

`pybabel compile -d src/cryptoadvance/specter/translations`

PR #1292, opened 2021-07-10
Add support for unconfidential addresses

stepansnigirev:
Fixes liquid-related issues when:
- sending to unconfidential addresses
- receiving to unconfidential addresses

Also:
- Adds a toggle for address in the receive tab to switch between confidential and unconfidential addresses.
- Implements `.get()` function for LAddressList that supports both confidential and unconfidential inputs (like `__getitem__`).
- Tests: Added `mine()` after sending initial balance to the wallet as currently new psbts in Liquid don't support unconfirmed outputs.


src/cryptoadvance/specter/liquid/addresslist.py
```diff
@@ -1,8 +1,24 @@
...
+        super().__init__(*args, **kwargs)
+        self._unconfidential = to_unconfidential(self.address)
+
+    @property
+    def unconfidential(self):
+        return self._unconfidential or self.address
+
+    @property
+    def is_confidential(self):
+        return self.address == self.unconfidential
```
ben-kaufman, line 16: Isn't this reversed? Seems like it should be `self.address != self.unconfidential`

stepansnigirev: You are right! Fixed!

ben-kaufman (approved):
LGTM

PR #1297, opened 2021-07-11
Update messages.po

KST-Energy:
Completed translation up to @bitpaint 's latest submission.


kdmukai:
@ben-kaufman The *.po has been fully updated against the latest in `master`. Should be ready to merge.

PR #1299, opened 2021-07-13
removing blobs and adjusting automations

k9ert:
The current practice of checking in the mo-files is creating repo-size bloat. This PR fixes this:
* Removes the `*.mo` files
* add them to `.gitignore`
* adds a hook in the `setup.py` to create the files with `python setup.py install`
* adjusting the setup procedure in `gitlab-ci.yml` and documentation


setup.py
```diff
@@ -1,11 +1,31 @@
...
+
+class InstallWithBabelCompile(install):
+    """from this stackoverflow question
+    https://stackoverflow.com/questions/40051076/compile-translation-files-when-calling-setup-py-install
+    """
+
+    def run(self):
+        from babel.messages.frontend import compile_catalog
+
+        print("TRANSLATIONS")
```
ben-kaufman, line 17: Maybe edit to something nicer or remove?

src/cryptoadvance/specter/translations/.gitignore
```diff
@@ -0,0 +1 @@
+./**/messages.mo
```
ben-kaufman, line 1: Why is this new file needed? Should just be `messages.mo` in the existing .gitignore I think

setup.py
```diff
@@ -28,6 +48,12 @@
     url="https://github.com/cryptoadvance/specter-desktop",
     packages=find_namespace_packages("src", include=["cryptoadvance.*"]),
     package_dir={"": "src"},
+    package_data={
+        "": [
+            "src/cryptoadvance/specter/translations/*/*.mo",
+            "src/cryptoadvance/specter/translations/*/*.po",
+        ]
+    },
```
kdmukai, line 56: When I run `setup.py install` it does the message compilation but when I look in the `build` dir the `translations` dir is missing. I tweaked `setup.py` to get it to include them:
```
    package_data={
        "": [
            "translations/*/LC_MESSAGES/messages.mo",
            "translations/*/LC_MESSAGES/messages.po",
        ]
    },
```
The `package_data` is relative to the `cryptoadvance.*` declared above.

That being said, I don't believe the *.po files are necessary for the build at this stage, right?

kdmukai (changes requested):

src/cryptoadvance/specter/translations/.gitignore
k9ert: I think it's a styling-question. Putting everything in one .gitignore is possible but maintaining gets a bit more difficult with bigger projects.
I moved it to the central one but kept a more specific path. I think that's a good idea in order to get a better idea why (where) certain files should be ignored.

setup.py
k9ert: Thank you, very essential point. I checked further and realised, that the files also don't end up in the pip/tar-package we're creating/uploading. The reason is that the package_data stuff is not integrated in a sdist, but only in a [bdist](https://stackoverflow.com/questions/7522250/how-to-include-package-data-with-setuptools-distutils) but we're currently creating sdist-binaries only. 
So the solution mentioned there was including it in the `MANIFEST.in` which i did as well.

I'm not 100% sure about the po-files. I think we should imho merge this PR and create a new pre-release testing this.

PR #1302, opened 2021-07-13
Refactor out ugly hack in components JS

ben-kaufman:
Replaces hackish `self` passing with normal `this` use.


src/cryptoadvance/specter/templates/includes/address-label.html
```diff
@@ -172,98 +172,96 @@
...
+            this.label.removeAttribute('contenteditable')
+            if (this.isTitle) {
+                this.label.style['font-size'] = '1.5em';
+            }
+            this.label.style.border = 'none';
+            this.edit.classList.remove('hidden');
+            this.cancel.classList.add('hidden');
+            this.update.classList.add('hidden');
+            this.isEditing = false;
+            enableExplorerLink(this);
```
stepansnigirev, line 186:
```suggestion
            this.enableExplorerLink();
```

src/cryptoadvance/specter/templates/includes/tx-table.html
```diff
@@ -471,227 +471,225 @@ <h1>{{ _("Export transactions to CSV") }}</h1>
...
-                }
-                break
-            default:
-                self.export.href = ``
-                self.emptyRow.children[1].innerText = '{{ _("Failed to fetch transactions...") }} ';
+        /**
+         * Fetches txlist from the Specter API and loads the result into TxRowElement
+        */
+        async fetchTxItems() {
+            self.callId++;
```
stepansnigirev, line 478: still self? and everywhere below as well.

stepansnigirev (changes requested):
Doesn't look complete yet.

Also instead of having hard-coded html in javascript strings consider having this html in `<template>`. Then syntax highlighting will help to spot errors in opening/closing tags.

As you work on that what do you think about creating a custom element `<editable-label>` that `<address-label>` and `<asset-label>` can inherit from?

src/cryptoadvance/specter/templates/includes/tx-table.html
ben-kaufman: Yes sorry I was in a hurry because I had to go and thought I could finish before leaving but clearly was a stupid idea

src/cryptoadvance/specter/templates/includes/address-label.html
```diff
@@ -99,7 +99,7 @@
                 this.label.innerText = this.labelValue;
             } else {
                 this.label.innerText = this.address;
-                fetchAddressLabel(this);
+                fetchAddressLabel();
```
stepansnigirev, line 102: Shouldn't it be `this.fetchAddressLabel()`? And everywhere below as well.

ben-kaufman:
Ok should be ready now. I think we can leave the template stuff and inheritance to a separate PR.

stepansnigirev (approved):

PR #1232, opened 2021-06-14
Feature: Api framework

k9ert:
This is a continuation of the PR #713 but moving the code to a non-master branch on my remote.
Will close the original PR. With this PR, Specter gets an API. The documentation is in the docs-folder.

The API is only activated if:
```
export SPECTER_API_ACTIVE=true
```
It's also active by default in Dev-Mode.

Apart from this new feature:
* The test-framework needed some refactoring. Better some bug-fixing. The biggest change is that the `bitcoin_regtest` fixture changed scope from `module` to `session`. This means that the regtest-node is only be spun up exactly once for the tests.
* Made an update of a bunch of dependecies
* One of the dependencies (`cryptography`) needed a newer OS-version but the used cypress-image was based on an old debian-image. So i've created a new base-image ` docker/cypress-base-ubuntu-focal/Dockerfile` which is ubuntu based and what is from now on the new base for ` docker/cypress-python/Dockerfile`
* The `user.password` is now a property which will automatically detect a plain-password and will do the necessary salted hashing
* The wallet-enpoint for creating PSBT's got a refactoring and now we have `psbt_creator.py` very similiar to the `wallet_imported.py`. There are people calling that a "command-pattern".
* The `wallet_importer.py` got some improvements as well
* The `wallet.py` has a change where we're rather throwing an error than returning 0-balances. We should have more courage to raise `SpecterErrors`
* I've changed the port-numbers of adhoc-bitcoind-nodes spun up in tests and i've wrapped them in try/raise to make sure that they will shutdown in any case of result of the tests (several tests).
* Finally there is also a small bugfix for this:
k9ert, [21.06.21 14:26]
```
[2021-06-18 14:36:39,975] INFO in wallet_manager: Wallet already in Specter, updating wallet: testwallet_ghost
[2021-06-18 14:36:40,107] ERROR in controller: Uncaught exception: 'dict object' has no attribute 'devices_signed'
[2021-06-18 14:36:40,108] ERROR in controller: Traceback (most recent call last):
  File "/tmp/cirrus-ci-build/.env/lib/python3.8/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/tmp/cirrus-ci-build/.env/lib/python3.8/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
[...]

PR #1303, opened 2021-07-14
Adds steps to install Rust compiler dependency

kdmukai:
The updated `cryptography==3.4.x` dependency requires a Rust compiler. Updating docs with steps to install Rust compiler. Windows instructions are minimal for now.


k9ert:
Good point. I've struggled already a bit with that in https://github.com/cryptoadvance/specter-desktop/pull/1232
I hope that this is an intermediate problem and imho also related to the python-version and the platform people are using.
If they provide a wheel-pakage for the combination, the compiler is not needed:
* https://pypi.org/project/cryptography/3.3.2/#files
* https://pypi.org/project/cryptography/3.4.7/#files

Right? In my PR mentioned above, as far as i can remember, i moved from debian:something to ubuntu:focal and avoided the compiler doing that.
Does it make sense so mention it?

kdmukai:
> If they provide a wheel-pakage for the combination, the compiler is not needed:

@k9ert Ah, great info! I've slightly reorganized and updated the doc to reflect this.

PR #1306, opened 2021-07-16
Release process

k9ert:
* This tries to implement a signing procedure directly on the buildagent which is failing unfortunately. Hint for the code: [here](https://stackoverflow.com/questions/58418626/how-to-prevent-gpg-from-choosing-fallback-secret-key-when-default-key-is-not)
* The modification of the env-var get rolled back (see comments below)
* `babel` needed to be added as a requirement for compiling the po-files
* The `pyinstaller/specterd.spec` file needed an adjustement for packaging that in the AppImage
* Similar to the templates/static folders, Babel need to know where to pick it us: `BABEL_TRANSLATION_DIRECTORIES` (app-config)


src/cryptoadvance/specter/config.py
```diff
@@ -134,9 +134,6 @@ class DevelopmentConfig(BaseConfig):
     # API active by default in dev-mode
     SPECTER_API_ACTIVE = _get_bool_env_var("SPECTER_API_ACTIVE", "True")
 
-    # Env vars take priority over config settings so ensure that this is set
-    os.environ["FLASK_ENV"] = "development"
```
k9ert, line 138: @kdmukai This command is not specific to The DevelopmentConfig as it is no Class-Member but a statement which is executed while evaluating the whole file.

kdmukai: I don't understand. Are you saying that this would return True:

```
import os
if True == False:
    os.environ["foo"] = "bar"

print("foo" in os.environ.keys())
```

If you execute that in its own test file, I only get `False`.

k9ert: Maybe this clarifies it better:

```
import os

class Dev:
    os.environ["env"] = "dev"

class Prod(Dev):
    os.environ["env"] = "prod"

my_class = Dev()

print("now i'm on dev, right? " + os.environ["env"])
```

This line made sure that you're not able to run in production. (In the example it's the other way around).

k9ert: Yes, `__init__()`  might do it, but as far as i can remember, that didn't worked either when i tried it some time ago.

PR #1309, opened 2021-07-17
Update messages.po

KST-Energy:
Grammar correction + verification for terms redundancy of final portion of the lines that previously had to be reattached. (line ~3400+)


k9ert:
Thank you very much :+1:

PR #1310, opened 2021-07-17
Docs: message signature

k9ert:


k9ert:
Couldn't get reviews. Added @moneymanolis and @moritzwietersheim

moneymanolis:
I would add that the functionality is currently only available via USB / HWI.

moneymanolis (changes requested):
I would add that the functionality is currently only available via USB / HWI and that it is only possible for Single Sig (at least if using https://www.verifybitcoinmessage.com/).

Issue #528, opened 2020-10-18
Label import from electrum


relativisticelectron:
Is there any way to import labels from electrum?  This would make a migration a lot easier.

stepansnigirev:
I think this is possible when we get to importing electrum wallet files.
Not sure when we will do it, but importing wallets from other software including electrum is definitely on the roadmap.

relativisticelectron:
I started experimenting with how to import it: https://github.com/relativisticelectron/specter-desktop/commit/997ed683921505df469b91269c57e15b2b022e14

relativisticelectron:
Done in https://github.com/cryptoadvance/specter-desktop/pull/1314

PR #1314, opened 2021-07-19
Electrum address label import

relativisticelectron:
This PR allows to import the electrum_labels.json 

- electrum_labels.json contain the address labels. These are updated in specter.  If the address does not exist yet it will be ignored.
- electrum_labels.json also contains txid labels. The matching output addresses for a txid is looked up and then all these addresses get the txid label.  

Screenshot_20210721_193521

Matching issue: https://github.com/cryptoadvance/specter-desktop/issues/528


relativisticelectron:
Todo:

- what happens when someone imports addresses that do not belong to the wallet? This could be a security issue.

src/cryptoadvance/specter/wallet.py
```diff
@@ -402,6 +402,39 @@ def fetch_transactions(self):
                 self._addresses.add(addresses, check_rpc=False)
                 self._addresses.add(change_addresses, check_rpc=False)
 
+    def import_electrum_label_export(self, electrum_label_export, check_rpc=False):
+        import json
```
ben-kaufman, line 406: Unnecessary import, `json` is already imported at the top of the file

relativisticelectron: Thanks. I removed it.

ben-kaufman:
Hi, thanks for the PR! Could you please provide files for example case to test that? I.e. a regtest/ testent Electrum wallet file with labels?

relativisticelectron:
> Hi, thanks for the PR! Could you please provide files for example case to test that? I.e. a regtest/ testent Electrum wallet file with labels?

Yes I can provide a testnet electrum wallet. How should I send it to you?

ben-kaufman:
You could send me on email at ben.kaufman10@gmail.com

src/cryptoadvance/specter/wallet.py
```diff
@@ -402,6 +402,37 @@ def fetch_transactions(self):
...
+        # write tx_label to address_label in labels
+        for txitem in self._transactions.values():
+            if txitem["txid"] not in labeled_addresses:
+                continue
+
+            address_list = (
+                [txitem["address"]]
+                if isinstance(txitem["address"], str)
+                else txitem["address"]
+            )
```
relativisticelectron, line 421: By the way: I noticed that `txitem["address"]` is a `str` if there is only 1 output address that belongs to the wallet. However if there are multiple addresses that belong to the wallet the type is `list`.  This seems not ideal.  Is there any plan to change that to a `list` always in the future?

relativisticelectron (commented):
See comment at code

relativisticelectron:
- [x] I improved the UI by adding drag&drop functionality (I need 1 button less).

k9ert:
Sorry for the broken test. This one is still sometimes a false positive. I'm working on fixing it.
However, the testfile could be used to also make an import via a python-test. Wouldn't that make sense?

relativisticelectron:
> Sorry for the broken test. This one is still sometimes a false positive. I'm working on fixing it.
> However, the testfile could be used to also make an import via a python-test. Wouldn't that make sense?

Yes. One could test:

- does the import work on 1 address
- Import an address+label that does not belong  to the wallet. This input should be disregarded. Test: Is the total amount of addresses in the wallet unchanged and  the new address not in the wallet?

However I am not sure when I will have time to write these tests.

k9ert:
I'm not saying that we won't merge this one without tests. I might even can help with creating the tests. What i'm saying is that it would be very beneficial if you can attach an example-file either somewhere in the code or simply here in the comments.
I guess it's called a test-vector:
https://en.wikipedia.org/wiki/Test_vector

relativisticelectron:
> I'm not saying that we won't merge this one without tests. I might even can help with creating the tests. What i'm saying is that it would be very beneficial if you can attach an example-file either somewhere in the code or simply here in the comments.
> I guess it's called a test-vector:
> https://en.wikipedia.org/wiki/Test_vector

The input file (or label export of electrum) could look like
```
{
    "8a5e405ff3a18ba7b2dabe6fd664714bc320f8cda3cf03e21079307124507711": "this is a tx label of the 2. tx of block 693434",
    "1a1d0b0baa91db578439e8a8eaeb30de4d449a5577ea7d284680e06b310149eb": "this is an address label of an address in the 3. tx of block 693434",
    "3EbCB6Bu6DVeabLrkiekwCqFKLLXyjKcrb": "this is an address label of an address in the 3. tx of block 693434"
}
```
(randomly chosen transactions)

If these tx and addresses belong to the specter wallet, the address labels associated with the addresses and txids should be updated.  If they do not belong to the wallet, they should be completely ignored.

k9ert:
This is a great explanation! So maybe we can put that as a small information-popup next to the import button just like here:
image

The json example can then be much shorter like e.g.:
`
{
  "8a5e4...": "some Label",
  "1a1d0...": "some other Label"
}
`
Does that make sense?

relativisticelectron:
> This is a great explanation! So maybe we can put that as a small information-popup next to the import button just like here:
> image
> 
> The json example can then be much shorter like e.g.:
> `{ "8a5e4...": "some Label", "1a1d0...": "some other Label" }`
> Does that make sense?

I think that's a good idea. I will add it soon.

relativisticelectron:
I added the tooltip.
image

k9ert:
Thank you very much :+1:

PR #1315, opened 2021-07-20
Fix specifying datadir for internal node

### LINKED ISSUE #1313: Install BTC Full Node Data on external Drive
Hello Everyone,

would it be possible to add a custom data path for BTC Core within the first run of Specter?

Right now you could choose between pruned and full node on the same machine within the
default data path - but what if you did not have much space on the internal drive and would
like to install das BTC Core Node Data on an external Drive connected to the same Machine?

Would make life so much easier - right now I m downloading Bitcoin Core - and after that I have
to handle JSON RPC Stuff with User and Password Stuff - and I couldn't find good guides for that
as well - so any unexperienced user would stop here and give up...

Many thanks and kind regards
Simon

ben-kaufman:
Fix #1313


src/cryptoadvance/specter/internal_node.py
```diff
@@ -76,6 +76,8 @@ def __init__(
                 self.datadir = os.path.join(self.datadir, "signet")
             logger.info(f"persisting {self} in __init__")
             write_node(self, self.fullpath)
+        if not os.path.isdir(os.path.expanduser(self.datadir)):
+            os.makedirs(os.path.expanduser(self.datadir))
```
k9ert, line 80: I just recently replaced this idiom in device_manager because of some strange race-condition where the error-message claimed that the directory already exists. Can you replace it with this:
```suggestion
        pathlib.Path(self.datadir).mkdir(parents=True, exist_ok=True)
```

src/cryptoadvance/specter/server_endpoints/setup.py
```diff
@@ -189,11 +183,8 @@ def setup_bitcoind_datadir():
         node = app.specter.node_manager.add_internal_node(
             node_name,
             network=network,
+            datadir=request.form.get("bitcoin_core_datadir", None),
```
k9ert, line 186: The idiom `request.form.get("bitcoin_core_datadir", None)` is used 4 times in that method. Can you replace it with a local variable initialized at the top?

k9ert:
As mentioned on the #1313 , the user does not seem to be aware that the folder gets deleted. Can we put a warning about that on the page?

ben-kaufman:
@k9ert we already have a warning if the directory already exists. Does it not show up for you?

k9ert:
Yes, it shows up! 
image
Sorry forgot about that. Not sure how the user messed up his existing directory but it doesn't seem to be related to specter.

Can we have some logging here as well? I think at minimumu we should have logging whenever we're doing something severe like deleting a directory or so.

PR #1342, opened 2021-08-12
Update wallet.account_map to contain more info using wallet.to_json

thehanimo:
* Updates data contained in the qr_code, json file and the pre-text to be the same as the data that gets downloaded on backup.
* Updates Backup PDF generated to also contain the same data.


thehanimo:
Closing in favour of #1345

PR #1343, opened 2021-08-12
[Devices]: Add Passport support

ben-kaufman:
Add support for Foundation's Passport device


src/cryptoadvance/specter/templates/includes/qr-scanner.html
```diff
@@ -104,7 +104,12 @@
         // expected: crypto-psbt
         let re3 = /UR:CRYPTO-(.*)/;
         let p3 = result.match(re3);
-        if(p3){
+
+        // check if it is a bcur2 Foundation Passport stuff
+        let re4 = /UR:BYTES\/(.*-*\/)/;
```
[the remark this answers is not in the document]
FoundationKen: See here for an example of a UR2 code with a single QR: https://github.com/BlockchainCommons/Research/blob/master/papers/bcr-2020-005-ur.md#ur-encoding

FoundationKen (commented):
Looks good.  We'll try it out.

src/cryptoadvance/specter/templates/includes/qr-scanner.html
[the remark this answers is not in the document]
stepansnigirev: We need to keep supporting bc-ur-1 because of the Cobo vault.
So this check should be expanded. I'd say if we see UR:BYTES we need to try bcur2 decoding first, and if it fails - treat it as bc-ur-1.

src/cryptoadvance/specter/templates/device/new_device/new_device_keys.jinja
```diff
@@ -459,6 +459,73 @@
                 if(result==null){
                     return;
                 }
+
+                // foundation's passport returns coldcard format encoded in bcur2
```
stepansnigirev, line 463: If passport uses the same format as coldcard it's better to create a function and reuse it instead of copy-pasting the same code.

src/cryptoadvance/specter/templates/includes/qr-scanner.html
```diff
@@ -104,16 +104,29 @@
...
+              result = this.bcur2decoder.resultUR();
+              this.progress.innerHTML = "";
+            } else {
+              this.progress.innerHTML = `Progress: ${Math.round(this.bcur2decoder.getProgress()*100)}%`;
+              return;
+            }
+            p1 = false;
+            p2 = false;
+          } catch {
+            p2 = true;
```
stepansnigirev, line 125: If you set `p2` to `true` instead of the regexp result how will the next block work (after `if (p1 || p2 )` where we check `m` and `n`?
Are you sure you've tested it with Cobo?

Here are a few test vectors, please make sure they are handled properly. If bcur1 data is scanned by the qr scanner it should be returned as is.

bcur1, single piece, short:
```
UR:BYTES/TQDXSETVD3HJQAM0WFKXGGRZD3SKSGRZD3SKSGRZD3SKSTUYY85
```
bcur1, single piece:
```
UR:BYTES/TYQ3WURNVF607QGQWYPQQQQQQYY7JNENCCVD6T97T2S66DUL9VGJP62M6QCKJ466KVV8NWTMY3FAYQGQQQQQPL0LLLLS9QYKNQQQQQQQQQTQQ98XDTL0LSUR3ECLPGNMQL3MQRK7DT5WZCRXF9WS2QQQQQQPVQQ50APH5G3XCYYGA2RKRM4R46JM3MALWSVYQQQQQQQQQYQ37Q8P75ZSQQQQQQTQQ9XRT72X24GWMRM8AQPG0J3V7ZJ8EXHZXCPZQCPEKD9AWRR0G38R3VGXQX56EQ0U9Y090HCM77RPZ40JJNGHUEKCVPSC8FEF9VZ5QQQGQQGQQZQQQQQQSQQQQQQQQYQQQQQQQQ3QYQ7P8Y4ZNDYWZL0XVN5LEF6U3PYARF0W55X4A7G6DG5MWH2JZR25UVVR5U5JKP2QQQYQQYQQPQQQQQQGQQGQQQQQQQQQQQQQGDUR6H
```
bcur1, two parts:
```
UR:BYTES/1OF2/2CDRWMGF36W66M59HPSE3K24GRMAJ9NR4G4UXJZCPHPX3JNWTJJSE336TQ/TYQ3WURNVF607QGQWYPQQQQQQYY7JNENCCVD6T97T2S66DUL9VGJP62M6QCKJ466KVV8NWTMY3FAYQGQQQQQPL0LLLLS9QYKNQQQQQQQQQTQQ98XDTL0LSUR3ECLPGNMQL3MQRK7DT5WZCRXF9WS2QQQQQQPVQQ50APH5G3XCYYGA2RKRM4R46JM3MALWSVYQQQQQQQQQYQ37Q8P75ZSQQQQQQTQQ9XRT72X2

UR:BYTES/2OF2/2CDRWMGF36W66M59HPSE3K24GRMAJ9NR4G4UXJZCPHPX3JNWTJJSE336TQ/4GWMRM8AQPG0J3V7ZJ8EXHZXCPZQCPEKD9AWRR0G38R3VGXQX56EQ0U9Y090HCM77RPZ40JJNGHUEKCVPSC8FEF9VZ5QQQGQQGQQZQQQQQQSQQQQQQQQYQQQQQQQQ3QYQ7P8Y4ZNDYWZL0XVN5LEF6U3PYARF0W55X4A7G6DG5MWH2JZR25UVVR5U5JKP2QQQYQQYQQPQQQQQQGQQGQQQQQQQQQQQQQGDUR6H
```

stepansnigirev (changes requested):
- code duplication
- bcur1 is not properly handled

src/cryptoadvance/specter/templates/includes/qr-scanner.html
```diff
@@ -104,16 +104,29 @@
...
-            this.progress.innerHTML = "";
-          }else{
-            this.progress.innerHTML = `Progress: ${Math.round(this.bcur2decoder.getProgress()*100)}%`;
-            return;
+
+        // check if it is a bcur2 Foundation Passport stuff
+        let re4 = /UR:BYTES\//;
+        let p4 = result.match(re4);
+
+        if (p3 || p4){
```
stepansnigirev, line 112: If you add one more check:
```js
if (p3 || (p4 && !p2) ){
```
you will not need to handle multi-piece bcur1 strings, so in the exception you could treat your result as final right away.

stepansnigirev (approved):
LGTM

PR #1345, opened 2021-08-13
Add Export Specter Format for wallets

thehanimo:
Adds a new "Export Specter Format" option in Wallet Settings > Export. This uses `wallet.to_json(for_export=True)` to generate a JSON file similar to the one that gets downloaded on backup from general settings.


ben-kaufman (approved):
Looks good, thanks!

PR #1346, opened 2021-08-13
More information while backing up specter data

singlatushar07:
While backing up the specter data, private keys of the hot wallets are
not backed up. This commit shows a warning to the user informing the
same.

### DIFF
--- a/src/cryptoadvance/specter/templates/settings/general_settings.jinja
+++ b/src/cryptoadvance/specter/templates/settings/general_settings.jinja
@@ -24,7 +24,7 @@
 			</div>
 			{{ _("Specter Data Backup") }}:
 			<div class="note">
-			    {{ _("Warning: This backup does not include private keys of the hot wallets") }}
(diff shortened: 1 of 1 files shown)


src/cryptoadvance/specter/templates/settings/general_settings.jinja
```diff
@@ -23,7 +23,9 @@
 				</p>
 			</div>
 			{{ _("Specter Data Backup") }}:
-			<br><br>
+			<div class="note">
+			    {{ _("Warning: This backup does not include private keys of the hot wallets") }}
```
k9ert, line 27:
```suggestion
			    {{ _("Warning: This backup does not include private seed of the hot wallets or, obviously, the private seed of your Hardwarewallets.") }}
```
It's pretty obvious but maybe worth mentioning nevertheless?

k9ert (approved):

src/cryptoadvance/specter/templates/settings/general_settings.jinja
singlatushar07: Yeah we may mention that, will be helpful to new users.

PR #1333, opened 2021-08-04
Update rpc None error

thehanimo:
The default Bitcoin Core node throws this error:
```
[2021-08-05 00:19:59,932] ERROR in node: connection of <Node name=Bitcoin Core fullpath=/Users/x/.specter_dev/nodes/default.json> is None in check_info
```
when Bitcoin Core is not installed.

This PR:
* Adds a `{datadir} not found` warning.
* Adds a `No rpc was found for {node}` warning.
* De-escalates the above error to a warning.
* Removes duplicate code in `src/cryptoadvance/specter/rpc.py`:
  * `get_configs` already calls `get_rpcconfig` when `config` is None.
  * `detect_rpc_confs` does not have to do this before calling `get_configs`.
* Fixes a typo (BitcoiCore-API -> BitcoinCore-API)

PR #1348, opened 2021-08-14
Use the configured bitcoind data directory (regression)

shesek:
Without this fix, the configured data directory is ignored and the default is always used instead.

This did not happen with v1.5.0, it appears that the regression was introduced [in 22cd3e93](https://github.com/cryptoadvance/specter-desktop/commit/22cd3e9372abd0f95ce679d02711fb2bde86e126#diff-04de6bab3ea93d6516d4769480f8ab06fa7e2eb339fc1c0dd302292fa820fd94L115) (#1333).

### DIFF
--- a/src/cryptoadvance/specter/rpc.py
+++ b/src/cryptoadvance/specter/rpc.py
@@ -113,7 +113,7 @@ def get_configs(config=None, datadir=get_default_datadir()):
 
 
 def detect_rpc_confs(config=None, datadir=get_default_datadir()):
-    rpcconfs = get_configs(config)
+    rpcconfs = get_configs(config, datadir)
     rpc_arr = []
     for conf in rpcconfs:
         rpc_arr.append(conf)


shesek:
BTW, on an unrelated note, it would be nice if the data directory was configurable using environment variables too and not only using the config file, so that it wouldn't be necessary to [fiddle with the config file](https://github.com/ez-org/eznode/blob/13a30717beac68d2d51130f2152501b35ade282f/specter/run#L22) to configure the data directory programmatically.

ben-kaufman:
Thanks! I think adding the data-dir also as parameter makes sense. Would you like to do that or you'd prefer that I will?

shesek:
> adding the data-dir also as parameter

What do you mean? Isn't that what 695515939367b2bf5d8b3ef645e266779e213e01 does?

ben-kaufman:
Yeah I meant the env variable. I'll check about it then, thanks!

PR #1349, opened 2021-08-17
docs: Suggest updating pip to resolve `cryptography` installation issues

shesek:
See https://cryptography.io/en/latest/installation/#building-cryptography-on-linux


k9ert:
Thank you, very good point. Took some time to qualify due to "august". Will merge soon.

PR #1350, opened 2021-08-17
Fix Liquid addresses UTXO count

ben-kaufman:
This fixes the UTXO counting and amount on each address on the addresses list tab.


src/cryptoadvance/specter/templates/includes/address-row.html
```diff
@@ -135,7 +141,13 @@
                 this.amount = parseInt(this.amount * 1e8);
             }
 
-            this.amountText.innerText = this.hideSensitiveInfo ? '########' : `${numberWithCommas(this.amount.toString())}`;
+            if (this.assetsCount > 1) {
+                this.amountText.innerText = this.hideSensitiveInfo ? '########' : `Multiple assets`;
+            } else if (this.assetsCount == 1) {
+                this.amountText.innerHTML = this.hideSensitiveInfo ? '########' : `${numberWithCommas(this.amount.toString())} <asset-label data-asset="${this.addressData.assets[0]}" data-label="${this.addressData.assetLabels[0]}"></asset-label>`;
```
stepansnigirev, line 147: It is not required to pass asset label to the `<asset-label>` component, if it's not provided the component will fetch it automatically.
So passing `asset_labels` to all functions can be avoided - we can reduce amount of code.
```suggestion
                this.amountText.innerHTML = this.hideSensitiveInfo ? '########' : `${numberWithCommas(this.amount.toString())} <asset-label data-asset="${this.addressData.assets[0]}"></asset-label>`;
```

src/cryptoadvance/specter/liquid/wallet.py
```diff
@@ -502,6 +503,53 @@ def adjust_fee(self, psbt, fee_rate):
             * (tx_full_size / psbt["tx"]["vsize"])
         )
 
+    def addresses_info(self, is_change, asset_labels=[]):
```
stepansnigirev, line 506: Passing asset_labels to all functions is not necessary, see comment below - `<asset-label>` component can fetch the label by itself.

stepansnigirev (changes requested):
This PR breaks address list in Bitcoin.
Here is a fix for your current implementation:
https://github.com/cryptoadvance/specter-desktop/pull/1351
Please next time check that you don't break Bitcoin functionality with your PRs.

Otherwise see comments, I think this PR can be simplified.

src/cryptoadvance/specter/liquid/wallet.py
```diff
@@ -5,6 +5,7 @@
 from embit.liquid.transaction import LTransaction
 from .txlist import LTxList
 from .addresslist import LAddressList
+from ..helpers import get_asset_label
```
stepansnigirev, line 8: this import is not needed anymore

src/cryptoadvance/specter/liquid/wallet.py
```diff
@@ -502,6 +503,49 @@ def adjust_fee(self, psbt, fee_rate):
...
+            addresses_info.append(
+                {
+                    "index": addr.index,
+                    "address": addr.address,
+                    "label": addr.label,
+                    "amount": addr_amount,
+                    "used": bool(addr.used),
+                    "utxo": addr_utxo,
+                    "type": "change" if is_change else "receive",
+                    "assets": addr_assets,
```
stepansnigirev, line 543: So the amount here in case of multiple assets is the sum of all assets?
This means there is no way to learn about the balances of different assets from this endpoint.
Maybe it would be better to change `assets` from a list to a dict with `{asset:amount}`?
Then we could potentially show more detailed info about the balance on the address in the future, it also could be used by other apps in the API calls.

src/cryptoadvance/specter/liquid/wallet.py
ben-kaufman: Removed

src/cryptoadvance/specter/liquid/wallet.py
ben-kaufman: Ok changed to this

stepansnigirev (approved):

PR #1353, opened 2021-08-20
Improved export and restore hot wallet process

### LINKED ISSUE #948: Data Backup does not restore Bitcoin Core hot wallet.

singlatushar07:
Fixes #948 

Added bitcoincore_watchonly device type and now while exporting
bitcoincore hot wallets, they will exported as watchonly wallets.

### DIFF
--- a/src/cryptoadvance/specter/devices/bitcoin_core.py
+++ b/src/cryptoadvance/specter/devices/bitcoin_core.py
@@ -256,3 +256,46 @@ class BitcoinCore(Device):
                         break
         except:
             pass  # We tried...
+
+
+class BitcoinCoreWatchOnly(BitcoinCore):
+    device_type = DeviceTypes.BITCOINCORE_WATCHONLY
+    name = "Bitcoin Core (watch only)"
+    icon = "bitcoincore_icon.svg"
+
+    hot_wallet = True
+
+    def sign_psbt(self, base64_psbt, wallet, file_password=None):
+        raise Exception("Cannot sign with a watch-only wallet. Convert")
+
+    def sign_raw_tx(self, raw_tx, wallet, file_password=None):
+        raise Exception("Cannot sign with a watch-only wallet")
+
+    def add_hot_wallet_keys(
+        self,
+        mnemonic,
+        passphrase,
+        paths,
+        file_password,
+        wallet_manager,
+        testnet,
+        keys_range=[0, 1000],
+        keys_purposes=[],
+    ):
+        # Convert the watch-only wallet to a hot wallet then fix up its internal attrs to
+        # match.
+        super().add_hot_wallet_keys(
+            mnemonic,
+            passphrase,
+            paths,
+            file_password,
+            wallet_manager,
+            testnet,
+            keys_range,
+            keys_purposes,
+        )
+
+        # Change type (also triggers write to file)
+        self.set_type(DeviceTypes.BITCOINCORE)
+        # After update this device will be available as a BitcoinCore (hot) instance
+        self.manager.update()
--- a/src/cryptoadvance/specter/devices/bitcoin_core.py
+++ b/src/cryptoadvance/specter/devices/bitcoin_core.py
@@ -261,9 +261,6 @@ class BitcoinCore(Device):
 class BitcoinCoreWatchOnly(BitcoinCore):
     device_type = DeviceTypes.BITCOINCORE_WATCHONLY
     name = "Bitcoin Core (watch only)"
-    icon = "bitcoincore_icon.svg"
-
-    hot_wallet = True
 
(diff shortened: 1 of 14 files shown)


src/cryptoadvance/specter/templates/includes/overlay/sign_tx_method.jinja
```diff
@@ -62,7 +62,7 @@
         {% endif %}
         {% if device.hot_wallet %}
             <button id="{{ device.alias }}_hot_sign_btn" class="btn flex-item" style="width: 190px; margin: 15px auto;">
-                {% if device.device_type == "bitcoincore" %}
+                {% if device.device_type == "bitcoincore" or device.device_type == "bitcoincore_watchonly" %}
```
ben-kaufman, line 65: Hmm this probably should be treated differently as a watch only can't sign

src/cryptoadvance/specter/server_endpoints/settings.py
```diff
@@ -117,6 +117,7 @@ def general():
                         continue
                 write_wallet(wallet)
                 app.specter.wallet_manager.update()
+                time.sleep(1)
```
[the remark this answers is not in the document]
singlatushar07: During restoration process, after selecting the files I was getting the error that the wallets linked with the hot wallet were not restored because they do not exist. So after seeing logs I concluded that before wallet manager could update all the wallets, following line was being executed and giving the error. So to avoid that this sleep is needed.
`wallet_obj = app.specter.wallet_manager.get_by_alias(wallet["alias"])`

src/cryptoadvance/specter/templates/includes/overlay/sign_tx_method.jinja
singlatushar07: right, maybe i think we should not render the button for signing the transaction in this case.

src/cryptoadvance/specter/devices/bitcoin_core.py
```diff
@@ -256,3 +256,46 @@ def delete(
...
         except:
             pass  # We tried...
+
+
+class BitcoinCoreWatchOnly(BitcoinCore):
+    device_type = DeviceTypes.BITCOINCORE_WATCHONLY
+    name = "Bitcoin Core (watch only)"
+    icon = "bitcoincore_icon.svg"
+
+    hot_wallet = True
```
ben-kaufman, line 266: Unnecessary since it inherits BitcoinCore which has it already

src/cryptoadvance/specter/server_endpoints/settings.py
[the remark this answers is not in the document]
ben-kaufman: I think a better approach here would be to not use the threading in `wallet_manager.update()` you could add to that update funtion a parameter like `threading=True`, then in line 105 of the wallet_manager do:
```
if self.allow_threading and threading:
```

Or something like that...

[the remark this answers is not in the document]
singlatushar07: Updated in latest commit

src/cryptoadvance/specter/devices/bitcoin_core.py
singlatushar07: Removed in latest commit

k9ert:
Can we have a bit more verbose documentation on this feature? Comments in the code are very rare. Maybe you can do one or more of these things:
* get a python-class documentation for BitcoinCoreWatchOnly-class
* For the creating backup procedure, there is a blue (i) icon which has a documentation to show if you hover. Can we put a sentence or two in there?
* Can we get an explanation how you can transform a BitcoinCoreWatchOnly wallet to a normal Hotwallet again? The button-description seem to make that possible but i struggle to understand Or is that even possible? I guess it's not, so that would be a recreation-process?
* Added kdmukai to the reviewers due to: https://github.com/cryptoadvance/specter-desktop/issues/948#issuecomment-791479513
* Also for some reason this doesn't seem to work for me. It behaved like it wasn't fixed: The wallet was there, transactions were shown but when i tried to sign a PSBT (which shouldn't be possible) i get a `Failed to sign PSBT: Request error: Requested wallet does not exist or is not loaded` (tried with a regtest)

PR #1355, opened 2021-08-20
Fix add keys

singlatushar07:
The mnemonic value entered in the textbox while adding keys to existing
hot wallet was not used to add keys instead some random mnemonic was
used. This commit fixes this behavior.


stepansnigirev:
Tested, works!

stepansnigirev (approved):

PR #1356, opened 2021-08-20
Warn user when hidden advanced fields are not empty

ben-kaufman:
Some browsers might automatically fill up the password field even if the field is hidden from the user. This PR adds a warning and informs the user if hidden fields are used and passed.

### DIFF
--- a/src/cryptoadvance/specter/templates/device/new_device/new_device_mnemonic.jinja
+++ b/src/cryptoadvance/specter/templates/device/new_device/new_device_mnemonic.jinja
@@ -32,7 +32,7 @@
             visibility: hidden;
         }
     </style>
-    <form id="form" action="{{ url_for('devices_endpoint.new_device_mnemonic', device_type=device_type) }}" method="POST" style="width: 100%;" onsubmit="showPacman();">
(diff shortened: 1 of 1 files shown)


src/cryptoadvance/specter/templates/device/new_device/new_device_mnemonic.jinja
```diff
@@ -107,12 +107,31 @@
...
+            if ((!hiddenAdvancedConfirmed && advancedSettings.style.display !== 'block') &&
+                (
+                    document.getElementById('file_password').value ||
+                    document.getElementById('passphrase').value ||
+                    document.getElementById('range_start').value != 0 ||
+                    document.getElementById('range_end').value != 1000
+                )
+            ) {
+                toggleAdvanced();
+                showError("You have set values in the Advanced section but eft it hidden. PLease verify they are correct or delete them, then click again to submit.")
```
jeffthibault, line 128: There are a couple typos here.
```suggestion
                showError("You have set values in the Advanced section but left it hidden. Please verify they are correct or delete them, then click again to submit.")
```

stepansnigirev (approved):

PR #1359, opened 2021-08-23
Feature: Uncle jim paperwallet

k9ert:
This will add another way to download a PDF in the settings-menu of a single-sig-wallet.
From the tooltip:
```
This PDF is optimized for singlesig Paperwallets: Wallets created by uncle Jim (<a href="https://stephanlivera.com/episode/265/">inspired by</a>) for family&friends which want a coldstorage but can\'t afford a Hardwarewallet yet.
So we enable the import in a specter-desktop later and list many single addresses (#0 is dedicated for a penny-test) for later use when the HWW is gone.
A checkbox can be used to verify the address on the (probably Specter-DIY or seedsigner) hardwarewallet as long as it\'s around and still hold that seed in memory.
It's assumed to be folded into DIN-A5 to fit in a transparent sealing-bag (together with the seed).
```


k9ert:
PDF looks like this:
image

k9ert:
Maybe someone can improve the UX. I'd have added the "i" to the button but the tooltip contains a link. If you click the link, you're clicking the button at the same time :-(.
image

PR #1360, opened 2021-08-23
Add assets registry for Liquid wallets

ben-kaufman:
Add a registry with the full known asset list to the main screen of the wallet. User can add or edit assets in the registry.


stepansnigirev (changes requested):
A few things I noticed:
- Adding the asset that is already in the list should not be possible
- There is no way to delete assets from the registry
-  Adding asset that is not 32-byte hex (64 hex characters) should not be possible

stepansnigirev:
Can you center it?

image

PR #1375, opened 2021-08-31
Bugfix: CI failure because of missing wget dependency

k9ert:
For yet unknown reasons, the cache missed for `bitcoind_installation_cache` (`cirrus.yml`) and for yet again unknown reasons, the wget command was not available in registry.gitlab.com/cryptoadvance/specter-desktop/cirrus-focal:latest


k9ert:
Merging with admin-privilidges as the CI-Infra is broken anyway but at least less broken with this commit.

PR #1379, opened 2021-09-01
Bugfix: rollback embit to 0.4.5 due to failing cypress-tests

k9ert:


k9ert:
Merging as this fix will make CI green again on master.

PR #1395, opened 2021-09-12
Added "Choose File" button inside the drag & drop texarea

relativisticelectron:
- Added "Choose File" button inside the drag & drop textarea
image

- On hovering it gets full opacity
image

- dragging a file onto the button itself will also work
image

This textarea now supports: 
1. drag&drop 
2. manual copy and paste 
3. file picker

If the feedback and usability is good, this combination could be rolled out to all textareas that can get filled by a file-picker.


stepansnigirev:
Would it be possible to make it a web component so we can reuse it for PSBT import and other fields where we use "Select file"?

stepansnigirev:
Re UX, the button feels too big and is kinda on the way.
Maybe we could make something similar to GitHub text fields? See screenshot:
image


Re web component, sorry, I just realized that we already had drag-and-drop functionality there before your PR...
I should probably take some time and move it to a web component myself.

Just FYI, there is no need in libraries to write a web component, and they are pretty damn powerful. I wish we used them more...
If you are interested you can have a look at the components we have already. These two are pretty decent:
- [qr scanner](https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/templates/includes/qr-scanner.html)
- [qr code](https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/templates/includes/qr-code.html)

k9ert:
@moneymanolis  Please have a look at this. Maybe we can do the change recommended by Stepan.

k9ert:
We had a discussion whether the button inside the textarea is ugly like hell or state of the art best practise of UIUX.
@GBKS do you have a stance on this?

GBKS:
It's tempting to bundle a lot of functionality into a single UI element, but sometimes it's easier for users when things are clearly spelled out - especially for interactions they rarely do (compared to something like a chat box people use 20 times a day for years). Here's an idea for giving the user two clear top-level options (upload or paste) that they have to choose one from, and then offering the drag & drop as a sub-option of the upload option.

I also tried to make a visual distinction between the primary option (Import) and secondary UI buttons (Choose file), as I first had the impression that those two are equivalent options.

Do you maybe know what option people primarily use? Another idea could be to make the "Choose file" the big default and make the manual entry accessible via a small toggle. That could also provide a clear focus.

Specter address label import 211111

relativisticelectron:
Sorry for the very, very late reply.   I put the PR on draft mode because it is not clear to me what is best (and my time to work on it was and is very limited). Ideally the solution should be a web component that could be reused in:

- https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/templates/wallet/send/import/wallet_importpsbt.jinja   (TX import)
- https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/templates/wallet/settings/wallet_settings.jinja  (address label import)

> Re UX, the button feels too big and is kinda on the way. Maybe we could make something similar to GitHub text fields - a full-width footer in the textfield that one can click to select file? See screenshot: image
> 
> Re web component, sorry, I just realized that we already had drag-and-drop functionality there before your PR... I should probably take some time and move it to a web component myself.

Which existing UI element with drag&drop functionality are you referring to?


> Just FYI, there is no need in libraries to write a web component, and they are pretty damn powerful. I wish we used them more... If you are interested you can have a look at the components we have already. These two are pretty decent:
> 
>     * [qr scanner](https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/templates/includes/qr-scanner.html)
> 
>     * [qr code](https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/templates/includes/qr-code.html)

Thanks for the pointers to the qr-scanner. This gives a great template!


> Specter address label import 211111

Thanks for the great Mockup.    I am confused about the radio buttons. They would introduce extra clicks for the user.     Additionally introduced https://github.com/cryptoadvance/specter-desktop/pull/1470 more import options for different file formats which complicates the UI more.  It looks currently like:

image

moneymanolis:
@relativisticelectron I have already worked on a file uploader web component, once I have finished it, I will tag you, so you can have a look. I might just make a new PR then and mention this PR - probably cleaner.

relativisticelectron:
> 

That sounds great. Thank you!.  Then I close this PR.

PR #1405, opened 2021-09-20
Chore: Macos build

k9ert:
* Some documentations and minor improvements on how to build/release the MacOSX binaries.
* creating logs for the App and specterd and a new menu entry to gather these logs from within the App
* Changed the criteria for a successfull starup of specterd from instead having `Serving Flask app "cryptoadvance.specter.server` in the logs, seeing this: `  * Running on http`
* specterd is now the very same than the start of a regular specter. Internally it's called with `--no-logfile` (as the logs are streamed to stdout/-err via winston towards `specterApp.log`) and special `SPECTER_LOGFORMAT` 
* Also `LC_ALL` and `LANG` are hadrcodedly set to `en_US.utf-8` in order to fix https://github.com/cryptoadvance/specter-desktop/issues/1406 (which covered this underlying issue: https://stackoverflow.com/questions/36651680/click-will-abort-further-execution-because-python-3-was-configured-to-use-ascii)

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -52,16 +52,19 @@ test_task:
     - apt-get update && apt-get install -y --no-install-recommends python3-dev python3-pip wget
   << : *PREP_STUFF_TEMPLATE
   test_script: 
-    - pwd
-    - ls -l ./.env/bin
     - source ./.env/bin/activate
     - echo $PATH
     #- pip3 install -e .
-    - pytest --cov=cryptoadvance
+    - pytest --cov=cryptoadvance --junitxml=./testresults.xml
+  always:
+    junit_artifacts:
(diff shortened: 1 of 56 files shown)


k9ert:
I've also increased the resources provided by Cirrus and i should have done that a long time ago. Probably i could now reduce the timeout-values again as increasing resources was so helpful.
Here is a table showing time it takes for different configurations:
| CPU | MEM (GB) | TIME  |
|-----|-----|-------|
| 2   | 4   | 24min |
| 4   | 8   | 13min |
| 8   | 8   | 9min  |

8CPU/32GB is the max for the free plan. Let's stick to 8/8 in order to be mindful about resources.

stepansnigirev (approved):

PR #1408, opened 2021-09-27
Bitcoin Core as default for fee estimation, error handling improvements

k9ert:
With this PR:
* The default Fee-Estimation Service is no longer mempool.space but Bitcoin Core
* Also the Error-handling got improved. We have now a timeout and if timeouts occur, you get proper flash-messages in the UI
* Timeout is specified in FEEESTIMATION_REQUEST_TIMEOUT which can also be set via Envvar.


src/cryptoadvance/specter/util/fee_estimation.py
```diff
@@ -1,40 +1,115 @@
...
+                "minimumFee": 1,
+            }
+        )
+        fee_estimation.error_message = f"Failed to Fetch fee estimation. Please use manual fee calculation, check the logs for details. Error: {e}"
+        logger.exception(e)
+        return fee_estimation
+
+
+def _get_fees(specter, config):
+    fee_estimation_result = FeeEstimationResult(None)  # Just in case we have errors
```
stepansnigirev, line 55: Maybe better to pass here some defaults instead of `None`, like:
```
{
                "fastestFee": 1,
                "halfHourFee": 1,
                "hourFee": 1,
                "minimumFee": 1,
}
```

Otherwise there is no guarantee that the result will have `["hourFee"]` or other fees, that will cause exceptions at higher level.

stepansnigirev: Ah, nevermind, I see you catch exception above anyway,

src/cryptoadvance/specter/util/fee_estimation.py
```diff
@@ -69,5 +146,13 @@ def get_fees(specter, config):
...
             (float(specter.estimatesmartfee(20).get("feerate", 0.00001)) / 1000) * 1e8
         ),
-        "failed": "feerate" not in specter.estimatesmartfee(1),
     }
+    if "feerate" not in specter.estimatesmartfee(1):
+        # regtest does not seem to have a reasonable result for estimatesmartfee.
+        # The error-message is covering "transactions" and so the cypress-tests are not succeeding
+        # Not a perfect fix but better than to fix it everywhere in the test-code:
+        if specter.chain != "regtest":
+            fee_estimation_result.error_message = "There was an issue while fetching fee estimation with  Bitcoin core. Please use manual fee estimation"
```
stepansnigirev, line 155: Maybe we could also append here error message from Bitcoin Core?
The result from failed `estimatesmartfee` looks like this:
```
{
  "errors": [
    "Insufficient data or no feerate found"
  ],
  "blocks": 0
}
```

src/cryptoadvance/specter/util/fee_estimation.py
k9ert: I think you're right. If Core doesn't deliver anything useful, then it has reasonable error-messages but not reasonable default results. In that case, no exception is thrown.

Usually i try to avoid catching pure Exceptions AT ALL but in that case, external calls are done and we're very sure it's worth to recover from there (+ we do `logger.exception(e)` ).

src/cryptoadvance/specter/util/fee_estimation.py
```diff
@@ -69,5 +160,20 @@ def get_fees(specter, config):
...
+        # Not a perfect fix but better than to fix it everywhere in the test-code:
+        if specter.chain != "regtest":
+            fee_estimation_result.add_error_message(
+                "There was an issue while fetching fee estimation with  Bitcoin core. Please use manual fee estimation"
+            )
+            try:
+                for error in specter.estimatesmartfee(1)["errors"]:
+                    fee_estimation_result.add_error_message(error)
+            except Exception as e:
+                logger.exception(e)
```
stepansnigirev, line 176: Can we simplify it and avoid extra rpc call as well?

```suggestion
    fee_estimate = specter.estimatesmartfee(1)
    if "feerate" not in fee_estimate:
        # regtest does not seem to have a reasonable result for estimatesmartfee.
        # The error-message is covering "transactions" and so the cypress-tests are not succeeding
        # Not a perfect fix but better than to fix it everywhere in the test-code:
        if specter.chain != "regtest":
            fee_estimation_result.add_error_message(
                "There was an issue while fetching fee estimation with  Bitcoin core. Please use manual fee estimation"
            )
            for error in fee_estimate.get("errors", []):
                fee_estimation_result.add_error_message(error)
```

Issue #1394, opened 2021-09-12
Don't use descriptor wallets if the node is compiled without sqlite


stepansnigirev:
Currently if you compile core without sqlite Specter fails to create wallets. We should just switch to wallets without descriptors in this case.

Issue #1367, opened 2021-08-26
taproot descriptor is shown as pkh


stepansnigirev:
in the address display using taproot wallets the descriptor is shown incorrectly.
![image](https://user-images.githubusercontent.com/1706012/130980038-1ca8e9f6-b0d5-4142-b1eb-11a859cdc706.png)

Requires changes in `Wallet.get_descriptor` method

Issue #1241, opened 2021-06-18
Unable to create transaction: "Selected Coins Doesnt Cover Full Amount"


sako0938:
I am trying to consolidate coins in my wallet, but something is keeping me from sending all of the utxos that I want to send.

What is strange is it does seem to work with certain UTXOs and not others.

I have "Subtract fees from amount" checkbox checked, because my goal is to end up with a single utxo.

Specter Version: 1.4.5
Running using pip install method:
python3 -m cryptoadvance.specter server

Please let me know if you need any more data or screenshots. 
Thanks!

![specter3](https://user-images.githubusercontent.com/3485377/122590943-762a9d80-d01f-11eb-8c45-56056af7e8dc.png)
 
![specter1](https://user-images.githubusercontent.com/3485377/122590406-bdfcf500-d01e-11eb-80b8-245cd32b4b7a.png)
![specter2](https://user-images.githubusercontent.com/3485377/122590488-dff67780-d01e-11eb-9a1f-47bf269a8a3b.png)

wr30:
I’m having same exact issue, but I have no upgrades yet. Still on version 1.3.1. Have had 0 issues until this today. Will try to upgrade to 1.4.5 and see if it persists.

sako0938:
Im trying to track down the source of this bug. I modified wallet.py to print out the remainder that it is concerned about and it look to be a float issue.
```
            if selected_coins != []:
                still_needed = sum(amounts)
                for coin in selected_coins:
                    coin_txid = coin.split(",")[0]
                    coin_vout = int(coin.split(",")[1])
                    coin_amount = self.gettransaction(coin_txid, decode=True)["vout"][
                        coin_vout
                    ]["value"]
                    extra_inputs.append({"txid": coin_txid, "vout": coin_vout})
                    still_needed -= coin_amount
                    if still_needed < 0:
                        break
                if still_needed > 0:
                    raise SpecterError(
                        f"Selected coins does not cover Full amount! Please select more coins! {still_needed}"
                    )
```
This is the error I get, which shows its missing less than a satoshi, pointing to a zero float issue.

ERROR: Selected coins does not cover Full amount! Please select more coins! 5.551115123125783e-17

sako0938:
I modified the code again to instead compare the remainder to >= 1 sat instead of zero, and I was able to generate my PSBT. Ill try now to sign with my hww.

Let me know if this is PR worthy! Would love to be a contributer.
```
[...]

Issue #1101, opened 2021-04-08
nLockTime (Timelock) is set to 0


truebrown:
I noticed that nLockTime (timelock) of my multisig spend transaction was set to 0.
I understand that wallets should not do this. Wallets should set the timelock to the latest mined block to make fee-sniping less profitable. Example: Electrum sets it to the last block number (highest block height).

Is Specter setting this or is Bitcoin Core setting it?

https://b10c.me/mempool-observations/1-locktime-stairs/

`setting the current block height as a lock-by-block-height forbids a fee-sniping miner to use transactions from the mempool his replacement block. It creates an incentive to move the chain forward to be able to include the next batch of time-locked transactions in a block. While this does not fully mitigate fee-sniping, it reduces profitability. The more transactions enforce a time lock to the current block height, the less profitable fee-sniping is.`

https://en.bitcoin.it/wiki/Timelock
`Although every transaction contains the nLockTime field, every wallet up until recently set nLockTime to 0, meaning the transaction was valid in any block. Starting with Bitcoin Core 0.11.0, every normal transaction automatically generated by began including an nLockTime set to a recent block height as a way to make hypothesized fee sniping less profitable[4]; other wallets are recommended to do the same. Approximately 20% of all bitcoin transactions set a nLockTime value different from zero[5] as of early-2019.`

stepansnigirev:
Bitcoin Core has locktime of 0 as default, but we also set it here:
https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/wallet.py#L1332
and here:
https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/wallet.py#L1368

I think you are right and we should use the current blockheight instead.

moneymanolis:
[...]

PR #1411, opened 2021-09-28
Refactor wallet class

stepansnigirev:
This PR makes a lot of changes in the wallets logic.
The most noticeable changes:
- now internally wallets use combined Embit descriptor. It simplifies address derivation and PSBT parsing
- PSBT decoding is done internally using `util.psbt.SpecterPSBT` and `liquid.util.pset.SpecterPSET` classes - they are independent of the PSBT version, can detect which outputs or inputs belong to wallet etc.
- Refactors RBF logic
- Simplifies PSBT creation logic

Other changes:
- If Bitcoin Core is compiles without SQLite support we fall back to BDB wallets (before Specter was failing)
- Locktime for new transactions is set to current blockchain height

Liquid-related changes:
- Wallet import/export now works
- PSBT import/export now works
- Added SIGHASH_RANGEPROOF for elements hot wallet
- Added check for dynafed activation (either to use SIGHASH_RANGEPROOF or not)

Close https://github.com/cryptoadvance/specter-desktop/issues/1394
Close https://github.com/cryptoadvance/specter-desktop/issues/1367
Close https://github.com/cryptoadvance/specter-desktop/issues/1241
Close https://github.com/cryptoadvance/specter-desktop/issues/1101

### DIFF
--- a/src/cryptoadvance/specter/util/psbt.py
+++ b/src/cryptoadvance/specter/util/psbt.py
@@ -3,7 +3,7 @@ The goal of this module is to slowly migrate from json-like representation of PS
 to a normal PSBT class that does not require RPC calls and can do more things.
 to_dict and from_dict methods are maintained for backward-compatibility
 """
-from embit.psbt import PSBT, InputScope, OutputScope
+from embit.psbt import PSBT, InputScope, OutputScope, DerivationPath
 from embit.transaction import Transaction, TransactionOutput
 from embit.liquid.networks import get_network
 from embit import bip32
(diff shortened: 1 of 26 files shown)


src/cryptoadvance/specter/util/psbt.py
```diff
@@ -0,0 +1,364 @@
...
+from embit.psbt import PSBT, InputScope, OutputScope, DerivationPath
+from embit.transaction import Transaction, TransactionOutput
+from embit.liquid.networks import get_network
+from embit import bip32
+from math import ceil
+import time
+
+
+class SpecterTx:
+    def __init__(self, parent, tx):
```
k9ert, line 15: If we migrate here from json to classes, can't we also use static types in here? If a complete newbie Dev looks at this class, he has no idea what "parent" and "tx" should be even though in the `_to_dict` method, he learns that it needs to have a `txid-method`.
So maybe we can use static type checking for all the constructors of these "value-types" ?
https://medium.com/@ageitgey/learn-how-to-use-static-type-checking-in-python-3-6-in-10-minutes-12c86d72677b

stepansnigirev: Done. I hope it will be easier to understand now.

PR #1414, opened 2021-10-01
Feature: a framework for migrations

k9ert:
We did not yet put enough attention on changes on the data-layer or the organisations of files in the .specter-folder
So this PR includes a framework for migrations and two examples where the second one is a real example migrating the single-bitcoind-instance implemented in v.1.3.0 to a multiple-instance.

### DIFF
--- a/src/cryptoadvance/specter/migrations/migration_0000.py
+++ b/src/cryptoadvance/specter/migrations/migration_0000.py
@@ -1,0 +1,19 @@
+import logging
+from ..specter_migrator import SpecterMigration
+
+logger = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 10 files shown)


kdmukai:
Minor note: Would be nice to hide the `migrations` somewhere a level deeper since they're not quite a core module the way other things at the root are. There isn't really a natural home for it, though.

Maybe move `specter_migrator.py` and the `migrations` subdir under `util`:
```
util
  +-migrations
  |   +-migrations
  |   |   +-migration_0000.py
  |   |   +-migration_0001.py
  |   +-specter_migrator.py
  +-base43.py
  +-base58.py
  +-etc...
```

PR #1415, opened 2021-10-02
Auto privacy settings

kdmukai:
Users shouldn't have to remember to click "Hide Sensitive Info" or logout from Specter. It's especially dangerous in our current era of doing everything over videoconferencing with users streaming their screens over the video feed. We should take pro-active measures to autohide sensitive info and auto logout users after appropriate timeouts. But users should still be able to customize or deactivate these safety measures as they see fit. 

NEW FEATURES:
* Automatically activates the Hide Sensitive Info mode after X minutes have elapsed on the current page. Defaults to 20 minute timeout but can be set to NEVER, 10, 20, 40min or custom.
* IF a login method is enabled, automatically logs a user out after Y hours have elapsed on the current page. Defaults to 1hr but can be set to NEVER, 1, 4, 24hrs or custom.
* These new settings can be applied to the `admin` user and each user account can have their own customized settings.

CHANGED BEHAVIOR:
The Hide Sensitive Info mode is now **always disabled** the next time the user logs in (i.e. sensitive info will be displayed again). Users who expect their info to remain hidden on login will need to take care. If there is enough trouble over this, a new setting can be added for something like "Always/Never show sensitive info after login".

Rationale for the change: Using the new defaults, a typical user will first trigger the auto hide timeout which, for better and worse, still writes to persistent user config. They will then get auto logged out after more time elapses. When they finally log back in, the "Show Sensitive Info" toggle would normally be set to `False` at this point (reflecting its last state). So the user who's session timed out while they were away returns to Specter with everything obscured.

This seems kind of silly as the user _just_ logged in; we shouldn't need to hide anything from them at that point (so perhaps letting that "Hide" setting persist through the next login was always kind of nonsensical).

MISC:
* minor js bugfix in `auth_settings.jinja` when selecting different auth types.
* minor note in docs to ensure debug mode is on in local dev

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -94,7 +94,7 @@ Run the server:
 
 ```sh
 cd specter-desktop
-python3 -m cryptoadvance.specter server --config DevelopmentConfig
+export FLASK_ENV=development && python3 -m cryptoadvance.specter server --config DevelopmentConfig
 ```
 
 #### If `pip install` fails on `cryptography==3.4.x`
--- a/src/cryptoadvance/specter/managers/config_manager.py
+++ b/src/cryptoadvance/specter/managers/config_manager.py
@@ -66,6 +66,8 @@ class ConfigManager(GenericDataManager):
             "fee_estimator": "mempool",
             "fee_estimator_custom_url": "",
             "hide_sensitive_info": False,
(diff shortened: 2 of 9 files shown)


kdmukai:
Notes: This PR depends on #1417 which should be merged before this.

PR #1424, opened 2021-10-10
Bugfix: refactor WalletManager and Wallet

k9ert:
This PR is targetting too many calls of WalletManager._update. After that is done, this should also remove the dependency from `Wallet` to `WalletManager` but one thing at a time.


k9ert:
Tested this with renaming wallets: Works as well @stepansnigirev

stepansnigirev (approved):

PR #1443, opened 2021-10-21
Tweaks to local dev setup

kdmukai:
Previous versions of these instructions included this setup step:
```
pip3 install -e .
```
But this can make local dev behave in unexpected ways:
* If you start the server from the root dir, everything runs as expected but local code changes are ignored as it is using the version of the code that gets copied to the virtualenv's `site-packages` dir.
* If you start the server from the `src/` subdir, the above problem is solved but in many (all?) cases, local jinja template changes are ignored and only the versions that existed at the time of the `pip install -e .` are run.

Also fixing server launch instructions to properly make use of the `--debug` option.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -196,6 +196,14 @@ there is a git [pre-commit hook](https://ljvmiranda921.github.io/notebook/2018/0
 pre-commit install
 ```
 
+If this throws an error: `command not found: pre-commit`, you need to install `pre-commit`.
+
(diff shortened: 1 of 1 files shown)


k9ert:
I didn't have the issues you describe with `pip3 install -e .` but my changes are not picked up if i ommit it. I'm not sure where the differences are coming from and how the changes should be picked up if you ommit it (no matter whether `src` or not).

Maybe you can clarify?

PR #1453, opened 2021-10-28
Encrypted user data storage

kdmukai:
* Enhances the `GenericDataManager` to optionally include fields that are encrypted at rest (but are plaintext when loaded).
* Adds an `encrypted_user_secret` to the `User` json. The user secret is a `Fernet` key that is encrypted during storage using the user's password. **This AES encryption needs a thorough review.**
* A new `encrypted_user_secret` is generated if the `User` json doesn't yet have one.
* When the `encrypted_user_secret` is decrypted at login it is stored in plaintext in the `User`. It can then be used at any time to encrypt or decrypt the encrypted fields in `GenericDataManager`. This design allows us to avoid storing the `User`'s password but still be able to do on-demand encrypt/decrypt.
* Adds `GenericDataManager.encrypted_storage_version` so that any future encryption changes can be differentiated and migrate older data forward.
* Extensive tests added for `User` around the `user_secret` generation, encryption, and decryption.
* Test added for `GenericDataManager`'s new encrypted fields.

Note: `GenericDataManager` is in `managers` but is really just a base Storage class. The concept of a "manager" is quite blurred at the moment. For example, ideally a `User` instance would be created and manipulated by the `UserManager` which in turn would manage on-disk storage through a base Storage class. But currently the base Storage class *IS* the base Manager class (`GenericDataManager`). The `UserManager`'s storage functionality really should have been implemented in something like a `UserStorage` class that is instantiated and controlled within `UserManager`.

Dependencies:
* Adds `pycryptodome` pip dependency for AES encryption/decryption support for the new `user_secret`. `requirements.in` and `requirements.txt` updated accordingly.

Misc notes:
* Update to DEVELOPMENT.md for correct way to launch in debug mode.


src/cryptoadvance/specter/user.py
```diff
@@ -36,69 +44,146 @@ def verify_password(stored_password, provided_password):
...
+        # create the cipher config
+        cipher = AES.new(private_key, AES.MODE_GCM, nonce=nonce)
+
+        # decrypt the cipher text
+        self.plaintext_user_secret = cipher.decrypt_and_verify(cipher_text, tag)
+
+    def _generate_user_secret(self, plaintext_password):
+        # Encryption using the user_secret uses a Fernet key. But the Fernet
+        #   key itself will be encrypted with the user's password.
+        self.plaintext_user_secret = Fernet.generate_key()
```
k9ert, line 159: I'm a bit confused. Shouldn't that be done just initially? If the user sets a new password, the `self.plaintext_user_secret` should be the same, no? Here, it's created anew every time this method get called (e.g. in `set_password`).

src/cryptoadvance/specter/user.py
```diff
@@ -36,69 +44,146 @@ def verify_password(stored_password, provided_password):
...
+        # Must keep encrypted_user_secret in sync with password changes
+        if self.encrypted_user_secret is None:
+            self._generate_user_secret(plaintext_password)
+        else:
+            if self.plaintext_user_secret is None:
+                raise Exception(
+                    "encrypted_user_secret wasn't decrypted during user login"
                 )
-        except Exception as e:
-            raise SpecterError(f"Unable to parse user JSON.:{e}")
```
k9ert, line 93: please `handle_exception(e)` whenever raising Exceptions (or be more specific in the exception you're catching).
We never know what kind of exception will come up here and if it's an unexpected one, we definitely want to have a stacktrace (and handle_exception will print one out).

tests/test_user.py
```diff
@@ -0,0 +1,119 @@
...
+    first_plaintext_user_secret = user.plaintext_user_secret
+
+    new_password = "mynewpassphrase"
+    user.set_password(new_password)
+
+    # The new encrypted_user_secret will be different...
+    assert first_encrypted_user_secret != user.encrypted_user_secret
+
+    # ...but the plaintext_user_secret remains unchanged
+    assert first_plaintext_user_secret == user.plaintext_user_secret
```
k9ert, line 119: Here i see that you have the same assumption than me in my comment above. However i don't understand how it works like this.

src/cryptoadvance/specter/server_endpoints/auth.py
```diff
@@ -141,9 +146,13 @@ def register():
                 "explorers": {"main": "", "test": "", "regtest": "", "signet": ""},
                 "hwi_bridge_url": "/hwi/api/",
             }
+
+            # TODO: These four lines should be a single call to UserManager
```
k9ert, line 150: Yes, i agree! Why not do it now?

k9ert (approved):
Have some comments but none of them critical. LGTM, some changes would be good!

stepansnigirev:
Would be better to use cryptography package
https://cryptography.io/en/latest/

The best is to use some defaults schemes, or something like aes-cbc + hmac

src/cryptoadvance/specter/managers/genericdata_manager.py
```diff
@@ -33,30 +68,41 @@ def load(self):
...
 
-    @classmethod
-    def convert_to_list_of_dict(cls, some_list):
-        """This is not yet used but might be convenient
-        An array of a specific type gets converted to a list of dicts
-        Maybe this method simply does nothing, though
-        """
-        return some_list
-        # example implementation for users:
-        # return [u.json for u in some_list]
```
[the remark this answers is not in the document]
kdmukai: Yeah, it wasn't clear to me how old these placeholders were or if there was still any anticipated use for them. Happy to delete them.

src/cryptoadvance/specter/user.py
kdmukai: Yes, I think you're right. The existing `plaintext_user_secret` needs to be re-encrypted on password change but regenerating would render the downstream encrypted data orphaned.

src/cryptoadvance/specter/server_endpoints/auth.py
kdmukai: I was trying to minimize the footprint of the PR, but yes, will update with this change!

kdmukai:
> Would be better to use cryptography package https://cryptography.io/en/latest/
> 
> The best is to use some defaults schemes from this lib (fernet), or something like aes-cbc + hmac

Updated code now uses the example listed here:
https://cryptography.io/en/latest/fernet/#using-passwords-with-fernet

User's password is used as the Fernet key to encrypt the `user_secret`.

kdmukai:
fyi since the pyca/cryptography docs reference Django's iteration count, I matched Django's current 390k iterations:

https://github.com/django/django/blob/main/django/contrib/auth/hashers.py#L274

k9ert:
LGTM, let's merge.

PR #1454, opened 2021-10-29
Feature: Send dialog refactored and sped up

k9ert:
This is refactoring and speeding up the "send_new" dialog:
* The coin-selection and fee-selection jinja templates get replaced by web-components
* The coin-selection webcomponent is reusing the `tx-table` webcomponent which gets heavily refactored for that purpose
* Each request does now get time-measured. If the call takes longer than `REQUEST_TIME_WARNING_TRESHOLD` , then you get a warning via flash. Default is 20 seconds for now. There is also a small perfomance-improvement-guide
* The tracing-code is now also including how long an RPC-call took towards BitcoinCore.
After:
```
[  DEBUG] in             rpc: call(wallet/specter/feasfd       )(    4ms)  payload:[]
```
Before:
```
[  DEBUG] in             rpc: call(wallet/specter/feasfd       ) payload:[]
```


src/cryptoadvance/specter/server_endpoints/wallets_api.py
```diff
@@ -0,0 +1,933 @@
...
+        )
+        # add a filename
+        response.headers.set("Content-Disposition", "attachment", filename="utxo.csv")
+        return response
+    except Exception as e:
+        handle_exception(e)
+        return _("Failed to export wallet utxo. Error: {}").format(e), 500
+
+
+@wallets_endpoint_api.route("/wallet/<wallet_alias>/asset_balances")
```
kdmukai, line 544: Looks like this needs `methods=["POST"]` specified to work with the JS.

Seeing:
```
[2021-12-13 09:22:36,500] INFO in _internal: 127.0.0.1 - - [13/Dec/2021 09:22:36] "POST /wallets/wallet/seedsigner_demo/asset_balances HTTP/1.1" 302 -
[2021-12-13 09:22:36,508] INFO in _internal: 127.0.0.1 - - [13/Dec/2021 09:22:36] "GET /wallets/wallet/seedsigner_demo/asset_balances HTTP/1.1" 200 -
```

Which seems to trigger the "Session expired. Please refresh and try again" error on the _next_ url.

src/cryptoadvance/specter/server_endpoints/wallets_api.py
```diff
@@ -0,0 +1,933 @@
...
+                "balance": int(wallet.full_available_balance * 1e8),
+                "label": "sat",
+            },
+        }
+        for asset in wallet.balance.get("assets", {}).keys():
+            asset_balances["asset"] = {
+                "balance": wallet.balance.get("assets", {})
+                .get(asset, {})
+                .get("trusted", 0),
+                "label": "{{{% endfor %} asset | assetlabel }}",
```
kdmukai, line 573: This jinja content can't make sense here, can it?

kdmukai:
Should `wallets_api` be part of the `api` subdir? I'm not sure how the Specter API is meant to be used. I suppose there's maybe a difference between API-level access vs minor data that a JS presentation page might need? I don't have a strong opinion either way. Just thinking out loud here.

src/cryptoadvance/specter/templates/includes/coin_selection.html
```diff
@@ -0,0 +1,232 @@
...
+            shadow.appendChild(clone);
+        }
+        /**
+         * Browser calls this method when the element is added to the document
+         * (can be called many times if an element is repeatedly added/removed)
+        */
+        connectedCallback() {
+            this.walletAlias = this.getAttribute('wallet-alias');
+            this.spendableAmount = this.getAttribute("spendable-amount") ?? 0
+            this.assetBalances = this.fetchAssetBalances()
```
kdmukai, line 71: I think this would be better off as a json blob `data-` attribute so that we can eliminate the fetch.

src/cryptoadvance/specter/templates/includes/coin_selection.html
```diff
@@ -0,0 +1,232 @@
...
+        async fetchAssetBalances(wallet_alias) {
+            let url = `{{ url_for('wallets_endpoint_api.asset_balances', wallet_alias='WALLET_ALIAS') }}`.replace("WALLET_ALIAS", this.walletAlias);
+            var formData = new FormData();
+            formData.append('address', this.address);
+            formData.append('csrf_token', '{{ csrf_token() }}');
+            try {
+                const response = await fetch(
+                    url,
+                    {
+                        method: 'POST',
```
kdmukai, line 205: Shouldn't this be a GET when just retrieving data? (that would also solve the `POST` issue in `wallets_api.py`).

kdmukai:
Because the fee estimation is potentially slow, it definitely makes sense to retrieve it `async` in the JS.

But there's still a slow and seemingly redundant call here:
https://github.com/cryptoadvance/specter-desktop/blob/032b1fe1209a63c128feefe3a88be0574f00fb51/src/cryptoadvance/specter/server_endpoints/wallets.py#L466

I can't tell if the `estimate_fee` option in there pulls fee data only from the POST or if maybe makes another call out in python. Obv that should be avoided since the JS just got the latest data.

Separate issue: That endpoint is doing a lot. Seems like it would be better to refactor out some of those POST actions to their own dedicated endpoints.

src/cryptoadvance/specter/server_endpoints/wallets_api.py
k9ert: Definitely. Fixed.

k9ert:
> Should wallets_api be part of the api subdir? I'm not sure how the Specter API is meant to be used. 
> I suppose there's maybe a difference between API-level access vs minor data that a JS presentation 
> page might need? I don't have a strong opinion either way. Just thinking out loud here.

It's a good point and i'm thinking about this as well. However, there is a catch: The REST-Api currently works via basic-Auth. all the stuff under server_endpoints which is returning json is using the same auth than the normal frontend. That's why the fetch-API is working so seamlessly. So i agree that it would be cool to better split that ups: Ideally it should be possible to use each json-endpoint with both auth-mechanisms but not sure how to achieve that best.
At least, we can't yet simply move the wallet_api stuff to the rest-API directories.

k9ert:
> But there's still a slow and seemingly redundant call here:
Yes, removed!

> I can't tell if the estimate_fee option in there pulls fee data only from the POST or if maybe makes another call 
> out in python. Obv that should be avoided since the JS just got the latest data.

Maybe a better name for it would be "draft-run". It's only creating the psbt and returns it.

> Separate issue: That endpoint is doing a lot. Seems like it would be better to refactor out some of those POST 
> actions to their own dedicated endpoints.

Yes. I carved out the `estimate_fee` action and put it to the `wallet_api` where it belongs to. However, splitting the rest needs some more thought:
There is some clever error-handling in that complexity. When an error occurs, all the data get returned and you don't have to start over filling all the fields in the form. Replicating that in each endpoint would result in MANY more lines of duplicated code. Basically this:
```
    show_advanced_settings = (
        ui_option != "ui"
        or subtract
        or fee_options != "dynamic"
        or not rbf
        or selected_coins
    )
    wallet_utxo = wallet.utxo
    if app.specter.is_liquid:
        for tx in wallet_utxo + rbf_utxo:
            if "asset" in tx:
                tx["assetlabel"] = app.specter.asset_label(tx.get("asset"))
    return render_template(
        "wallet/send/new/wallet_send.jinja",
        psbt=psbt,
        ui_option=ui_option,
        recipients_txt=recipients_txt,
        recipients=list(zip(addresses, amounts, amount_units, labels)),
        subtract=subtract,
        subtract_from=subtract_from,
        fee_options=fee_options,
        rbf=rbf,
        selected_coins=selected_coins,
        show_advanced_settings=show_advanced_settings,
        rbf_utxo=rbf_utxo,
        rbf_tx_id=rbf_tx_id,
        wallet_utxo=wallet_utxo,
        wallet_alias=wallet_alias,
        wallet=wallet,
        specter=app.specter,
        rand=rand,
        error=err,
    )
```
Wrapping that in an own function would need a function with18 params which doesn't really make it any better. Maybe we should pass the form around? I'm not sure yet.

src/cryptoadvance/specter/templates/includes/coin_selection.html
k9ert: Yes, done!

src/cryptoadvance/specter/templates/includes/coin_selection.html
k9ert: I passed it as a property. That's much easier as you don't have to fiddle with json and you don't have to implement `attributeChangedCallback` and `observedAttributes`.
Not entirely happy about the implementation, though. The assetBalances are not yet retrieved via the fetch-API in the wallet_sen.jinja but by templating in there.
But maybe we can change that in some other PR.

src/cryptoadvance/specter/server_endpoints/controller.py
```diff
@@ -145,6 +148,33 @@ def selfcheck():
...
+    if (
+        (response.response)
+        and (200 <= response.status_code < 300)
+        and (response.content_type.startswith("text/html"))
+    ):
+        treshold = app.config["REQUEST_TIME_WARNING_TRESHOLD"]
+        if diff > treshold:
+            flash(
+                _(
+                    "The request before this one took {} seconds which is longer than the threshold ({}). Checkout the perfomance-improvement-hints in the documentation".format(
```
moneymanolis, line 169: My proposal here: "The last request took very long. Check out the performance improvement-hints in the documentation"
I would also add a link to `performance.md` on GitHub (remark, one needs the Markup class from flask to show a href in flash message: https://stackoverflow.com/questions/21248718/how-to-flashing-a-message-with-link-using-flask-flash)

src/cryptoadvance/specter/rpc.py
```diff
@@ -413,7 +428,9 @@ def trace_call(cls, url, payload):
                 else:
                     cls.last_call_hash = current_hash
             logger.debug(
-                "call({: <28}) payload:{}".format("/".join(url.split("/")[3:]), payload)
+                "call({: <28})({: >5}ms)  payload:{}".format(
```
moneymanolis, line 431: Using` {:^}` would make it more readable.

src/cryptoadvance/specter/templates/wallet/send/new/wallet_send.jinja
```diff
@@ -350,7 +361,7 @@
 			subtract.checked = (units[i] == 'sat' || units[i] == 'btc');
 			if (amounts.length > 1) {
 				document.getElementById('subtract_from').style.display = 'block';
-				document.getElementById('subtract_from_input').value = i + 1;
+				document.getElementById('substract_from_input').value = i + 1;
```
moneymanolis, line 364: Typo --> subtract

src/cryptoadvance/specter/templates/includes/coin_selection.html
```diff
@@ -0,0 +1,201 @@
+<template id="coin-selection">
+    <span class="toggle_coinselection" style="cursor: pointer;">{{ _("Coin selection") }} {% if show_advanced_settings %}&#9660;{% else %}&#9654;{% endif %}</span>
```
moneymanolis, line 2: Not sure about intended behaviour, should coin selection only be selectable (down-pointing triangle aka &#9660) if advanced settings are set to True? In testing it looks like coin selection is independent from the advanced settings.

src/cryptoadvance/specter/templates/includes/coin_selection.html
```diff
@@ -0,0 +1,201 @@
...
+            blockhash="true" hide-switcher="true">
+        </tx-table>
+    </div> 
+</template>
+
+
+<script type="module">
+    /**
+     * Custom element for showing the coin-selection-dialog
+       attributes:
```
moneymanolis, line 22: Minor thing but aren't `is-liquid` and `selected-coins` also attributes of the web component?

k9ert:
Let's merge this. Thanks for the review and the refinements :1st_place_medal:

PR #1457, opened 2021-11-02
Bugfix: no threading for update after creation of wallets

k9ert:
Importing a backup or a single Wallet results sometimes in flask-messages like "Failed to import wallet {}".
As the `handle_exception` call reavealed, this is due to "wallet_not existing" in get_by_alias in https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/server_endpoints/settings.py#L154
It makes sense when you look some lines before where the update-call will be done in threading-mode to improver performance.

When creating wallets, we have to wait on the update so the update-method now allows preventing threading.

On top, there are some changes for the upcoming release to make the release-process more smooth.

### DIFF
--- a/utils/release-helper.py
+++ b/utils/release-helper.py
@@ -82,6 +82,7 @@ class ReleaseHelper:
         pass
 
     def init_gitlab(self):
+        # https://python-gitlab.readthedocs.io/en/stable/api-usage.html
         import gitlab
 
         if os.environ.get("GITLAB_PRIVATE_TOKEN"):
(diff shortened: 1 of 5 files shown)


utils/release-helper.py
```diff
@@ -138,7 +143,10 @@ def init_gitlab(self):
                 if pipeline.ref == self.tag:
                     self.pipeline = pipeline
                     logger.info(f"Found matching pipeline: {pipeline}")
-            if not self.pipeline:
+            if not hasattr(self, "pipeline"):
+                logger.error(
+                    f"Could not find tag {self.tag} in the peipline-refs {[pipeline.ref for pipeline in self.project.pipelines.list()]}"
```
moneymanolis, line 148: typo: pipeline-refs

moneymanolis (approved):

PR #1470, opened 2021-11-13
More options for address labels import

### LINKED ISSUE #1419: Add CSV Address (and txn) Label Imports
I have had issues keeping address labels across Specter instances (i.e. the labels disappear when I import the wallet into a new specter-desktop instance). However, I had the specter-desktop CSV label file that can be exported by the wallet: `${WALLET_NAME}_addr.csv`

In the Settings -> Advanced tab, the first option is `Import Electrum address labels` but these require a json format of 

```
{"8aasd324..": "some utxo label",
...,
"12asdu34q5...": "some txn label"}
```

I was able to import my CSV by doing a CSV to JSON conversion using the python pandas library, but I think this would be a waste of resources to use this approach in specter.

Instead I suggest using the following code as a guide and adding in a CSV to JSON parser so that specter can consume it's own label export CSVs.

Example code taken from (https://pythonexamples.org/python-csv-to-json/) :

```
import csv 
import json 

def csv_to_json(csvFilePath, jsonFilePath):
    jsonArray = []
      
    #read csv file
    with open(csvFilePath, encoding='utf-8') as csvf: 
        #load csv file data using csv library's dictionary reader
        csvReader = csv.DictReader(csvf) 

        #convert each csv row into python dict
        for row in csvReader: 
            #add this python dict to json array
            jsonArray.append(row)
  
    #convert python jsonArray to JSON String and write to file
    with open(jsonFilePath, 'w', encoding='utf-8') as jsonf: 
        jsonString = json.dumps(jsonArray, indent=4)
        jsonf.write(jsonString)
          
csvFilePath = r'data.csv'
jsonFilePath = r'data.json'
csv_to_json(csvFilePath, jsonFilePath)

```

moneymanolis:
### Electrum labels import functionality extended
Now also Specter CSV address lists (video 1 below) and Specter JSONs (video 2) can be used to import labels if labels got corrupted or lost - for example when switching from one Specter instance to another. Tests for both were added. 
This resolves #1419 
There will probably be another update to this functionality because of #1395 once we have a web component for the file upload.

Specter CSV import:
https://user-images.githubusercontent.com/70536101/141657763-2396a3de-f34f-431a-8561-c6ebb161100b.mov

Specter JSON import:
https://user-images.githubusercontent.com/70536101/141657783-23b40bce-2c88-49ee-8b6c-885d8e3ac74e.mov

### Some additional minor changes in this PR: 
Change of redirect in the errorhandler for the `SpecterError` (to `request.url` for POST requests) and a fringe case fix in VersionChecker.


src/cryptoadvance/specter/server_endpoints/controller.py
```diff
@@ -63,13 +63,16 @@ def server_rpc_error(rpce):
...
+    """Specific SpecterErrors get passed on to the User as flash"""
     flash(str(se), "error")
     try:
         app.specter.wallet_manager.update()
     except SpecterError as se:
         flash(str(se), "error")
-    return redirect(url_for("about"))
+    if request.method == "POST":
+        return redirect(request.url)
+    else:
```
k9ert, line 74: Might not be obvious why we're doing this here. So put a comment there, like
```suggestion
    # potentially avoiding http loops. Might be improvable but how?
    else:
```

k9ert (approved):
Great! LGTM and ready to merge. Just one tiny comment.

src/cryptoadvance/specter/wallet.py
```diff
@@ -529,36 +531,70 @@ def fetch_transactions(self):
...
+            except (Error, KeyError) as e:
+                raise SpecterError(
+                    f"Labels import failed. Check the import info box for the expected formats. Error: {e}"
+                )
+        # convert labeled_addresses to arr (for AddressList.set_labels), only allow existing addresses for which a label had been set
         arr = [
             {"address": address, "label": label}
             for address, label in labeled_addresses.items()
-            if address in self._addresses
+            if address in self._addresses and self._addresses[address].is_labeled
```
relativisticelectron, line 593: While this change might be the intended behavior for your specter-use-case, it is NOT for the electrum use case.  This could be copied and moved to the if statements to facilitate different behavior.

tests/test_labels_import.py
```diff
@@ -112,20 +113,37 @@ def test_electrum_label_import(
...
             json.dumps({"12dRugNcdxK39288NjcDV4GX7rMsKCGn6B": "address label"})
         )
         assert wallet._addresses[test_address]["label"] is None
         assert len(wallet._addresses) == number_of_addresses
 
         # Test it with a txid label
-        wallet.import_electrum_label_export(json.dumps({txid: "txid label"}))
+        wallet._addresses[test_address].set_label(
+            "some_label"
+        )  # some label has to be set before
```
relativisticelectron, line 125: These 3 lines have to be removed again.

relativisticelectron (commented):
This PR breaks the electrum import use-case! Users should be able to import labels of addresses, that had no label before.

PR #1478, opened 2021-11-16
Bugfix: csrf-issues when /toggle_hide_sensitive_info endpoint got called

k9ert:
When javascript times out and calls the /toggle_hide_sensitive_info endpoint, it shouldn't have any csrf-issues which could occur for a lot of reasons.
As this is only increasing security, shouldn't be an issue.


kdmukai (approved):
Works great!

Confirmed on the Settings page that sets the autohide timeout. Save the settings but stay on the page and wait for the autohide timeout. Previous code would hit a popup prompt that blocks autohide from activating. New code simply flows through and sets privacy mode as expected.

PR #1482, opened 2021-11-17
Taproot test case

kdmukai:
* Adds a basic Taproot wallet creation and deposit test case. More test cases to come (e.g. Taproot spend).
* Bumps pytest suite to Bitcoin Core v22+ in order to test Taproot wallets
* Increases `TestConfig` timeout to fix timeouts even on a M1 Pro MacBook w/32GB of RAM (the RPC calls seem way too slow in general. Maybe there's an optimization we're missing).
* Assumes the production v0.21.0 Elements node (will have to update CirrusCI's installed node?).
* Fixes inconsistent `test_WalletImporter_integration` test case that had an unpredictable timing issue with the RPC call.
* improved  `install_noded.sh` script to to deal with the new v22+ inconsistent binary naming convention (22.0/22.0.0 vs 0.21.1).

NOTE: pytest fails `tests/test_node.py::test_Node_elm` as expected. It needs its Elements node cleared out so the production v0.21.0 can be installed.


tests/install_noded.sh
```diff
@@ -79,18 +79,24 @@ function maybe_update {
...
-        PINNED=$(cat ../pytest.ini | grep "addopts = " | cut -d'=' -f2 |  sed 's/--/+/g' | tr '+' '\n' | grep ${node_impl} |  cut -d' ' -f2)
-       
-        if [ "$node_impl" = "elements" ]; then
-            # in the case of elements, the tags have a "elements-" prefix
-            PINNED=$(echo "$PINNED" | sed 's/v//' | sed 's/^/elements-/')
-        fi
+    if [ "$node_impl" = "bitcoin" ]; then
+        PINNED=v22.0
+    elif [ "$node_impl" = "elements" ]; then
+        PINNED=v0.21.0
```
k9ert, line 87: This doesn't look right. It's hardcoding the version here?! I'm aware that the above code is not particulary nice and maybe even not robust. Didn't it work anymore?

kdmukai: @k9ert the problem is that v22+ changes the numbering convention. The existing code was parsing v22 into "v22.0.0" (note the extra ".0" at the end) which then fails the binary download because no such version number exists. I didn't have the bash/regex skills to handle both version number formats (v0.21.0 vs v22.0).

k9ert: Really stupid. The `getnetworkinfo()["subversion"]` call is returning `/Satoshi:22.0.0/` however the binaries are skipping one [zero, like 22.0 ](https://bitcoincore.org/bin/bitcoin-core-22.0/).

@Sjors maybe you have an idea why that is? Is it worth filing an issue about? It creates needless friction e.g. here.

k9ert (approved):

tests/install_noded.sh
Sjors: I opened an issue: https://github.com/bitcoin/bitcoin/issues/23615

PR #1499, opened 2021-11-30
Bugfix: deactivate internal Tor for non-AppImages

### LINKED ISSUE #1196: Tor setup is broken (Ubuntu 20.04, pip installation)
After installing specter via pip in Ubuntu 20.04, setting up the built-in tor fails with the following error message:

> ERROR in tor_setup_tasks: Failed to install Tor. Error: [Errno 2] No such file or directory: '/home/xphade/.venv/specter/lib/python3.8/site-packages/cryptoadvance/specter/util/../../../../pyinstaller/torbrowser/tor-browser-linux64-10.0.15_en-US.tar.xz'

The whole `pyinstaller` subdirectory is actually missing. I was able to work around the error by simply downloading the archive and putting it into the given path. 

This happened to me on a native Ubuntu 20.04 install as well as on a WSL2 Ubuntu installation in Windows. I don't think it's exclusive to Ubuntu, though, because I see some people complaining about it in the Umbrel forum as well. See here: https://community.getumbrel.com/t/running-tor-on-specter/1139

-----

Another (maybe unrelated) issue: On native Ubuntu, I cannot use the built-in tor at all because after installation I'm getting an error that tor could not be started, due to returning non-zero exit code 127. I think it fails at the following function: https://github.com/cryptoadvance/specter-desktop/blob/fb64940812438601ff177fde0fe8e8910867ad0e/src/cryptoadvance/specter/tor_daemon.py#L58-L61

This only happens on native Ubuntu, however. On Windows, tor is running fine (after the workaround described above). Should I create another issue for this?

k9ert:
fixes #1196

### DIFF
--- a/src/cryptoadvance/specter/server_endpoints/setup.py
+++ b/src/cryptoadvance/specter/server_endpoints/setup.py
@@ -1,32 +1,25 @@
-import copy, random, json, os, threading, shutil, logging
+import copy
+import json
+import logging
+import os
+import platform
+import random
+import shutil
+import sys
+import threading
 
-from flask import (
-    Flask,
-    Blueprint,
-    render_template,
-    request,
-    redirect,
-    url_for,
-    jsonify,
-    flash,
-)
-from flask_babel import lazy_gettext as _
-from flask_login import login_required, current_user
+from flask import Blueprint
 from flask import current_app as app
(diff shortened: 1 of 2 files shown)


moneymanolis (approved):
Alright!

PR #1500, opened 2021-11-30
Bugfix: prevent using keys twice

### LINKED ISSUE #1137: Same device can be imported multiple times
Not really really sure if this is bug or feature.

It is possible to import same device (with same keys) multiple times. It is also possible to use them as cosigners in multi-sig, in such a case:
* consigner's xpubs will be identical in the wallet info
* tx will be completely signed after the first signature

k9ert:
Preventing that keys got used twice in a multisig if they are the same
fixes #1137


stepansnigirev (commented):
Two more efficient versions:
Just detect that some key is not unique:
```py
if len(set(keys)) != len(keys):
    raise SpecterError("Keys are the same")
```
If you want to raise with key value:
```py
for key in keys:
    if keys.count(key) > 1:
        raise SpecterError(f"Key {key} is used more than once")
```

Also what about keys from the same device (same root fingerprint) but different derivations?
What about the same key but with different origins? Should we check these cases?

k9ert:
I've simplyfied the calculation as to your suggestion.
> Also what about keys from the same device (same root fingerprint) but different derivations?
> What about the same key but with different origins? Should we check these cases?
I would allow different derivations. Even though the seed would be the same, the hardware-devices could be different and create different signatures. Especially if you non-transient devices in your mind like a trezor (maybe better a coldcard).

What do you mean with different origins?

k9ert:
Thinking about it and i agree that same root-fingerprint but different derivations should only be allowed if ALL of the keys are coming from the same master. I could imagine some casa-style hardware-focused setups where this makes sense. But if you do it, at least you can ask that all the devices have the same masterseed.
Creating a warning would also be a good idea but not sure whether it's possible in this PR.

stepansnigirev:
I am not sure I understand the use case with the same fingerprint. If all keys are derived from the same master how is it different from a single-sig wallet?
By different origins I mean keys where xpub is the same but derivation path either contains derivation from root or not (like in electrum), or if slip132 is used or not.

Here are three keys that are effectively the same, but use different representation:
```
[f3e6eaff/84h/0h/0h]xpub6C5cCQfycZrPJnNg6cDdUU5efJrab8thRQDBxSSB4gP2J3xGdWu8cqiLvPZkejtuaY9LursCn6Es9PqHgLhBktW8217BomGDVBAJjUms8iG
xpub6C5cCQfycZrPJnNg6cDdUU5efJrab8thRQDBxSSB4gP2J3xGdWu8cqiLvPZkejtuaY9LursCn6Es9PqHgLhBktW8217BomGDVBAJjUms8iG
zpub6qk8ok1ouvwM1NkumKnsteGf1F9UUNshFdFdXEDwph8nQFaj8qEFry2cxoUveZCkPpNxQp4KhQwxuy4R7jXDMMsKkgW2yauC2dHbWYnr2Ee
```

We can define `__eq__` statement in the `Key` class that only checks canonical xpub equality for example.

moneymanolis (changes requested):

moneymanolis (approved):
Now it looks good!

PR #1508, opened 2021-12-11
fix keypoolrefill perfomance and adjusted callback logic

k9ert:
When a wallet is created, the keypoolrefill-function get called for each address of probably 50 addresses wasting lots of time to write a csv-file dozens of times. This fixes this behaviour (in `wallet.py`).

It also adjusts the behaviour of the `SPECTER_PERSISTENCE_CALLBACK` (or `SPECTER_PERSISTENCE_CALLBACK_ASYNC` for a script which would run in a thread) to make it more flexible. This env-var takes an executable which will get called any time the persistence-layer is modifying a file.
Example:
* You have written a script called `/usr/local/bin/my_rsync_script.sh`
* You specify that script like `export SPECTER_PERSISTENCE_CALLBACK=/usr/local/bin/my_rsync_script.sh`
* In the past, that script simply get called
* Now if gets additional parameters:
  * the mode which is either `write` or `delete`
  * the filepath
* So an example-call would be `/usr/local/bin/my_rsync_script.sh write /home/someuser/.specter/config.json`
* You can also specify it in the `SPECTER_PERSISTENCE_CALLBACK_ASYNC` and it'll be called in a thread.
Those changes are in `persistence.py`.

Also this PR contains an autodetection-mechanism for Liquid-nodes. So in parallel to `BTC_RPC_USER, BTC_RPC_PASSWORD, BTC_RPC_HOST, BTC_RPC_PORT, ELM_RPC_PROTOCOL` we now also have `ELM_RPC_USER, ELM_RPC_PASSWORD, ELM_RPC_HOST, ELM_RPC_PORT ELM_RPC_PROTOCOL`.
This was a bit tricky as a node was, up to now, indifferent about the type (until it queried the RPC-interface). Now we need to specify the type of the node upfront in order to make the autodetection work if someone uses both autodetection-mechanisms at the same time.
Also, the `NodeManager` created an initial Node if there wasn't one. So we needed to adjust that and we create a second Liquid-Node if there isn't one (AND an env-var is existing). See the node_type as the first argument (which will then differ in which Env-vars to pick up):
```
            if os.environ.get("ELM_RPC_USER"):
                self.add_node(
                    node_type="ELM",
                    name="Blockstream Liquid",
                    autodetect=True,
                    datadir=get_default_datadir(node_type="ELM"),
                    user="",
                    password="",
                    port=7041,
                    host="localhost",
                    protocol="http",
                    external_node=True,
                    default_alias=self.DEFAULT_ALIAS,
                )
            self.add_node(
                node_type="BTC",
                name="Bitcoin Core",
                autodetect=True,
                datadir=get_default_datadir(),
                user="",
                password="",
                port=8332,
                host="localhost",
                protocol="http",
                external_node=True,
                default_alias=self.DEFAULT_ALIAS,
```

### DIFF
--- a/src/cryptoadvance/specter/internal_node.py
+++ b/src/cryptoadvance/specter/internal_node.py
@@ -55,6 +55,7 @@ class InternalNode(Node):
             protocol,
             False,
             fullpath,
+            "BTC",
             manager,
         )
 
--- a/src/cryptoadvance/specter/liquid/rpc.py
+++ b/src/cryptoadvance/specter/liquid/rpc.py
@@ -1,21 +1,21 @@
-from ..rpc import RpcError, BitcoinRPC
+import copy
+import logging
+from io import BytesIO
+
+from embit.descriptor.checksum import add_checksum
 from embit.ec import PrivateKey, PublicKey
 from embit.hashes import tagged_hash
-from embit.descriptor.checksum import add_checksum
-from embit.liquid.descriptor import LDescriptor
-from embit.liquid.pset import PSET, PSBTError
+from embit.liquid import finalizer, slip77
 from embit.liquid.addresses import addr_decode
 from embit.liquid.addresses import address as liquid_address
+from embit.liquid.descriptor import LDescriptor
 from embit.liquid.networks import get_network
+from embit.liquid.pset import PSET, PSBTError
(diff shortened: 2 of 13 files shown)


moneymanolis (approved):
LGTM!

PR #1511, opened 2021-12-14
Restore ability to import address labels of unlabeled addresses

relativisticelectron:
As described in https://github.com/cryptoadvance/specter-desktop/pull/1470#pullrequestreview-823070898  the PR https://github.com/cryptoadvance/specter-desktop/pull/1470 broke the ability to import electrum address labels (for which most of the cases no prior address label was defined).

This PR implements recommended changes of https://github.com/cryptoadvance/specter-desktop/pull/1470#pullrequestreview-823070898 .   
@moneymanolis :  If these changes are not desired for the use-case of Specter CSV  and Specter JSOn import, please comment.

### DIFF
--- a/src/cryptoadvance/specter/wallet.py
+++ b/src/cryptoadvance/specter/wallet.py
@@ -590,7 +590,7 @@ class Wallet:
         arr = [
             {"address": address, "label": label}
             for address, label in labeled_addresses.items()
-            if address in self._addresses  
+            if address in self._addresses
         ]
         logger.info(f"Array for set_labels is: {arr}")
         self._addresses.set_labels(arr)
--- a/src/cryptoadvance/specter/wallet.py
+++ b/src/cryptoadvance/specter/wallet.py
@@ -580,13 +580,16 @@ class Wallet:
                 reader = csv.DictReader(f, delimiter=dialect.delimiter)
                 reader.fieldnames = [name.lower() for name in reader.fieldnames]
                 for row in reader:
-                    labeled_addresses[row["address"]] = row["label"]
+                    if not row["label"].startswith(
+                        "Address #"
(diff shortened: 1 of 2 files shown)


moneymanolis:
First, thanks very much for your PR @relativisticelectron! I only made a slight modification of the import of a Specter CSV file to not import all the labels from such a CSV, that would probably confuse users (that was also the reason to use only labeled addresses in the first place). I think the logic is better handled now and less error prone. I also made minor changes to comments. @k9ert if relativisticelectron is OK with my modification, my review would be done.

moneymanolis (changes requested):
See my comment above.

relativisticelectron:
> First, thanks very much for your PR @relativisticelectron! I only made a slight modification of the import of a Specter CSV file to not import all the labels from such a CSV, that would probably confuse users (that was also the reason to use only labeled addresses in the first place). I think the logic is better handled now and less error prone. I also made minor changes to comments. @k9ert if relativisticelectron is OK with my modification, my review would be done.

@moneymanolis : Thanks for the changes.  To summarize; the behavior is now:

- Specter JSON and Electrum JSON will import all labels
- Labels from Specter CSV which start with "Address #" will be disregarded.  The remainder will be imported


 
Am I correct in assuming, that the Specter JSON does not contain labels starting with "Address #"?
If this is so, then this PR can be merged.

moneymanolis:
> Am I correct in assuming, that the Specter JSON does not contain labels starting with "Address #"? If this is so, then this PR can be merged.

Yes, the Specter JSON actually has a special key "labels" which has as its value a dictionary with all the labels set by the user that are not the standard "Address #" and its corresponding addresses.

moneymanolis (approved):

PR #1513, opened 2021-12-16
[do not merge] PoC of multisig support for Ledger wallets

bigspider:
Hello!
The new Ledger Bitcoin app has full multisig support, but it's not (yet?) supported in HWI.
The integration is not fully trivial because Ledger devices are stateless, so the wallet descriptor template cannot be stored into the device. Therefore, we resort to statelessly registering the "policy" on the device which returns a hmac (using a key derived from your seed), so that the policy and its hmac can be stored on the client side. You can read about it [here](https://blog.ledger.com/bitcoin-2/) or find more [technical docs of the app](https://github.com/LedgerHQ/app-bitcoin-new/tree/master/doc) if you're interested.

I wanted to see what would it take to integrate the multisig support in Specter, so I hacked together a Proof-of-Concept. It's nowhere near the quality needed to be merged, but it should be good enough to plug-and-play if you have a Nano.

#### Why 'do not merge'

Limitations of this PoC include:

- The connections done via [ledger_bitcoin](https://pypi.org/project/ledger-bitcoin/) won't work correctly if there's more than one Ledger device connected; in this PoC, I hacked the `ledger_bitcoin` library inside Specter's hwi support in inappropriate ways; a more careful design is needed there.
- Persistence of the HMAC is missing, so you'll have to re-register if you close Specter
- Only on testnet, doesn't detect the chain appropriately.
- Missing error handling, so make sure that the Nano S is unlocked and open on the Bitcoin Testnet app.
- I never used jinja before and I did _very_ bad things to convince it to pass the expected data around between JS and Python world...
- Not much though was put into the policy registration UX, which is a bit hidden.

### Walkthrough

You'll need a Nano S or Nano X with the latest firmware, and the Bitcoin Testnet application (you can install it from Ledger Live after enabling "Developer mode" in the settings) with version `>= 2.0.0`.

You'll also need to install `ledger_bitcoin` in the virtual environment where Specter is running:
```
$ pip install ledger_bitcoin[hid]
```
Then, you should be good to go.

Create a multisig wallet that includes a Ledger device, using a simple ASCII name of at most 16 characters (I used "Cold storage").
The policy registration UX is accessible in the "Settings" tab of the wallet:

0-settings-unregistered png

After clicking "Register policy", you'll have a flow on the device that looks like this:

1-register

Once completed, you should be able to verify addresses or send transactions as usual, and the device will show the wallet name on-screen:

2-show-address
3-spend

### Conclusions

Happy to hear your thoughts, and if there is anything I could do to make a production-ready integration easier. I learnt a lot from Specter while making the new Bitcoin app at Ledger, and would really love to see it finally working together!

Ideally, extending `hwilib` with support for the wallet policy registration flow would be the easiest way to bring this approach to many different software wallets; nothing about the design of registration flow is specific to Ledger, and I did my best to design a general approach that can work for any stateless hardware wallet.

### DIFF
--- a/src/cryptoadvance/specterext/devhelp/console.py
+++ b/src/cryptoadvance/specterext/devhelp/console.py
@@ -22,7 +22,7 @@ class Console:
         self.namespace.update(
             {
                 "help": f"You can try calling any of the following objects: {list(self.namespace.keys())}\n"
-                f"Use a trainling '.' to get available methods and propeties, e.g. try: '{list(self.namespace.keys())[-1]}.'\n"
+                f"Use a trailing '.' to get available methods and propeties, e.g. try: '{list(self.namespace.keys())[-1]}.'\n"
                 "The command 'run' can execute python scripts, e.g., run('myscript.py')"
             }
         )
--- a/src/cryptoadvance/specterext/devhelp/service.py
+++ b/src/cryptoadvance/specterext/devhelp/service.py
@@ -4,9 +4,11 @@ from flask import render_template
 from cryptoadvance.specter.services.service import Service, devstatus_alpha
 from cryptoadvance.specter.specter_error import SpecterError
 from cryptoadvance.specter.wallet import Wallet
+from cryptoadvance.specter import util
 from .console import Console
 import flask
 import flask_login
+from flask_login import current_user
 
 logger = logging.getLogger(__name__)
 
@@ -22,11 +24,7 @@ class DevhelpService(Service):
     devstatus = devstatus_alpha
     console = Console()
     console.updateNamespace(
-        {
-            "app": app,
-            "flask": flask,
-            "flask_login": flask_login,
-        }
+        {"app": app, "flask": flask, "flask_login": flask_login, "util": util}
     )
 
     sort_priority = 2
--- a/src/cryptoadvance/specterext/devhelp/templates/devhelp/dev-console.jinja
+++ b/src/cryptoadvance/specterext/devhelp/templates/devhelp/dev-console.jinja
@@ -5,25 +5,43 @@
 
 <div class="card" style="width:1000px">
     <p>
-        Python-expressions here:
+        
     </p>
     <form name="console" onSubmit="return handleIt()">
 
         <div style="display: flex;">
-            <span>Code:</span>
+            <span>Python-expressions here:</span>
+            <tool-tip width="500px">
+                <h4 slot="title">How to use the Console</h4>
+                <span slot="paragraph">
+                    Try these things:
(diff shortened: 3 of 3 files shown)


stepansnigirev:
Thank you very much for this PoC! It will help a lot! And I'm really excited to start using Ledger in multisig wallets - this is a very important feature for sure!
We need to decide what would be the best way to store the hmacs for registered multisig wallets on the Specter side, I think we could have some extra metadata for cosigners in the wallet, or in the device.

One question about the HMAC - is it derived from the seed loaded in the device?
If I have two Ledger devices loaded with the same recovery phrase would the HMAC be the same for both?

k9ert:
Tests are not happy because
```
E   ModuleNotFoundError: No module named 'ledger_bitcoin'
```
Would be cool if we could get that to a state where it's ready to be merged.
Check here on how to generate the `requirements.txt`:
https://github.com/cryptoadvance/specter-desktop/blob/master/docs/development.md#some-words-about-dependencies

bigspider:
Rebased and added `ledger_bitcoin` to the requirements. Didn't bother doing it before as I'm not sure you'd prefer to import the library or embed it by hand (bitcoin-core/HWI chose the second approach, for example).

Still a lot of work to do before it can be merged, though! I hope someone more familiar with specter's codebase could take it from here.

jonathancross:
> The new Ledger Bitcoin app has full multisig support, but it's not (yet?) supported in HWI.

Note:  https://github.com/bitcoin-core/HWI/commit/3fe369d0379212fae1c72729a179d133b0adc872 which includes [ledger: Support multisigs](https://github.com/bitcoin-core/HWI/commit/dffd014fedea94e420f60563af0224af7c5338b0)

bigspider:
> > The new Ledger Bitcoin app has full multisig support, but it's not (yet?) supported in HWI.
> 
> Note: [bitcoin-core/HWI@3fe369d](https://github.com/bitcoin-core/HWI/commit/3fe369d0379212fae1c72729a179d133b0adc872) which includes [ledger: Support multisigs](https://github.com/bitcoin-core/HWI/commit/dffd014fedea94e420f60563af0224af7c5338b0)

Great progress on HWI's side thanks to @achow101's work; support is still partial in the sense that wallet registration made via HWI can't be persisted (HWI registers the multisig at signing time, and there's no way of passing the hmac and wallet-policy info when signing or viewing addresses).

bigspider:
Closed for now, as it's outdated. The branch from the PoC is stored for future reference in https://github.com/bigspider/specter-desktop/tree/specter-integration-poc.

PR #1517, opened 2021-12-22
Initial Service integration: Swan [work-in-progress]

kdmukai:
See the new documentation in this branch: `docs/services/services.md`.

This PR adds:
* `Service` base structure (primarily built by @k9ert).
* `Service`-related data storage and other management mechanisms.
* An initial `Service` implementation to Swan Bitcoin.
* `Service` "reserved" receive addresses w/logo display; same for transaction history display.
* Can manually specify an `Address` as being associated with a `Service` to enable logo display and `Service`-based filtering (e.g. the Swan history page).

Known issues:
* Upon completing Swan integration, the callback that returns to Specter will require the user to log back in to Specter. I could not find a way to avoid this as the request was seen by Flask as belonging to an `AnonymousUserMixin` instead of the already-logged in `User` (presumably due to the exterior redirect).
* The Swan API `access_token` will expire after one hour. During beta we will not have a `refresh_token` which we would normally use to request a new `access_token` automatically. If your `access_token` is expired, the Swan settings page will indicate that you need to re-link your Swan account.
* TODO: Test suite needs more test cases for `Service`; some currently broken tests.

Additional UI notes:
* Begins the branding shift to the `ghost_3d.png` instead of the Pac-Man ghost.
* Adds a light hover effect on txlist and address list rows.
* NOTE the slight display changes on "Transactions" and "Addresses" tabs within a wallet.

General code notes:
* Python hint syntax via `typing` (e.g. `def foo() -> List[str]`) assumes **python 3.5+**
* More extensive python hinting added in places to enable more IDE assist for vars that it otherwise wouldn't know the type.

Detailed changes:
* `Address` now includes a `service_id` field. Services can "reserve" an unused `Address` by setting this field or historical Addresses can be identified as being associated with a `Service` via this field.
* `BaseConfig.SERVICES_DEVSTATUS_THRESHOLD` determines which `Service`s are enabled, based on their devstatus release level (e.g. alpha, beta, etc).
* Adds `Singleton` and `ConfigurableSingleton` classes. Many objects were already treated as if they were singletons so this just adds explicit support for them. Also allows us to not have to pass objs around and can instead just call `SomeSingleton.get_instance()` whenever we need it. `ConfigurableSingleton` handles classes where a one-time configuration is needed to initialize it.
* `WalletManager`: can now filter `full_txlist` and `full_addresses_info` by `Service.id` in order to reuse the `tx-table` and `address-table` web components but have them just list the `Service`-related data. Improved comments around complicated code.
* `server_endpoints/auth.py`: Adds a post-login check to each `Service.update()` to give them a chance to update balances, etc, etc. Should eventually move to a back-end process. Lots of temporary handling while Swan is in beta.
* `wallets.py` endpoints updated to pass the `services` list to the template where needed and to provide `Service`-aware data retrieval (e.g. `wallets_overview_txlist()` to get either txs combined from all wallets or filter that list to just return the combined txs from a specific `Service`). Performs some additional work in `decoderawtx()` to reduce the number of fetches from JS that web components have to make to fill out their data.
* `services.controller.py`: Very few `Service`-specific endpoints needed. Mostly just the index page to present the `Service` options.
* `Service`, `ServiceManager`, `ServiceEncryptedStorage`, `ServiceAnnotations`: see the `services.md` writeup.
* `CustomElement.js`: @k9ert addition.
* `services.swan.api.py`: Encapsulates all the back-end api calls. Goal is to keep this as simple as possible with little to no knowledge of the broader data classes. That kind of logic should all be in the `SwanService` class. It does manage its own storage vars (e.g. for `access_token`) but even then it requests access to the storage through `SwanService`.
* `services.swan.controller.py`: The few Swan UI pages are contained here. Also any user touchpoints required in the OAuth2 process (e.g. the callback page). Again, as much as possible the controller is very simple with most of the work being done in the `SwanService`.
* `SwanService`: Performs all the logic heavy-lifting. Its `sync_swan_data` needs some thorough testing as the possible scenarios get a bit complex. The important `update()` call can't be fully tested until we exit beta and have the `refresh_token`.
* `swan_client.py`: outdated artifact, to be removed from this PR.
* `swan/templates/swan/settings.jinja`: Includes temporary workaround for not having a Swan `refresh_token`.
* `swan/templates/swan/withdrawals.jinja`: TODO: display current autowithdrawal settings, if known.
* `templates/includes/services-data.html`: Loads info about each `Service` into the global JS context (required for pages that display addresses associated with a `Service` in order to retrieve the logo, etc).
* `address-data.html`: Now aware of addresses associated with a `Service`
* `address-row.html`: Displays the icon for an associated `Service`. Truncates the full address to increase horizontal real estate.
* `tx-data.html`: `Service`-aware; pre-injects more data into `<address-label>` when possible to avoid unnecessary additional JS fetches. Same for the "isMine" check.
* `tx-row.html`: `Service`-aware; more horizontal space savings. 
* `tx-table.html`: More debugging to note how many extraneous JS fetches are being made. I could not fix it, though.
* `wallets_overview.jinja`: Just tab vs space consistency fixes.
* `User`: Support for listing the `Service`s that the `User` has enabled.
* `Wallet`: Enhancements to filter `Address` and tx retrieval by `Service`.

Misc:
* `install_noded.sh`: Clarify usage info and a bugfix

TODOs
- [x] No clear error messaging when there is no authentication set up. Flash a message when trying to access without authentication for example. In doing so, briefly explain why authentication is needed (Service data is encrypted using the user's password). 
- [x] Try to better vertically align the abbreviated TX ids and addresses. 
Update: Looks like it is a browser issue. Looks good with Chrome, not so much with Firefox. Investigate what the different browser behaviour is. 
- [x] Explain address reservation scheme for Swan service (every so and so address is reserved)
- [x] Redirect to services endpoint after setting up auth or at least flash a message ("Password set successfully, go back to Choose services").
- [x] Improve the feedback at the auth endpoint (basically there is none, not "even password was set successfully" or so).
- [x] Change the abbreviation form of the addresses to the "7...7" format.
- [x] Some way to reset the address index for the receiving address endpoint (perhaps reset the address index to the latest unused and non service id tagged address?). Update: Postponed to another PR
- [x] Implement a way to reset all the address labels that are set when setting up a service ("Reserved for Swan"). Implemented when clicking the "Delete API token" button (`SwanService.unreserve_addresses()`)
- [ ] Delete API key button seems to not remove the token from disk. Swan icon still visible after deletion of the key. Also the autowithdrawal should be deactivated with pushing that button.
- [ ] Rollback-test: txlist.csv, addresses.csv and users-json might get changes. How does specter behaves if you rollback to the older version?
- [ ] Create a testing-env markdown-document
- [ ] test production
- [ ] pre-release


src/cryptoadvance/specter/server_endpoints/wallets.py
```diff
@@ -612,6 +616,12 @@ def send_new(wallet_alias):
                 specter=app.specter,
                 rand=rand,
             )
+        elif action == "fillform":
+            # can be used to recommend a transaction from a service (goind to an exchange or so)
+            addresses = request.form.getlist("addresses[]")
+            labels = request.form.getlist("labels[]")
+            amounts = request.form.getlist("amounts[]")
+            fillform = True
```
kdmukai, line 624: @k9ert, I'm unclear how this was meant to be used. Is it ever actually exposed/used in the UI or more of a future feature?

src/cryptoadvance/specter/services/static/CustomElement.js
```diff
@@ -0,0 +1,97 @@
+
+export class CustomElement extends HTMLElement {
```
kdmukai, line 2: @k9ert this doesn't seem to be used anywhere. Should this be removed from this PR?

moneymanolis:
TODOs
- [ ] No clear error messaging when there is no authentication set up. Flash a message when trying to access without authentication for example. In doing so, briefly explain why authentication is needed (Service data is encrypted using the user's password).

src/cryptoadvance/specter/addresslist.py
```diff
@@ -48,6 +50,16 @@ def set_label(self, label):
...
+    
+    def set_service_id(self, service_id: str):
+        if service_id:
+            # Declares that this Address is associated with a Service
+            self["service_id"] = service_id
+        else:
+            # Frees a reserved Address; reset its label
+            self["service_id"] = None
+            if not self.used and self["label"]:
+                self.set_label("")
```
moneymanolis, line 62: Wouldn't that mean that the address property `is_labeled` is still set to true when using an empty string? An empty string would also not be the "default label" that the address had before (usually sth. like "Address #32").

src/cryptoadvance/specter/config.py
```diff
@@ -135,16 +135,25 @@ class BaseConfig(object):
...
+
+    # EXPLAIN_TEMPLATE_LOADING = os.getenv("EXPLAIN_TEMPLATE_LOADING", "False")
+
     SPECTER_DATA_FOLDER = os.path.expanduser(
         os.getenv("SPECTER_DATA_FOLDER", "~/.specter_dev")
     )
     # API active by default in dev-mode
     SPECTER_API_ACTIVE = _get_bool_env_var("SPECTER_API_ACTIVE", "True")
 
+    # One of "prod", "beta" or "alpha". Every Service below will be not available
```
moneymanolis, line 154: When / how can "alpha" services ever be used if it is not possible in the Development Config?

src/cryptoadvance/specter/addresslist.py
kdmukai: Python bools treat empty string and empty lists as `False`:

```
>>> test = ""
>>> bool(test)
False

>>> if test:
...     print("hello!")
... else:
...     print("nope!")
... 
nope!
```

src/cryptoadvance/specter/specter.py
```diff
@@ -428,7 +441,11 @@ def update_alt_rate(self, alt_rate, user):
     def update_alt_symbol(self, alt_symbol, user):
         self.config_manager.update_alt_symbol(alt_symbol, user)
 
-    # mark logic!
+    def update_services(self, services):
+        """takes a list of service_names which should be activated"""
+        self.config["services"] = services
+        self._save()
+
```
k9ert, line 448: This is no longer needed as with this commit, services are dependent on the user:
https://github.com/k9ert/specter-desktop/commit/6d64a2964f9a7cc3d0b36e5b0d455fc7b6569430#diff-3a0c72fdfb0c27876d7c5efc052274c635e69d022508f8bf2f93c56ceaf7a9de
(However this code did not get deleted)

src/cryptoadvance/specter/services/service_manager.py
```diff
@@ -0,0 +1,163 @@
...
+        return [self._services[s] for s in service_names]
+
+    def set_active_services(self, service_names_active):
+        logger.debug(f"Setting these services active: {service_names_active}")
+        self.specter.update_services(service_names_active)
+        for _, service in self.services.items():
+            logger.debug(
+                f"Setting service '{service.id}' active to {service.id in service_names_active}"
+            )
+            service.active = service.id in service_names_active
```
k9ert, line 157: I think this is no longer needed. The call of this in the service_endpoint is also no longer needed.

k9ert:
can we rename `src/cryptoadvance/specter/services/swan/api.py` to swan_client.py ? I think that's a better term. You can delete the file which currently has that name.

src/cryptoadvance/specter/services/swan/api.py
```diff
@@ -0,0 +1,400 @@
...
+        if method == "GET":
+            response = requests.get(api_url + endpoint, headers=auth_header)
+        elif method in ["POST", "PATCH", "DELETE"]:
+            response = requests.request(
+                method=method,
+                url=api_url + endpoint,
+                headers=auth_header,
+                json=json_payload,
+            )
+        print(response)
```
k9ert, line 173: Please remove or logger

k9ert (changes requested):
Awesome work! Congratulations. Looking forward to bring this to live.

docs/archblog.md
```diff
@@ -1,7 +1,14 @@
 # Purpose
 Let's find a place to docoment non straightforward design decisions.
 
-## 02nd Oct 2020 - k9ert
+## 09nd Nov 2021 - k9ertYou
+The service-integrations should somehow scale and it should be easy to add more services. Let's have a look on how they are implemented.
+On the data side of the topic, there is a key in the `config.json` callled `services`. It contains a map of services where each one is described with `name`, `title`, `logo` (link to an image) and a `desc`. They are accessed and (de-)activated via the service_manager. That can be managed in the settings by the admin. Each service is either activated for all users or not.
```
moneymanolis, line 6: Services key is not created during testing (in my case at least).

kdmukai:
>     * [ ]  Some way to reset the address index for the receiving address endpoint (perhaps reset the address index to the latest unused and non service id tagged address?).

@moneymanolis I would leave this last one as a TODO in a separate PR.

src/cryptoadvance/specter/server_endpoints/auth.py
```diff
@@ -196,6 +198,20 @@ def redirect_login(request):
         response = redirect(request.form["next"])
     else:
         response = redirect(url_for("index"))
+    
+    # Check the user's Service integrations
+    # TODO: Can these be done in a background process so we don't slow down the user experience?
```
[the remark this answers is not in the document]
kdmukai: Agreed, too much work to go in this PR.

src/cryptoadvance/specter/addresslist.py
```diff
@@ -179,10 +199,16 @@ def add(self, arr, check_rpc=False):
                 )
         self.save()
 
-    def set_label(self, address, label):
+    def get_address(self, address: str, add_if_missing: bool = True) -> Address:
         if address not in self:
```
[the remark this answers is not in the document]
kdmukai: This `add_if_missing` bool shouldn't actually be needed in the one place where it's called with this method. Removing it.

src/cryptoadvance/specter/server_endpoints/auth.py
```diff
@@ -196,6 +198,20 @@ def redirect_login(request):
...
         response = redirect(url_for("index"))
+    
+    # Check the user's Service integrations
+    # TODO: Can these be done in a background process so we don't slow down the user experience?    
+    from cryptoadvance.specter.services.swan.api import SwanApiRefreshTokenException  # TODO: Remove this import after Swan exits Beta testing
+    for service_id in app.specter.user_manager.get_user().services:
+        try:
+            service_cls = app.specter.service_manager.get_service(service_id)
+            service_cls.update()
+        except SwanApiRefreshTokenException as e:  # TODO: Remove after Swan exits Beta
```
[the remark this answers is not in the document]
kdmukai: @k9ert This whole section will be removed soon (days?). And then when we implement a backend update process, we'll have to figure out how it should handle exceptions and relay messages to the user. So basically this is just a very temporary bandage with better work to come.

src/cryptoadvance/specter/services/swan/service.py
```diff
@@ -0,0 +1,223 @@
...
+        # Did we at least get a Swan walletId that we can update later?
+        if swan_wallet_id:
+            cls.update_current_user_service_data({
+                SwanService.SWAN_WALLET_ID: swan_wallet_id,
+            })
+
+
+
+    @classmethod
+    def update(cls):
```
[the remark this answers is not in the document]
kdmukai: Parent `Service` class does already define an empty placeholder. The temporary call in `redirect_login` is only there until a backend process is ready to take that task over.

As for the name, it shouldn't be `auth_` prefixed since the current `update()` call doesn't belong in there in the first place (again, it's only there because we don't have the backend process to call it).

I don't know that I'd call it a `callback`. Some process (the login flow for now, a backend process soon) has to tell each `Service` to update the state of its data (either remote 3rd party data or detect new local changes like a newly received utxo to an addr reserved for that `Service`). So `update` makes sense to me. But perhaps `sync` or `resync` if `update` really is the wrong name?

To be clear, I'm assuming the upcoming backend process would just do something like:
```
for service_cls in self.services:
    service_cls.update()
```
...repeatedly on some periodic basis.

src/cryptoadvance/specter/services/swan/controller.py
```diff
@@ -0,0 +1,220 @@
...
+        user = app.specter.user_manager.get_user()
+        if SwanService.id not in user.services:
+            user.services.append(SwanService.id)
+
+        # Sync Specter with any previous Swan-Specter integrations
+        # TODO: This still needs more testing; it's irrelevant for first-time Swan authorizations anyway
+        # SwanService.sync_swan_data()
+
+        return redirect(url_for(".oauth2_success"))
+    except Exception as e:
```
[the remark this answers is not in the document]
kdmukai: This particular one is encapsulating a lot. It's doing the api callback to exchange the `authorization_code` for the `access_token` and then calling the complex `sync_swan_data` (which still needs a lot of testing). Would probably make sense to at least break it up into two separate `try/except` blocks rather than stacking `Exception` handlers.

moneymanolis:
Yeah, last TODO doesn't have to be in this PR.

src/cryptoadvance/specter/config.py
k9ert: Yes, good point. Should be "alpha" in the DevelopmentConfig.

src/cryptoadvance/specter/services/swan/service.py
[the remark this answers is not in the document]
k9ert: From a framework-perspective, it doesn't matter what the service is doing it's called regulary through some look like the one you described above. That's why i proposed a naming-scheme for such methods which rather describes when they got called, not what they are doing. 
If a service wants to update something, it can still define a  `_update()` method and internally call that in there function:
```
for service_cls in self.services:
    if hasAttr(service,"cron_callback_daily"): # Would not need the empty placeholder
        service_cls.cron_callback_daily()
```
And then in that method, one would call update:
```
    def cron_callback_daily():
        _update()
```

Does that make sense?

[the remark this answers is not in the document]
k9ert: Let's use hooknames like those:
* hooks called: on_user_login
* cron hooks like on_schedule_daily

k9ert (approved):

PR #1522, opened 2022-01-04
Bump HWI to 2.0.2

### LINKED ISSUE #1219: Keep Key issue
I have been able to set up both my ledger and my trezor, but when I try to add the keep key I get all the way to the dialog that asks for the password (BIP39).  I enter it, and at this point I would expect it would appear on the Keep Key after hitting submit but it does not. And I get this error in the screenshot:

image

It SEEMS as if it skips over the part where it sends to the keep key and waits for a response.  It does handle the PIN correctly.

I am using BOTH the latest firmware on the keep key, as well as the latest specter I think
Bootloader 1.1.0
Firmware 7.1.2
Specter:1.4.2 

(What a great project this is, by the way... VERY well done.)

aphex3k:
This can potentially fix #1219, see: https://github.com/bitcoin-core/HWI/issues/526

- [x] Correct the hash in the requirements.txt


k9ert:
Thanks for your PR, looks good to me. 
Release seems to be mostly bugfixes.
https://github.com/bitcoin-core/HWI/releases/tag/2.0.2
Will merge soon when @kdmukai does not complain ...

PR #1526, opened 2022-01-13
Documentation: Create document for testing-setups and cleaning up swan

k9ert:


src/cryptoadvance/specter/services/swan/service.py
```diff
@@ -61,9 +62,15 @@ def get_associated_wallet(cls) -> Wallet:
...
-            service_data[cls.SPECTER_WALLET_ALIAS]
-        )
+        try:
+            return app.specter.wallet_manager.get_by_alias(
+                service_data[cls.SPECTER_WALLET_ALIAS]
+            )
+        except SpecterError as e:
+            logger.debug(e)
+            # Referenced an unknown wallet
+            # TODO: keep ignoring or remove the unknown wallet from service_data?
```
k9ert, line 72: I think it's ok to ignore but then we should make sure that the SpecterError is really about that. I'm not sure how to do that best other than (currently) simply comparing Exception-texts (or better `.startswith`/`endswith`). 
Potentially we could also do subClasses of SpecterError but let's not think about that here.

I will also merge this PR independent from this being fixed. So just an opinion ...

k9ert:
I reviewed and it LGTM. Will merge now and @moneymanolis can review the markdown document by opening another PR.

PR #1533, opened 2022-01-21
Final Swan service updates

kdmukai:
* Production Swan service api credentials
* Remove most debugging statements in the api calls
* Bugfix on `User._services` left nav
* Promote Swan service to "prod" status
* UI display formatting of autowithdrawal setting
* Possible bugfix for Liquid address details
* Moving `ServiceManager` Instantiation out of Specter.__init__()
* Loading for external extensions in DevelopmentConfig only
* build-script adjustments for clarity and right order
* monkey patch pyinstaller rthook in order to workaround https://github.com/pyinstaller/pyinstaller/issues/6537

### DIFF
--- a/docs/endless-pacman.md
+++ b/docs/endless-pacman.md
@@ -1,0 +1,15 @@
+Sometimes people have issues where they get an endless Pacman animation and the application is not coming up. If you suffer from this, here are some hints on how to deal with that.
+
+# Check The Logs
+The logs are usually in the `.specter` subfolder of your homediretory. There, you might find a file called `specter.log` and/or specterApp.log. If you're running Specter as a binary application (in contrast to a pip installation) which most people do that `specterApp.log` is the relevant file for you. This file might contain content which gives a hint on what's wrong. If you can't find anything suspicious, feel free to create a [pastebin](https://pastebin.com/) and ask in the chat for help (with a link to the created pastebin).
+
+# Check Port 25441
+Maybe there is another instance (still) running. Check that via opening your brower here: [http://localhost:25441](http://localhost:25441)
+If that's the case, the most easy solution is to reboot your computer.
+
+# Check Whether the Binary is Existing
+The first thing Specter is doing if you start up the app is downloading the correct specterd from the GitHub-release page and storing that executable in the `Homefolder/.specter/specterd-binaries` subfolder. You should find a file called `specterd`.
+If the file is there but you still get the endless Pacman, try one of the following things:
+* delete the file so it'll get redownloaded
+* download it from the GitHub-release page and place it manually there, especially if you're not properly connected to the internet
+* start the specterd manually via the command line
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -19,6 +19,7 @@ nav:
       - hwibridge.md
       - elements.md
       - windows.md
+      - 'Endless Pacman Troubleshooting': endless-pacman.md
     - Development: 
       - 'Contributing': contributing.md
       - 'Developing': development.md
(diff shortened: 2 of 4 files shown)


src/cryptoadvance/specter/services/swan/client.py
```diff
@@ -284,13 +266,10 @@ def update_autowithdrawal_addresses(
                         "clientId": "specter-dev"
                     },
                     "specter_wallet_alias": "seedsigner_demo"
-                },
```
k9ert, line 287: `"btcAddresses": []` seems to be disappearing all over. I don't understand why?

src/cryptoadvance/specter/services/swan/config.py
```diff
@@ -1,6 +1,14 @@
...
+    SWAN_CLIENT_ID = "specter-dev"
+    SWAN_CLIENT_SECRET = (
+        "BcetcVcmueWf5P3UPJnHhCBMQ49p38fhzYwM7t3DJGzsXSjm89dDR5URE46SY69j"
+    )
     SWAN_API_URL = "https://dev-api.swanbitcoin.com"
 
 
 class ProductionConfig(BaseConfig):
+    SWAN_CLIENT_ID = "specter"
+    SWAN_CLIENT_SECRET = (
```
k9ert, line 11: We should make a remark here for semi-educated security-researchers. Like:
```suggestion
    # Conceptionally, we're using PKCE which should not make any password necessary. However, technically, it
    # was too much challenge to avoid the password and so we technically this password is used but it's not 
    # considered to be a real secret. There is no reasonable attack-vector for this secret being public.
    # Obviously, you're happy to prove wrong but please don't complain about hard-coded secrets or so.
    SWAN_CLIENT_SECRET = (
```

src/cryptoadvance/specter/user.py
```diff
@@ -308,8 +308,8 @@ def save_info(self, delete=False):
 
     def add_service(self, service_id: str, autosave: bool = True):
         """Add a Service to the User. Only updates what is listed in the sidebar."""
-        if service_id not in self.services:
-            self.services.append(service_id)
+        if service_id not in self._services:
+            self._services.append(service_id)
```
k9ert, line 312: Why is this necessary? Uh, oh, now i see the issue. This is necessary because of my bad implementation of the property:
```
    @property
    def services(self):
        if self._services:
            return self._services
        return []
```
It should be changed to:
```
    @property
    def services(self):
        if not self._services:
            self._services = []    
        return self._services
```
Do you agree?

src/cryptoadvance/specter/services/swan/client.py
kdmukai: Pablo changed the return data at some point so we no longer get the list of addresses associated with a Swan wallet. We don't need it anyway, as we can always just overwrite the current with a new autowithdrawal call.

src/cryptoadvance/specter/user.py
kdmukai: Either way the `User._services` seems a little inconsistent in that it has a "member"-like level of protection in the class but that isolation isn't really accomplishing anything; external code can directly manipulate the `user_obj.services` prop just as easily as if the var was declared as `User.services` without the prop.

But for the `add_service` method itself, it's within `User` which should be free to modify its own internal `self._services` "private" member whenever/wherever it needs to.

I think this all boils down to philosophy/semantics. Bottom line for me is that the left nav is showing the active services once again!

src/cryptoadvance/specter/services/swan/config.py
kdmukai: Good call. Comment added at the top of the file.

src/cryptoadvance/specter/user.py
kdmukai: Applied your patch to the `@property`.

k9ert: Yes, it's not java. The reason i converted it to a property was avoiding a "has not attribute _services" issue. However, not implemented well. It doesn't matter to me that much whether to use internal or external (even though if it 's possible to use external, it's better in order to avoid the above issue) but the main thing was that this revealed my bad naive implementation.
Good that it's fixed!

PR #1534, opened 2022-01-24
fix service icons breaking liquid

stepansnigirev:


k9ert:
https://cirrus-ci.com/task/6043788106792960?logs=cypress#L951

```
     ReferenceError: The following error originated from your application code, not from Cypress.
  > services is not defined
When Cypress detects uncaught errors originating from your application it will automatically fail the current test.
This behavior is configurable, and you can choose to turn this off by listening to the `uncaught:exception` event.
https://on.cypress.io/uncaught-exception-from-application
      at HTMLElement.connectedCallback (http://localhost:25444/wallets/wallet/testwallet_ghost/receive/?newwallet=true:1795:13)
```

I guess this might fix it? https://stackoverflow.com/questions/5113374/javascript-check-if-variable-exists-is-defined-initialized

stepansnigirev:
fixed

Issue #1480, opened 2021-11-16
Cannot import keys from Jade HWW over USB


Lobbelt:
I updated both Specter Desktop (1.7.1) and the Jade firmware (01.30 no radio) to the latest version. When trying to import the keys into Specter Desktop, the screen is stuck at "Detecting...", even though the device is plugged in. Retried multiple times in different USB ports as well. Blockstream Green is perfectly detecting (and unlocking) the device.

It seems to be the same issue referenced in #1251.

ikohut:
Same for me. Specter Desktop (1.7.1) and the Jade firmware (0.1.30) too. No luck to connect to Specter, but Blockstream Green app works well

k9ert:
The thing is that i couldn't reproduce the issue. So more details are needed. Which OS are you using? WHich version of the jade-firmware? If you're using Windows, did you do this:
https://help.blockstream.com/hc/en-us/articles/900005443223-Nothing-happens-when-I-plug-my-Blockstream-Jade-to-my-computer

Lobbelt:
My OS is MacOS Big Sur 11.2. Jade firmware details I mentioned above: 01.30 no radio. 

Blockstream Green detects the Jade so no issue there...

evliu:
Same on macos Monterey. Blockstream Green has no issues, I see it in the system profiler under USB, latest Specter and Jade firmware

CoderBoogs:
I am also having the same issue. When I connect the Jade it stays stuck on detecting

k9ert:
I have only a MacOS catalina-machine. I just tried and it's working there. It's the same machine which is building the App-image so it could potentially be an issue with the AppImage optimized for Catalina and not working on later MacOS versions.

So are you all working with the AppImage or is there someone who can confirm that it's (not) working in a pip-installation on a newer MacOS ?

evliu:
I used the AppImage on my Catalina machine and have the same issue as my two Monterey machines. Both Monterey machines are m1, Catalina is Intel. Issue still persists for me.

CoderBoogs:
I am not sure I am using windows

stepansnigirev:
Jade added a new VID-PID pair to the list of USB devices, so it might be that devices that use these PID:VID were not detected by Specter.
We updated the list in https://github.com/cryptoadvance/specter-desktop/pull/1520 so next release should be able to detect all Jades.

Lobbelt:
[...]

PR #1535, opened 2022-01-24
Add suggestion to quit other wallet software when connecting HWW

Lobbelt:
Referencing https://github.com/cryptoadvance/specter-desktop/issues/1480. Tested this with a Blockstream Jade and BitBox 02. If resp. Blockstream Green or the BitBox app is open, Specter Desktop cannot detect the HWW connected over USB as I assume the other wallet software 'monopolises' the connection (there is undoubtedly a good tech term for it that I don't know). When the other wallet software is closed, Specter Desktop can again find the HWW and extract its keys.

I would therefore propose to suggest trying this to the user as well in the message that shows up if the device is not found immediately.


k9ert:
Thanks !

PR #1538, opened 2022-01-26
Bug fix: Multiple send bug fixes

moneymanolis:
- Multiple send bug fixes 
- Disabling RBF and subtract fee from for Liquid 
- Restoring Cypress tests


stepansnigirev:
Subtract from amount is not visible on Liquid.
image

stepansnigirev:
If I make two outputs, on the first output I select to send max of my LBTC,
THEN in the second output I click on Send Max of my assets,
I get an error "insufficient funds"
I think it's because of some bug in subtract from amount logic - I assume when I click on send max on the asset output, the subtract checkbox gets unchecked on the first checkbox that send LBTC.

src/cryptoadvance/specter/templates/wallet/send/new/wallet_send.jinja
```diff
@@ -156,6 +158,24 @@
 		})
 		var FeeSelectionComponent = document.getElementById('fee-selection-component')
 		FeeSelectionComponent = document.getElementById('fee-selection-component')
+
+		// Switch off RBF and subtract fee for Liquid
+		function adjustForLiquid() {
+			if (NETWORK == "Liquid") {
+			console.log("On Liquid, removing RBF and subtract from features ... ")
```
stepansnigirev, line 165: ... ?

src/cryptoadvance/specter/templates/wallet/send/new/wallet_send.jinja
```diff
@@ -372,15 +392,9 @@
...
-				showError(`{{ _("Please first enter a valid address") }}`);
 				return;
 			}
-			let subtract = document.getElementById('subtract');
-			subtract.checked = (units[i] == 'sat' || units[i] == 'btc');
-			if (amounts.length > 1) {
-				document.getElementById('subtract_from').style.display = 'block';
-				document.getElementById('subtract_from_input').value = i + 1;
-			}
+			FeeSelectionComponent.setSubtract(units[i] == 'sat' || units[i] == 'btc' || units[i] == 'lbtc');
```
stepansnigirev, line 397: As far as I remember LBTC has unit `btc` or `sat`, so it will never be `lbtc`

k9ert:
Additional refinements will be resolved in further PRs.

PR #1542, opened 2022-01-28
Fixed edge case for electrum label import

relativisticelectron:
Fixed edge case, in which an electrum address label is an empty string and included in the electrum_labels.json.  Then the empty  address label should not supercede the tx label.

This can happen if the address label was removed in electrum. However electrum could still have a tx label (and shows this also in the UI).

### DIFF
--- a/src/cryptoadvance/specter/wallet.py
+++ b/src/cryptoadvance/specter/wallet.py
@@ -567,7 +567,7 @@ class Wallet:
                         else txitem["address"]
                     )
                     for one_address in address_list:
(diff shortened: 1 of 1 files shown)


k9ert:
Can't think of a better timing :-D. Thank you !

moneymanolis:
Just tested it, works fine! Thanks @relativisticelectron for this!

relativisticelectron:
Thanks for testing and merging.   I only changed 1 line of code ;-)  The rest was comments.

Issue #1544, opened 2022-01-29
Importing Multisignature transaction from Electrum Fails in Specter


relativisticelectron:
Using a 2/2 testnet multisig electrum wallet: I created a PSBT with 
a) electrum 4.1.5 (partially signed)
b) specter   (unsigned)

Importing the electrum PSBT fails ([error message.txt](https://github.com/cryptoadvance/specter-desktop/files/7964754/error.message.txt)), while (re) importing the specter PSBT succeeds. 

After decoding the PSBTs I find:
- the specter PSBT has `"inputs": [    {      "witness_utxo"`, which is missing in the electrum PSBT

Is this a bug in specter or is the electrum PSBT missing a critical piece?  

This issue is a possible duplicate of: https://github.com/cryptoadvance/specter-desktop/issues/104

pleblira:
I think I got a similar error here while importing a partially signed transaction from Electrum - 3o2 signature already signed by one wallet.

`Traceback (most recent call last):
  File "flask/app.py", line 1950, in full_dispatch_request
  File "flask/app.py", line 1936, in dispatch_request
  File "flask_login/utils.py", line 269, in decorated_view
  File "cryptoadvance/specter/server_endpoints/wallets.py", line 686, in send_pending
  File "flask/templating.py", line 137, in render_template
  File "flask/templating.py", line 120, in _render
  File "jinja2/environment.py", line 1090, in render
  File "jinja2/environment.py", line 832, in handle_exception
  File "jinja2/_compat.py", line 28, in reraise
  File "/var/folders/cl/nxrkb1hx6yj99pgk8c3flyf40000gn/T/_MEIcLmxcD/templates/wallet/send/pending/wallet_sendpending.jinja", line 2, in top-level template code
    {% set tab = 'send' %}
  File "/var/folders/cl/nxrkb1hx6yj99pgk8c3flyf40000gn/T/_MEIcLmxcD/templates/wallet/components/wallet_tab.jinja", line 1, in top-level template code
    {% extends "base.jinja" %}
  File "/var/folders/cl/nxrkb1hx6yj99pgk8c3flyf40000gn/T/_MEIcLmxcD/templates/base.jinja", line 98, in top-level template code
    {% block main %}
  File "/var/folders/cl/nxrkb1hx6yj99pgk8c3flyf40000gn/T/_MEIcLmxcD/templates/wallet/components/wallet_tab.jinja", line 12, in block "main"
    {% block content %}
[...]

PR #1548, opened 2022-02-01
Fix Electrum PSBT import

relativisticelectron:
This PR is a fix for the following problem:

Importing an Electrum (multisig) PSBT failed. The Electrum PSBT, which contains only a `non_witness_utxo` field and not a `witness_utxo`.  The cause of this was that the json crashed here
https://github.com/cryptoadvance/specter-desktop/blob/eaeac1f20482dbb25ef6867a1a5414df9fae766a/src/cryptoadvance/specter/persistence.py#L77
because the field  ` 'non_witness_utxo': <cryptoadvance.specter.util.psbt.SpecterTx object at 0x7fbfbc3aebe0>` is a python class, which json doesn't know how to convert.  This was called in here
https://github.com/cryptoadvance/specter-desktop/blob/9d3a39461360e2238c654cd9a353cdeb99e44a64/src/cryptoadvance/specter/wallet.py#L887
which should have with `self.to_json()` ensured that all is json compatible, but the following statement was not converted to a dict:
https://github.com/cryptoadvance/specter-desktop/blob/eaeac1f20482dbb25ef6867a1a5414df9fae766a/src/cryptoadvance/specter/util/psbt.py#L201

Original Issue: https://github.com/cryptoadvance/specter-desktop/issues/1544


src/cryptoadvance/specter/util/psbt.py
```diff
@@ -76,6 +76,8 @@ def vout_to_dict(self, vout: TransactionOutput) -> dict:
         return obj
 
     def to_dict(self) -> dict:
+        if not self.tx:
+            return
```
k9ert, line 80: Isn't it guaranteed that it should return a dict? So why not:
```suggestion
            return {}
```

relativisticelectron: Yes, I think your way is best.

k9ert:
Thank you!

PR #1561, opened 2022-02-05
introduction of url-prefixes and defaults

k9ert:
With the recent move to an extension-concept, security challenges arised. Extensions might be as insecure as we don't trust them to share a session-cookie with them. The main focus of this PR is the concept of a `isolated_client` extension. Those are mounted below `/ext` whereas specter is mounted below `/spc`. Normal extensions are (by default) mounted below `/spc/ext` . In combination with setting the cookie-path to `/spc`, this effectively locks out isolated_client-extensions from the session-cookie of specter.

So here is how this can be configured:
* By default each extension is `isolated_client` `True`. Switch it to false of you want to integrate with specter on a session-basis
* `SPECTER_URL_PREFIX = "/spc"` is set by default. If you want to have the behaviour as before, set it to`""`
*  `EXT_URL_PREFIX = "/spc/ext"` is the default "mountpoint" for normal extensions. As `SESSION_COOKIE_PATH = SPECTER_URL_PREFIX`, the normal extensions share a session (-cookie) with specter. 
* `ISOLATED_CLIENT_EXT_URL_PREFIX = "/ext"` is the default "mountpoint" of isolated_client-extensions which do not share a session with specter, hence the name.
* If you want to mount "everything" somewhere else than `/`, you can set `APP_URL_PREFIX`. However, in that case it's your responsibility to make sure no one ends up at the `/`.


k9ert:
@Kexkey might be interesting for you. Specter is still used in cyphernode, no?

moneymanolis (approved):
LGTM

PR #1567, opened 2022-02-09
Feature: Message signing via QR code

moneymanolis:
# Flow
You ideally come from a wallet and a specific address, but you could also input the derivation path by hand directly from the devices setting where the first approach is redirecting to.

# Visuals

https://user-images.githubusercontent.com/70536101/153229011-e5f7d44f-78fd-4722-8abd-02afaa31e401.mov

- Would currently only work with Specter DIY (if in the end, this still holds true, we'd need to condition the functionality on DIY devices
- Reached out to Keystone and Seed Signer
- How the signature is processed in Specter Desktop (another pop up or change of the last pop up to show the signature etc.) still needs to be implemented
- There also has to be a little change to this DIY code snippet:
https://github.com/cryptoadvance/specter-diy/blob/c91a36dfd9ebabf5b22b6835b40beab9b8fd6fec/src/apps/signmessage/signmessage.py#L77
# Known issues
- Refactoring of `get_descriptor()` currently breaks stuff with `displayAddressOnDevice()`
- Cypress tests not working due to an EventListener in `device.jinja`

# TODOs
- [x] Fix issues and tests
- [x] Implement signature handling / displaying in Specter Desktop
- [x] Adjust DIY code
- [x] Polish UIUX
- [x] Check USB message signing flow (anything broken, any easy UX improvements?)


src/cryptoadvance/specter/server_endpoints/wallets_api.py
```diff
@@ -425,15 +427,26 @@ def addressinfo(wallet_alias):
         wallet = app.specter.wallet_manager.get_by_alias(wallet_alias)
         address = request.form.get("address", "")
         if address:
-            descriptor = wallet.get_descriptor(address=address)
```
moneymanolis, line 428: In the old version this descriptor was a class: `<class 'embit.descriptor.descriptor.Descriptor'>`

src/cryptoadvance/specter/server_endpoints/wallets_api.py
```diff
@@ -425,15 +427,26 @@ def addressinfo(wallet_alias):
...
+            descriptor = add_checksum(
```
moneymanolis, line 430: In the changed version we are generating two descriptors as strings at the api endpoint (and not as before in a confusing way with JS objects in the jinja template): one with pubkey and one with an xpub.

moneymanolis:
# Update
Commit: https://github.com/cryptoadvance/specter-desktop/pull/1567/commits/e0dd454ba7a9bc877d097a2a497a2d1a50cbfc39:
- Message signing button better integrated in address info 
- Restriction to single sig wallets and (for now) DIY devices
- UI flow added to get signature from DIY QR code back into Specter

## Current visuals
https://user-images.githubusercontent.com/70536101/167807068-8c3d4a36-f91e-42bf-9a22-4dea16890398.mov
https://user-images.githubusercontent.com/70536101/167807079-a77215e2-306c-4129-b50c-fb7c1523c872.mov

## TODOs
- [ ] If UIUX is aggreed upon, add Cypress test for the signing flow
- [ ] Adjust DIY code
- [ ] Check USB message signing flow (anything broken, any easy UX improvements?)
- [ ] Fix issues and tests

b30wulffz:
image
I don't think that it's a good idea to give device name within the button, as it will vary the widths for different names. 
Have you handled the case, when device name causes the text to wrap? Basically how will the icon and the text aligns.

moneymanolis:
> image I don't think that it's a good idea to give device name within the button, as it will vary the widths for different names. Have you handled the case, when device name causes the text to wrap? Basically how will the icon and the text aligns.

@b30wulffz, first off, the device name is pretty useful here IMO in the UX flow to point the user in the right direction (that he will sign with the device associated with the wallet, so he should also not be surprised to end up on the respective device endpoint). In short: I'd leave it like this. 

You have a valid point, though, we need to make sure that longer device names don't mess up the UI. 

I think there are two options here:
1) Using` width: min-content `for the button (and only for the signing button), would look like this:
image
2) Using a `max-width `on the button and making the text a div and using `word-wrap: anywhere`, would look like this:
image

What do you think?

src/cryptoadvance/specter/server_endpoints/wallets_api.py
```diff
@@ -489,15 +491,35 @@ def addressinfo(wallet_alias):
...
+            # The last two regex groups are optional since Electrum's derivation path is shorter
+            derivation_path_pattern = (
+                r"(\/[0-9h]+)(\/[0-9h]+)(\/[0-9h]+)(\/[0-9h]+)?(\/[0-9h]+)?"
+            )
+            # Only "descriptor" gives full derivation path, looks usually like this:
+            # wpkh([8c24a510/84h/1h/0h/0/0]0331edcb16cfd ... e02552539d984)#35zjhlhm
+            match = re.search(derivation_path_pattern, descriptor)
+            if not match:
+                logger.debug(
+                    "Derivation path could not be parsed. Sth. wrong with the regex?"
```
k9ert, line 509: Always pass as much data into a debug/error message as possbile. This message is not helpful if someone is reading it. At best, pass in descriptor AND pattern.

k9ert (approved):
Great work! Works nicely.

src/cryptoadvance/specter/server_endpoints/wallets_api.py
moneymanolis: Changed as requested in https://github.com/cryptoadvance/specter-desktop/pull/1567/commits/94f6bef238d516b886a3e8bd8ef0fd5a34b41c18

k9ert (approved):

PR #1579, opened 2022-02-13
OAuth2_hostname

k9ert:
The swan integration only works with hostnames which are in the whitelist for the callback-url.
This PR ensures that the user is using one of the supported hostnames.


src/cryptoadvance/specter/services/swan/client.py
```diff
@@ -54,7 +54,7 @@ def get_oauth2_start_url():
     flow_url = f"{api_url}/oidc/auth?"
     query_params = [
         f"client_id={client_id}",
-        "redirect_uri=http://localhost:25441/svc/swan/oauth2/callback",  # TODO: Will localhost work in all usage contexts (e.g. standalone app)?
+        f"redirect_uri=http://{ callback_hostname }/svc/swan/oauth2/callback",  # TODO: Will localhost work in all usage contexts (e.g. standalone app)?
```
k9ert, line 57: remove comment!

src/cryptoadvance/specter/services/swan/controller.py
```diff
@@ -54,8 +55,15 @@ def index():
...
         # User has already completed Swan integration; skip ahead
         return redirect(url_for(f"{SwanService.get_blueprint_name()}.withdrawals"))
+    specter_used_hostname = urlparse(request.url).netloc
     return render_template(
-        "swan/index.jinja", swan_frontend_url=app.config["SWAN_FRONTEND_URL"]
+        "swan/index.jinja",
+        swan_frontend_url=app.config["SWAN_FRONTEND_URL"],
+        specter_used_hostname=specter_used_hostname,
+        # The oauth2-flow uses a whitelist. Is our hostname whitelisted?!
+        specter_hostname_supported=specter_used_hostname
```
k9ert, line 64:
```suggestion
        specter_used_hostname = specter_used_hostname
```

moneymanolis (approved):
Concept ACK

k9ert:
This is now working on my dev-environment and `https://dev-api.swanbitcoin.com`. @kdmukai  can you check on prod?

kdmukai:
Initial link out to Swan is failing on production, citing `uri_redirect_mismatch` with http://localhost:25441/...

k9ert:
> Initial link out to Swan is failing on production, citing uri_redirect_mismatch

Damn. This is pretty nasty. We have now the following situation:
* On prod, only one `redirect_uri` is configured. This limitates umbrel-users.
* If you configure a second `recirect_uri`, you'll need to add the `redirect_uri` to the call when you get the access-token. This means that `v1.8.1` is no longer compatible with two urls.
* However, the current version since `/spc` is also not compatible with the `redirect_uri` on prod.

Fortunately, we can remove the `/spc` change just by configuration. So as a result, we need to do the following:
* Remove the `/spc/ change by configuration
* Add the `redirect_uri` (as in this PR) to the access-token call
* Release `v1.9.0` and push people to migrate, wait some time so that people can migrate
* Add a whole bunch of new `redirect_uri` on production
* Remove the `/spc` config-change again

jimbrend:
hi the plugin redirect is currently broken, could you point me in the right direction @k9ert I would like to understand the Swan auto-withdrawal plugin and learn to debug if possible.

From my understanding OAuth2 resolves in the local host browser with varying URLs such as connected to your own node, depending on the user, and this causes a problem to resolve when communicating with websites, maybe there are a few tweaks and UX could be improved if this is a good part of the user flow to link Specter and withdrawals if a new or existing user starts this from within Specter app too... I would also like to help debug on the umbrelOS side.

Is this the current problem redirect_uri related here https://github.com/cryptoadvance/specter-desktop/blob/0ac4640a3fcbbb9fd734d01b6eb9cffa5ad9634c/src/cryptoadvance/specter/server_endpoints/controller.py#L38 

that the lines 38-59 there must be changed somehow and submitted as a new PR for Specter to account for that?  Thank you

moneymanolis:
@jimbrend I think the Swan plugin is also not supported from the Swan BE anymore. We probably need to just remove the plugin alltogether or put it to alpha @k9ert.

jimbrend:
@nmfretz ping for vis on the local host redirect issue as related to self-hosters on connecting here, ty!!

PR #1580, opened 2022-02-14
UIUX: Overhaul of UTXO list, handling of locked UTXOs and scrollbar added to tx-table web component

### LINKED ISSUE #1562: UX: Inconsistent behavior for "Create unsigned transaction"
I tested the following use case:

- Create the same unsigned transaction multiple times at different fee rates.  (This is useful for future transaction in an unknown fee environment)

There the following confusing and inconsistent behavior occurred:

1. If the tx spends more than what is unspent/unlocked in the wallet, I will be prevented from doing my use case 
    https://github.com/cryptoadvance/specter-desktop/blob/a55764ca3529cf64ca229e78f6dba74e2c04dcf3/src/cryptoadvance/specter/templates/wallet/send/new/wallet_send.jinja#L449
2. If the tx spends less than what is unspent/unlocked in the wallet, I can create the tx.  

This is inconsistent in the following way:

- It prevents double spending the same utxo only in certain cases.

My current workaround:

- Export transaction to file and delete it in specter. 

Possible ideas how to resolve this UX inconsistency:

1. Hide the locked utxos or grey their checkbox out.    This behavior would be consistent with electrum (electrum completely hides utxos which are in a saved tx).
2. Allow double spending utxos when using the coin selector. This is significant work, because all calculations are currently done with summing up utxo values -  and not unique utxos. It might also be difficult to understand for new users.

Do you think option 1 should be done? Then I would do that.

### LINKED ISSUE #1568: Bug: Rounding error when using coin selection
When using coin selection and send max (meaning using all the UTXOs selected) the amount selected is off. This seems to come from known issues in JavaScript when using floating point arithmetic (see here: https://ericlippert.com/2003/09/15/why-does-jscript-have-rounding-errors/). The relevant floating point arithemtic in this case is probably here:

https://github.com/cryptoadvance/specter-desktop/blob/c660e05ee9b8c2d7d40a272c73bbf040386b1fee/src/cryptoadvance/specter/templates/includes/coin_selection.html#L153

Screenshot of the issue:

image

moneymanolis:
### Main
- Ensure that locked UTXO / UTXO used in unsigned transactions can't be used for transactions (fixes #1562)
- Overhaul of checkbox logic in UTXO list
      1. If a locked or "locked-unsigned" UTXO is chosen, only UTXO of the respective type  can be selected anymore.
      2. Mouseover event for checkboxes added (flashing "Select checkboxes for actions.")
      3. Manage PSBT button action added
      4. Icon info popups added
See the following video:
https://user-images.githubusercontent.com/70536101/154949485-b7f45060-15f9-46a3-8b53-18268583720c.mov

- Locked UTXO or UTXO in unsigned transactions can't be chosen in the coin selection menu anymore, the checkbox is replaced by the snowflake image and the info text "Unsigned" is added to distinguish between locked and "unsigned-locked" UTXO:

image

### Misc
- Fix rounding error in coin selection (fixes #1568)
- Fix styling / text in general settings.
- Design changes to make the UTXO- / TX-list display quieter.


moneymanolis:
@relativisticelectron if you want, you could test the PR to check whether you are happy with the overall UX. Would be great. I am happy to adapt it.

relativisticelectron:
Hi, I think the freezing symbol is cute.  
I found however a way to still create multiple conflicting transactions:

- Create tx with a utxo
- Go to utxo overview.  Unfreeze the utxo.
- Create a tx with the utxo again

I am actually not sure if this should be prevented (As I said in another issue, I  actually want to create conflicting tx with differents fees).  I just wanted to point it out, and let you decide if you want to close this loophole or not.

moneymanolis:
Good point, I think we should add a little text in the front end to distinguish between UTXOs that are locked because they are used in an unsigned tx and UTXO that are locked by manually freezing them. Server sidewise, Specter lockes both via the `lockunspent` RPC call.

moneymanolis:
@relativisticelectron I reworked the UX quite a bit and also incorporated your latest point. If you could give it a try and provide feedback, that would be awesome!

relativisticelectron:
I really like it.  Good idea with the different buttons that conditionally appear. And even the checkboxes are conditionally now (cool logic!)
image

The only way to achieve an inconsistency (with the freezing status) was the following:
- Create tx that spends utxo0
- Import psbt that also spends this utxo0
- Deleting one of the transactions again   (then it can occur that the status of utxo0 is either locked, unlocked, or still frozen)

This is however such an edge case, that I am not sure it is worth it catching it.

moneymanolis:
> The only way to achieve an inconsistency (with the freezing status) was the following:
> 
>     * Create tx that spends utxo0
> 
>     * Import psbt that also spends this utxo0
> 
>     * Deleting one of the transactions again   (then it can occur that the status of utxo0 is either locked, unlocked, or still frozen)
> 
> 
> This is however such an edge case, that I am not sure it is worth it catching it. And if specter should catch it, it would be in the import logic of psbts; so it probably should not be part of this PR, but a separate one (if at all).

@relativisticelectron I will address this in a separate PR, already created an issue for it here:
https://github.com/cryptoadvance/specter-desktop/issues/1582
Thank you for taking the time and providing feedback!

relativisticelectron:
> @relativisticelectron I will address this in a separate PR, already created an issue for it here:
> #1582
> Thank you for taking the time and providing feedback!

Thank you for doing all this great work!

relativisticelectron:
I noticed after more testing, that sometimes the utxo appear "locked" and sometimes "unsigned". I don't know why.  There is 1 psbt consuming all these utxos:
image

moneymanolis:
Good catch, I will have a look, thanks.

src/cryptoadvance/specter/templates/includes/tx-table.html
```diff
@@ -1,9 +1,48 @@
...
+    type="utxo" 
+    hide-columns="category time txid confirmations action" 
+    blockhash="true" 
+    hide-switcher="true"
+    wallet="{{ wallet.alias }}"
+    scrollbar-height="333px"
+    scrollbar-default="off"
+    compact-switch="true">
+</tx-table> 
+-->
```
relativisticelectron, line 40: `#}`  end jinja2 comment.

src/cryptoadvance/specter/templates/includes/tx-table.html
```diff
@@ -1,9 +1,48 @@
+<!--
```
relativisticelectron, line 1: Please add `{#`  here, to indicate for jinja2 it should not replace here something.  Otherwise I get the error

```
  File "/home/<user>/programs/anaconda3/lib/python3.9/site-packages/cryptoadvance/specter/templates/includes/tx-table.html", line 35, in top-level template code
    wallet="{{ wallet.alias }}"
  File "/home/<user>/programs/anaconda3/lib/python3.9/site-packages/jinja2/environment.py", line 471, in getattr
    return getattr(obj, attribute)
jinja2.exceptions.UndefinedError: 'wallet' is undefined
```

src/cryptoadvance/specter/templates/wallet/history/components/total_wallet_balances.jinja
```diff
@@ -9,23 +9,24 @@
...
-            <small>( {{ wallet.balance.get("trusted", 0) | btcunitamount }} confirmed,
-                     {% if wallet.balance.get("untrusted_pending", 0) %}
-                        {{ wallet.balance.get("untrusted_pending", 0) | btcunitamount }} pending
-                     {% endif %}
-                     {% if wallet.balance.get("immature", 0) %}
-                        {{ wallet.balance.get("immature", 0) | btcunitamount }} immature
-                     {% endif %}
-            )</small>
+        {% if wallet.balance.get("untrusted_pending", 0) or wallet.balance.get("immature", 0) %}
+        <div class="smaller-card">
```
relativisticelectron, line 21: image
It would be even nicer if the amounts would be _left_ aligned, so the decimal separator is at the same x position.   And for completeness the unit  (tBTC) after each amount would be good too.

src/cryptoadvance/specter/static/styles.css
```diff
@@ -702,29 +707,67 @@ tbody tr:hover{
...
+pre::-webkit-scrollbar {
+    display: none;
+}
+
+/* Standard scrollbar along y axis
+scrollbar-color and scrollbar-width only supported by Firefox */
+
+main {
+  overflow-y: scroll;
+  scrollbar-color: hsl(40 100% 75%) hsl(40 100% 50%);
```
relativisticelectron, line 741: The chosen colors are not appealing.  How about something like  
image
it looks like
image

src/cryptoadvance/specter/templates/includes/tx-table.html
```diff
@@ -91,7 +125,87 @@
...
+            border: 1px solid var(--main-color);
+        }
+        .table-container {
+            width:  100%;
+        }
+        .scrollbar {
+            min-height: 350px;
+            overflow-y: scroll;
+            scrollbar-color: grey hsl(217, 15%, 55%); 
+            scrollbar-width: thin;
```
relativisticelectron, line 182: How about
image
looking like
image

relativisticelectron (commented):
Thanks so much for the great work. I could not create inconsistencies.

src/cryptoadvance/specter/templates/wallet/history/components/total_wallet_balances.jinja
moneymanolis: Good points, I've created a separate issue for that since this was not really meant to be in this PR in the first place.

src/cryptoadvance/specter/static/styles.css
moneymanolis: How can you not like orange :-). I wanted to have two distinct different colors for the two scrollbars. Let's do it like this I will discuss it with our design guy and present your suggestion.

src/cryptoadvance/specter/wallet.py
```diff
@@ -1544,21 +1544,21 @@ def update_balance(self):
...
-                delta = raw_tx["vout"][tx["vout"]]["value"]
-                if "confirmations" not in tx_data or tx_data["confirmations"] == 0:
-                    available["untrusted_pending"] -= delta
-                else:
-                    available["trusted"] -= delta
-                    available["trusted"] = round(available["trusted"], 8)
-            available["untrusted_pending"] = round(available["untrusted_pending"], 8)
-            balance["trusted"] = available["trusted"]
-            balance["untrusted_pending"] = available["untrusted_pending"]
+            # locked_utxo = self.rpc.listlockunspent()
```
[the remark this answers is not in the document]
moneymanolis: Changed in https://github.com/cryptoadvance/specter-desktop/pull/1580/commits/47da7a86330d3f0ef8f53ec559de682950574f44

src/cryptoadvance/specter/templates/includes/tx-table.html
moneymanolis: Added in https://github.com/cryptoadvance/specter-desktop/pull/1580/commits/05c5a62e0bac83eab31dfda5d2d935b3bb67adef

src/cryptoadvance/specter/static/styles.css
moneymanolis: Addressed in https://github.com/cryptoadvance/specter-desktop/pull/1580/commits/05c5a62e0bac83eab31dfda5d2d935b3bb67adef

moneymanolis:
@relativisticelectron see my latest commit: This is a compromise regarding the scrollbar functionality and design also after some internal discussions. Thanks for your input.

moneymanolis:
@cypherhoodlum would be great to get feedback from you on this PR regarding the scrollbar implementation for the main div, the tx table and the tx table used in the coin selection. (Of course, if you stumble upon anything else which you would improve UIUX-wise in this PR, let me know as well)

cypherhoodlum:
> @cypherhoodlum would be great to get feedback from you on this PR regarding the scrollbar implementation for the main div, the tx table and the tx table used in the coin selection. (Of course, if you stumble upon anything else which you would improve UIUX-wise in this PR, let me know as well)

I think the scrollbars look and feel great. No issues there.

However, I think the new way of showing the balances is slightly harder to read than the original version.

Original:
Screenshot from 2022-03-23 00-15-25


New:
Screenshot from 2022-03-23 00-15-59

I think it's good to unify the overall visual look by putting elements in boxes with rounded borders (new version) but the readability should not suffer. Maybe the first row could have a bigger font size than the others? Another thing worth trying out is to align the numbers/balances vertically. Maybe the whole box could be a bit bigger since it's probably the most important element on the whole page. It's at least the second most important element if the transaction history is the most important one. So I think the whole box and the total amount could be accentuated some more. What do you think @moneymanolis ?

cypherhoodlum:
Maybe something like this:

moneymanolis:
> I think it's good to unify the overall visual look by putting elements in boxes with rounded borders (new version) but the readability should not suffer. Maybe the first row could have a bigger font size than the others? Another thing worth trying out is to align the numbers/balances vertically. Maybe the whole box could be a bit bigger since it's probably the most important element on the whole page. It's at least the second most important element if the transaction history is the most important one. So I think the whole box and the total amount could be accentuated some more. What do you think @moneymanolis ?

Thanks a lot for the feedback and the PR, @cypherhoodlum.
Good points! 

Since I would like to wrap this PR up soon (it is already too big actually), though, I would keep the style for now and move the discussion and work on the balance box to this already opened issue: #1627

k9ert:
@cypherhoodlum awesome feedback and thanks for your PR. It's yet a big improvement. But i also agree with @moneymanolis , let's not make that PR any bigger. We struggled now a lot with the UI-tests in this PR. The last 12 commits are only regarding improving testing. It's high likely your PR would break them (again). And even if not, this PR is a moving goal-post since almost the beginning.
There are people who think that feature-branches should only be open for a couple of hours. This one is now open for more than 40 days.

cypherhoodlum:
> But i also agree with @moneymanolis , let's not make that PR any bigger. We struggled now a lot with the UI-tests in this PR. The last 12 commits are only regarding improving testing. It's high likely your PR would break them (again).

Right, that's understandable. I will make a separate PR related to #1627.

k9ert:
This is really an unlucky PR. After scratching our heads why the builds are not kicking in at all (probably an [issue](https://status.cloud.google.com/) with the Google-cloud), now the linter is broken.
```
  File "/usr/local/bin/black", line 8, in <module>
    sys.exit(patched_main())
  File "src/black/__init__.py", line 1423, in patched_main
  File "src/black/__init__.py", line 1409, in patch_click
ImportError: cannot import name '_unicodefun' from 'click' (/usr/local/lib/python3.10/site-packages/click/__init__.py)
```
Investigating now ...

k9ert:
Fixed in https://github.com/cryptoadvance/specter-desktop/pull/1642
Build is now hopefully green.

Issue #1467, opened 2021-11-09
Two minor UI things:


moritzwietersheim:
Two minor UI thingys:

![image](https://user-images.githubusercontent.com/47259243/141015287-5eb1ea6a-8aab-46ac-9d9d-8d1bf029fa27.png)
This looks not so nice. Could be shortened to "Cancel TX" or displayed in 2 lines. 
"Speed Up" button could be smaller or read "RBF" if that is not too cryptic.

and: 

![image](https://user-images.githubusercontent.com/47259243/141015304-7cd95f74-90e9-4a1e-ab06-e42dfb5b274e.png)
This maybe needs more or a better explanation where to find the logfile.

relativisticelectron:
Interesting.  I see no cancel button (current master)
![image](https://user-images.githubusercontent.com/60378539/154061986-f44b4b73-22fa-46b3-8fd1-e03581a1d36d.png)

relativisticelectron:
And the sat/vbyte is empty when I click Speed up.  
![image](https://user-images.githubusercontent.com/60378539/154115633-447c72e5-0eac-43ad-ab1d-a8a5b3bc5244.png)
Perhaps we could make a helpful default value as old tx fee + min relay fee?

relativisticelectron:
The cancel button only appears when sending out of the own wallet.
The positioning of the buttons depends on the width. here some examples:
![image](https://user-images.githubusercontent.com/60378539/154262285-3eb5a50f-c6fc-4bc8-82ef-125b177a6672.png)
![image](https://user-images.githubusercontent.com/60378539/154262351-543964b3-0aa7-474b-84fb-e3a4719bd132.png)

![image](https://user-images.githubusercontent.com/60378539/154262380-339f1abd-8058-4752-8672-d4966389112b.png)

moneymanolis:
Just for the record, the logging tool-tip was adjusted here: #1682.

PR #1585, opened 2022-02-15
BugFix in RBF flow

relativisticelectron:
The missing fee rate in https://github.com/cryptoadvance/specter-desktop/issues/1467#issuecomment-1040562857  let me to discover that `this.tx.vsize` in https://github.com/cryptoadvance/specter-desktop/blob/352355e8d711be8ec5ee9372c92dbabbd6435359/src/cryptoadvance/specter/templates/includes/tx-row.html#L318 was `None`.
This PR is a fix and now there is a suggested higher fee rate visible for RBF:
image

Additionally was the `fillform = True` not set, which led to that the "Create Transaction" form was not filled, after one clicked on 
image


relativisticelectron:
Thanks.

PR #1586, opened 2022-02-15
Update the daemon.md instructions to include the tarball release

GoofyAF:
The linux tarball release of specter-desktop has some locale issues that have not yet been fixed. To get around this, a short script is added to the systemd service file that sets the correct locale then launches the app.


k9ert:
Thank you very much. Quite a good point!

Issue #1577, opened 2022-02-11
Feature missing: Import of raw transaction not supported


relativisticelectron:
It seems that import of fully signed raw (hex) transactions is not supported.

Steps to reproduce:
- Create a fully signed tx with another wallet software (do not broadcast)
- In specter go to Send --> Import --> Paste the hex of the raw tx in the field

```
[2022-02-11 20:58:58,637] ERROR in specter_error: Unexpected error:
[2022-02-11 20:58:58,638] ERROR in specter_error: ----START-TRACEBACK-----------------------------------------------------------------
[2022-02-11 20:58:58,638] ERROR in specter_error: Incorrect padding
Traceback (most recent call last):
  File "/home/<user>/programs/anaconda3/lib/python3.9/site-packages/cryptoadvance/specter/server_endpoints/wallets.py", line 683, in import_psbt
    psbt = wallet.importpsbt(b64psbt)
  File "/home/<user>/programs/anaconda3/lib/python3.9/site-packages/cryptoadvance/specter/wallet.py", line 2008, in importpsbt
    psbt = self.PSBTCls(
  File "/home/<user>/programs/anaconda3/lib/python3.9/site-packages/cryptoadvance/specter/util/psbt.py", line 242, in __init__
    psbt = self.PSBTCls.from_string(psbt)
  File "/home/<user>/programs/anaconda3/lib/python3.9/site-packages/embit/psbt.py", line 578, in from_string
    return cls.from_base64(s, compress=compress)
  File "/home/<user>/programs/anaconda3/lib/python3.9/site-packages/embit/psbt.py", line 561, in from_base64
    raw = a2b_base64(b64)
binascii.Error: Incorrect padding
[2022-02-11 20:58:58,639] ERROR in specter_error: ----END---TRACEBACK-----------------------------------------------------------------
```

I could not find a decoderawtransaction analog  in https://github.com/diybitcoinhardware/embit.  But it seems the functionality is in https://github.com/cryptoadvance/specter-desktop/blob/6ab2bafbdb2a6b0d49061ed4e8c3f3fffa5fd725/src/cryptoadvance/specter/util/tx.py#L44

One should be able to reconstruct the full psbt using decoderawtransaction...

moneymanolis:
[...]

PR #1591, opened 2022-02-17
Feature: Import of raw transaction

relativisticelectron:
Original issue: https://github.com/cryptoadvance/specter-desktop/issues/1577

Status: works for segwit and nested inputs.

This PR enables to import raw (fully signed) tx (screenshots of regtest):
image
image


relativisticelectron:
The taproot signing error is 
```
[2022-02-18 09:18:53,586] ERROR in specter_error: Unexpected error:
[2022-02-18 09:18:53,586] ERROR in specter_error: ----START-TRACEBACK-----------------------------------------------------------------
[2022-02-18 09:18:53,586] ERROR in specter_error: Make sure you have entered the wallet file password correctly. (If your wallet is not encrypted submit empty password)
  File "/home/<user/programs/anaconda3/lib/python3.9/site-packages/cryptoadvance/specter/server_endpoints/wallets.py", line 562, in send_new
    signed_psbt = app.specter.device_manager.get_by_alias(
  File "/home/<user/programs/anaconda3/lib/python3.9/site-packages/cryptoadvance/specter/devices/bitcoin_core.py", line 225, in sign_psbt
    raise Exception(
Exception: Make sure you have entered the wallet file password correctly. (If your wallet is not encrypted submit empty password)
[2022-02-18 09:18:53,587] ERROR in specter_error: ----END---TRACEBACK-----------------------------------------------------------------
```

relativisticelectron:
**Status:**

- All relevant wallet types including multisig work and are included in the test

moneymanolis:
Thanks so much for the PR. Will do with the review and testing!

src/cryptoadvance/specter/wallet.py
```diff
@@ -2073,3 +2075,44 @@ def addresses_info(
 
     def __repr__(self) -> str:
         return f"<{self.__class__.__name__} name={self.name } alias={self.alias}>"
+
+    def convert_rawtransaction_to_psbt(self, rawtransaction) -> str:
```
k9ert, line 2079: Can you migrate that method maybe to `util/tx.py` ? There is not much `self` in that method apart from self.rpc and maybe you pass the rpc-object into the method?
wallet.py is already way too long. We need to find ways how to make it even shorter than it is now.

tests/test_specter.py
```diff
@@ -227,3 +227,206 @@ def test_abandon_purged_tx(
...
+def test_import_raw_transaction(
+    caplog, docker, request, devices_filled_data_folder, device_manager
+):
+    # Specter should support calling abandontransaction if a pending tx has been purged
+    # from the mempool. Test starts a new bitcoind with a restricted mempool to make it
+    # easier to spam the mempool and purge our target tx.
+    # TODO: Similar test but for maxmempoolexpiry?
+
+    # Copied and adapted from:
+    #    https://github.com/bitcoin/bitcoin/blob/master/test/functional/mempool_limit.py
```
k9ert, line 242: copy/paste fragment, please delete.

relativisticelectron:
Oh, I just noticed I also have to move the testcase to a different file.

k9ert:
I was not so happy with the test-method. I refactored a bit but didn't wanted to push it directly to the branch. Please check here:
https://github.com/relativisticelectron/specter-desktop/pull/1

src/cryptoadvance/specter/util/tx.py
```diff
@@ -71,3 +83,45 @@ def decoderawtransaction(hextx, chain="main"):
         ],
     }
     return result
+
+
+def convert_rawtransaction_to_psbt(wallet, rawtransaction) -> str:
```
k9ert, line 88:
```suggestion
def convert_rawtransaction_to_psbt(wallet_rpc, rawtransaction) -> str:
```
Please change it so that the function directly gets an RPC-object. It might look as it doesn't matter but from an architectual Point of view, it's not so cool to have higher order business-objects (like wallet) in lower level helper-functions.
It would implicitely have a dependency to Wallet which it definitely doesn't have.

relativisticelectron: Done

moneymanolis:
@relativisticelectron: @stepansnigirev showed me how we could do the whole approach with the help of his fantastic embit library (https://github.com/diybitcoinhardware/embit) - which is heavily used by Specter Desktop anyways. This would have the advantage of less RPC calls, we wouldn't have to rely on HWI and this works for unsigned TX as well. What do you think? Shall I push this directly in this branch or do you prefer a PR?

Quick preview of the rewrite:
image

relativisticelectron:
> @relativisticelectron: @stepansnigirev showed me how we could do the whole approach with the help of his fantastic embit library (https://github.com/diybitcoinhardware/embit) - which is heavily used by Specter Desktop anyways. This would have the advantage of less RPC calls, we wouldn't have to rely on HWI and this works for unsigned TX as well. What do you think? Shall I push this directly in this branch or do you prefer a PR?
> 
> Quick preview of the rewrite: image

@moneymanolis : Great. That is a significant simplification of the code. Yes you can push it directly onto the branch.

moneymanolis:
OK, will also adapt the front end.

moneymanolis:
@relativisticelectron would be great if you could have a look at the newest version of the PR after my commit (https://github.com/cryptoadvance/specter-desktop/pull/1591/commits/bcaacf7b4b4c7205a2f52658d6dfd7cf5c6bb995) and perhaps test with one or two tx. I also added a more robust test case, hope the test will be green. Took me a while to get it green in local dev. Would also be great if you could check whether sth. else in the front should be adjusted.

src/cryptoadvance/specter/templates/wallet/send/import/wallet_importpsbt.jinja
```diff
@@ -4,13 +4,13 @@
 	{% include "includes/qr-scanner.html" %}
 	{% from 'wallet/send/components/send_nav.jinja' import send_nav %}
 	{{ send_nav('import_psbt', wallet_alias) }}
-		<h1 class="padded">{{ _("Import PSBT transaction:") }}</h1>
-		<span class="center note padded">{{ _("Paste base64-encoded PSBT transaction here to import.") }}</span>
+		<h1 class="padded">{{ _("Import transaction") }}</h1>
+		<span class="center note padded" style="margin-bottom: 10px; font-size:1em">{{ _("PSBT (base64) or old-school transaction (hex)") }}</span>
```
relativisticelectron, line 8: "old-school" is maybe too imprecise, because for signed tx is it not "old" at all, but the default.

tests/test_util_tx.py
```diff
@@ -0,0 +1,119 @@
...
+
+        logging.info(
+            f"created wallet key_type '{key_type}' keys {[key.json for key in keys]}  wallet.account_map {wallet.account_map}"
+        )
+
+        if key_type == "wpkh":
+            # More solid test possible: Deriving the signed tx without any RPC calls used in convert_rawtransaction_to_psbt()
+            # fund it
+            bitcoin_regtest.testcoin_faucet(wallet.getnewaddress(), amount=3)
+            wallet.save_to_file()
```
relativisticelectron, line 82: `save_to_file` can be removed

tests/test_util_tx.py
```diff
@@ -0,0 +1,119 @@
...
+            assert signed_tx == wallet.rpc.finalizepsbt(b64_psbt)["hex"]
+            # Check import of unsigned raw tx
+            b64_psbt_from_unsigned = convert_rawtransaction_to_psbt(wallet.rpc, raw_tx)
+            psbt_dict = {"psbt": b64_psbt_from_unsigned, "complete": ""}
+            signed_psbt = sign_psbt_with_devices(psbt_dict)
+            assert signed_tx == wallet.rpc.finalizepsbt(signed_psbt)["hex"]
+        else:
+            # fund it
+            bitcoin_regtest.testcoin_faucet(wallet.getnewaddress(), amount=3)
+            wallet.save_to_file()
```
relativisticelectron, line 101: `save_to_file` can be removed

tests/test_util_tx.py
```diff
@@ -0,0 +1,119 @@
...
+            keys,
+            used_devices,
+        )
+
+        logging.info(
+            f"created wallet key_type '{key_type}' keys {[key.json for key in keys]}  wallet.account_map {wallet.account_map}"
+        )
+
+        if key_type == "wpkh":
+            # More solid test possible: Deriving the signed tx without any RPC calls used in convert_rawtransaction_to_psbt()
```
relativisticelectron, line 79: There is 1 rpc call hidden in `hot_wallet_device_1.sign_raw_tx(raw_tx, wallet)`

tests/test_util_tx.py
```diff
@@ -0,0 +1,119 @@
...
+            key_type,
+            keys,
+            used_devices,
+        )
+
+        logging.info(
+            f"created wallet key_type '{key_type}' keys {[key.json for key in keys]}  wallet.account_map {wallet.account_map}"
+        )
+
+        if key_type == "wpkh":
```
relativisticelectron, line 78: Do you know why `hot_wallet_device_1.sign_raw_tx(raw_tx, wallet)` works only for this wallet type?

relativisticelectron (commented):
Thanks for your work. I tested the UI and it works great.

src/cryptoadvance/specter/templates/wallet/send/import/wallet_importpsbt.jinja
k9ert: "conventional" or "traditional" might be a netter term?

tests/test_util_tx.py
k9ert: Sidenote: If you start the server, there is now not only `--tracerpc` but also `--tracerequests` which spits out logs of calls to core and/or ALL calls made with the requests-library.

src/cryptoadvance/specter/templates/wallet/send/import/wallet_importpsbt.jinja
moneymanolis: You are right, @relativisticelectron. How about "serialized"?

relativisticelectron: Yes, "serialized" is very precise.

tests/test_util_tx.py
moneymanolis: OK, I didn't understand what it was used for anyways.

relativisticelectron: I used it for debug purposes and forgot to delete it.

tests/test_util_tx.py
moneymanolis: It is defined here: https://github.com/cryptoadvance/specter-desktop/blob/e9791068f0fc680397debc7d2176ae0a35ea9957/src/cryptoadvance/specter/devices/bitcoin_core.py#L230
It is a method of the Bitcoin Core Hot Wallet "Device" in Specter. AFAIK, it hasn't been used anywhere else thus far.
I also tried to use it for the Nested Segwit and Multisig Test Wallets but it became too complicated. (At least this is my theory:) For `signrawtransactionwithwallet`(which is the RPC call used by `sign_raw_tx`) to work with those two you need to pass additional information about the redeem and witness scripts in the RPC call. I managed to make it work with bitcoin-cli "manually" for the Single Sig Nested Segwit case (with one ordinary Core wallet with private keys enabled) but Specter uses basically two wallets here, a watch only wallet in Core and an additional Core wallet with private keys (they get passed on with descriptors from a BIP 39 seed). So, somewhere along those lines our test Core hot wallet doesn't have all the info to sign.

I think the test suite is robust enough for this PR, and I would postpone further digging into this matter to a future use case of such tests - where we then perhaps can do it together with the grandmaster @stepansnigirev ;-). Having said this, @relativisticelectron, feel free to try it out.

tests/test_util_tx.py
relativisticelectron: Thanks :-)

tests/test_util_tx.py
relativisticelectron: Thanks for the explanation.

k9ert:
Thanks folks!

src/cryptoadvance/specter/templates/wallet/send/import/wallet_importpsbt.jinja
relativisticelectron: @k9ert :  I marked this point too early as resolved.   The https://github.com/relativisticelectron/specter-desktop/commit/3bc2f8f8dbe48f879f93969b2a99b5a12be0ba65  is after you merged the branch.

relativisticelectron:
@k9ert : could you also merge the last commit https://github.com/relativisticelectron/specter-desktop/commit/c3ca4eb65510a61f63d9ad4b2c3c87390366e33e  ?

Issue #1158, opened 2021-05-11
Linux: App icon in dock missing


dennisreimann:
The app icon of v1.3.1 is missing for me on Linux (Pop!_OS). The icon in the status bar is present though.

![specter-icon-linux](https://user-images.githubusercontent.com/886/117796805-b2642480-b24f-11eb-82f2-73aa737c42d5.png)

kravens:
![image](https://user-images.githubusercontent.com/9646902/151961341-da7e05ab-f0ca-45f7-9f97-8f152b5a88d4.png)
Also for me in the latest version 1.8.0 the taskbar icon is a blank sheet / missing, on Win10.

dennisreimann:
With v1.10 it is present, but very blurry.

![grafik](https://user-images.githubusercontent.com/886/171175048-71f952ac-e94e-49e5-a961-d0ac22a1fd09.png)

k9ert:
Yes, it has been fixed in https://github.com/cryptoadvance/specter-desktop/pull/1592/files . However, i guess the better image would have been the bigger one in the same directory. That's probably the reason.
https://github.com/cryptoadvance/specter-desktop/tree/master/pyinstaller/electron/assets

PR #1592, opened 2022-02-17
Fix App Icon for Gnome Dock #1158

salderma:
Fixes missing app icon.

Kudos to this for the hint - https://github.com/electron-userland/electron-builder/issues/2269#issuecomment-342168989


k9ert:
Great! Thank you very much. I've spent already a lot of time on this and couldn't find the solution for some reason ...
image

Issue #673, opened 2020-11-21
Scrollbar is missing on Transactions screen


BTCBellyButton:
Not a big deal but when you have hundreds of transactions to go through it'd be convenient.
It'd be useful also on the side bar when you have a lot of wallets.

relativisticelectron:
The scrollbar PR was merged. https://github.com/cryptoadvance/specter-desktop/pull/1598    Can you close the Issue?

PR #1598, opened 2022-02-20
UI: Added scrollbar for transaction list

relativisticelectron:
Implemented based on suggestion in https://github.com/cryptoadvance/specter-desktop/issues/673

It's a very quick-and-easy fix.  The scrollbar is not ideal, but better than no scrollbar at all.  Best would be to scroll only the transactions (without the header).  

(scrolling example)
image
image


k9ert:
Thanks, good point!

Issue #1565, opened 2022-02-09
UX/UI: Redesign of "Add new device" --> "Electrum"  --> "Paste xpub" required 


relativisticelectron:
The current "Add new device" --> "Electrum" --> "Paste xpub" flow is not usable.  See https://github.com/cryptoadvance/specter-desktop/issues/1539#issuecomment-1033506364

Situation:

- The "Add new device" -> "manual configuration" is very clear and works for this use-case. However users that want to add an "electrum" device will not click on "manual configuration" because "electrum" appears in the device list.
- https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/templates/device/new_device/new_device_keys.jinja Seems not to be designed to allow manual editing of fingerprint, derivation path and xpub, which would be required in this use case.

This issue should be a discussion on how to redesign the flow.

moneymanolis:
I came across this as well while testing for one of your PRs. I will have a look.

relativisticelectron:
Relevant is also  the comment https://github.com/cryptoadvance/specter-desktop/issues/1539#issuecomment-1033637295

relativisticelectron:
I removed electrum from the list:   https://github.com/cryptoadvance/specter-desktop/compare/master...relativisticelectron:remove_electrum_from_device_list?expand=1

Would this minimal change be enough, or do you think this needs more changes?

PR #1622, opened 2022-03-03
Add custom html page for "Add Device" --> "Electrum"

relativisticelectron:
This PR is a minimal change to solve https://github.com/cryptoadvance/specter-desktop/issues/1565


src/cryptoadvance/specter/managers/device_manager.py
```diff
@@ -125,8 +125,12 @@ def supported_devices_for_chain(self, specter):
                 for device_class in device_classes
                 if device_class.bitcoin_core_support
             ]
+
         if BitcoinCoreWatchOnly in devices:
             devices.remove(BitcoinCoreWatchOnly)
+        # electrum needs to be configured via the "Manual configuration" dialog and excluded from the list
+        if Electrum in devices:
```
k9ert, line 132: IMHO, the question is: Why it's in the list in the first place. And if there is a specific way to implement that, why don't we support that. I'll give more thoughts on a comment to the PR.

k9ert:
The Endresult of this PR looks like adding the Electrum Device and then removin it in a very sophisticated way. 
A better idea would imho be, that we experiment with UI adaptions, so that we have device-specific UI.
In this case, it's about the `new_device_keys` endpoint. What if we have a standard device-attribute like:
`template_new_device_keys = "device/new_device/new_device_keys.jinja"` ([here](https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/device.py#L34])
So if you inherit a Device, you could overwrite that like:
`Electrum.template_new_device_keys = "device/new_device/new_device_keys_electrum.jinja"` ([here](https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/devices/electrum.py#L16))

And then at the endpoint, we pick the template which is specified at the device?! 
CC @kdmukai

relativisticelectron:
Thanks for the great feedback!   I will try that.

relativisticelectron:
TODO:

- [ ] Instead of showing instruction on how to fill `new_device_manual` , collect the necessary information here and then add the device directly.

relativisticelectron:
While not pretty, the user instructions and the way to enter the data is now merged in 1 page:
image

relativisticelectron:
It looks now like this
image


A problem I identified is:  If there is an error after POST, then it is stuck on the page `new_device_manual.jinja`.   Does anyone have suggestions on how to fix this?

relativisticelectron:
I solved also the problem with the redirect if an error occurs; I created a new endpoint to solve that.  

@moneymanolis : Could you review the PR?

relativisticelectron:
@moneymanolis : It fully works now. However I think the code structure could be improved, and I would be glad for suggestions.

moneymanolis:
@relativisticelectron I could already boil down the server side code a lot and I am working on the front end now. I would toggle the explanations like this if you don't mind:
https://user-images.githubusercontent.com/70536101/164073445-568979b1-2073-4f3d-b1cc-cd486644b4aa.mov

If this is fine by you, I'd push directly to your remote repo, once I am finished, since I didn't really change a lot on the server side, just streamlined it.

relativisticelectron:
Thanks so much for taking the time to improve the code!  Yes please push on the branch.

moneymanolis:
I will include all screenshots. One question, regarding the derivation path, is m/0' the default in Electrum?

moneymanolis:
Noteworthy changes in https://github.com/cryptoadvance/specter-desktop/pull/1622/commits/f6d8d19d19f09a1bbbc3c16b49ba0402de3743bb:
- I picked up @k9ert idea with the template attribute in the device class, this allowed me to streamline the endpoint code quite a bit, especially since we can keep using the `new_device_keys` endpoint and all its functionality for the Electrum device as well. One of the advantages is that we now also get a popup asking to directly create a wallet with the new device (see also video 2 below).
- The front end code in `new_device_keys_electrum.jinja` to create the xpub in the right format for `Key.parse_xpub()` on the server side worked with `oninput` even though I don't exactly know why. I felt like it is safer and also more explicit to use a more dedicated JS function which is called `onsubmit` which is a regular EventHandler for the form element (unlike the oninput which only works for input fields etc. - it probably works here because we have input fields, but still, it felt a bit shaky). 

This is how it looks now:

Explainer demo:
https://user-images.githubusercontent.com/70536101/164201945-8a55c732-3486-4d15-8bef-7d53a5ecdc40.mov

Device creation demo:
https://user-images.githubusercontent.com/70536101/164201968-a6e08cd4-8c20-4f27-a922-3e2928d4468c.mov

TODOs

- [ ] If the front end is aggreed upon, a little additional cypress test: adding an Electrum device
- [ ] Make cypress tests green

src/cryptoadvance/specter/server_endpoints/devices.py
```diff
@@ -174,8 +180,8 @@ def new_device_keys(device_type):
                 )
 
     return render_template(
-        "device/new_device/new_device_keys.jinja",
-        device_class=get_device_class(device_type),
+        template,
+        device_class=device_class,
```
relativisticelectron, line 184: It would be good if the form fails and refreshed, that the previously filled out fields are in there.    
```
        device_name=device_name,
        xpub=xpub,
        derivation_path=derivation_path,
        root_fingerprint=root_fingerprint,
```

src/cryptoadvance/specter/server_endpoints/devices.py
```diff
@@ -81,6 +85,8 @@ def new_device_keys(device_type):
                 except:
                     err = _("Failed to parse these xpubs") + ":\n" + "\n".join(xpub)
                     break
+        if device_type == "electrum":
+            keys.append(Key.parse_xpub(request.form["xpub"]))
```
relativisticelectron, line 89: `parse_xpubs`  catches exceptions and gives the user a nice error message. However when entering a wrong xpub, `parse_xpub` raises an exception that is not caught.

relativisticelectron (commented):
Hi @moneymanolis   Thanks for taking the time to improve the code! It looks much cleaner.  :+1:
Thanks for making also the front-end cleaner.

relativisticelectron:
> I will include all screenshots. One question, regarding the derivation path, is m/0' the default in Electrum?

Yes.  Thanks for asking!  When investigating this If found out that https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/key.py#L144  detects depth 1 and  completely overrides the derivation path and root fingerprint!

So a user needs to **only** put in the master public key!

I guess for non-default electrum wallets we should keep the fingerprint and derivation path options. But they could be hidden under a collapsible menu.

src/cryptoadvance/specter/server_endpoints/devices.py
```diff
@@ -298,23 +304,11 @@ def new_device_manual():
...
-            if len(failed) > 0:
-                err = _("Failed to parse these xpubs") + ":\n" + "\n".join(failed)
-            if err is None:
-                device = app.specter.device_manager.add_device(
-                    name=device_name, device_type=device_type, keys=keys
-                )
-                return redirect(
-                    url_for("devices_endpoint.device", device_alias=device.alias)
-                )
+            err, redirect_obj = action_newcolddevice(
```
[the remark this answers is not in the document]
moneymanolis: Ah, @relativisticelectron used this / his function (that I deleted) here as well, shall we just revert this part to master again by having the part below again or do I miss sth.?
```
            if not device_name:
                err = _("Device name cannot be empty")
            elif device_name in app.specter.device_manager.devices_names:
                err = _("Device with this name already exists")
            xpubs = request.form["xpubs"]
            if not xpubs:
                err = _("xpubs name cannot be empty")
            keys, failed = Key.parse_xpubs(xpubs)
            if len(failed) > 0:
                err = _("Failed to parse these xpubs") + ":\n" + "\n".join(failed)
            if err is None:
                device = app.specter.device_manager.add_device(
                    name=device_name, device_type=device_type, keys=keys
                )
                return redirect(
                    url_for("devices_endpoint.device", device_alias=device.alias)
                )
```

src/cryptoadvance/specter/server_endpoints/devices.py
k9ert: I think if the `parse_xpub` method would raise `SpecterError` instead of Exceptions, we also would get better Error-messages?! In general, don't catch Exceptions if you don't know how such a potential issue arises. If you know, throw `SpecterError` with a message which even a user could understand.

src/cryptoadvance/specter/server_endpoints/devices.py
moneymanolis: I don't think this is so important, since we are dealing with only 3 input fields here. Having said that, feel free to implement.

src/cryptoadvance/specter/server_endpoints/devices.py
moneymanolis: When entering a wrong xpub I get this:

grafik

When entering a wrong xpub with manual configuration (which uses `parse_xpubs`), I get this:
grafik

I'd say both messages are not so great, but they don't really differ to me.

moneymanolis:
> When investigating this If found out that https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/key.py#L144 detects depth 1 and completely overrides the derivation path and root fingerprint!
> 
> So a user needs to **only** put in the master public key! The derivation path and root fingerprint can stay empty.
> 
> I think for non-default (BIP39 imported or so) electrum wallets we should keep the fingerprint and derivation path options. But they could be hidden under a collapsible "Advanced / BIP39 settings" item.

I think it is not really correct to talk of a master public key here, since the fingerprint of a master key is 0x00000000 and its depth is 0. But, since it displayed like this in Electrum, I'd suggest we keep it like this and better also keep the fingerprint and derivation path to minimise confusion.

src/cryptoadvance/specter/server_endpoints/devices.py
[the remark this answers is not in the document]
relativisticelectron: Yes this has to be reverted to master.

relativisticelectron:
> > When investigating this If found out that https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/key.py#L144 detects depth 1 and completely overrides the derivation path and root fingerprint!
> > So a user needs to **only** put in the master public key! The derivation path and root fingerprint can stay empty.
> > I think for non-default (BIP39 imported or so) electrum wallets we should keep the fingerprint and derivation path options. But they could be hidden under a collapsible "Advanced / BIP39 settings" item.
> 
> I think it is not really correct to talk of a master public key here, since the fingerprint of a master key is 0x00000000 and its depth is 0. But, since it displayed like this in Electrum, I'd suggest we keep it like this and better also keep the fingerprint and derivation path to minimise confusion.

Well, if 99% of all users (creating adding an electrum device)  just need to go into the electrum wallet info section and don't need to type anything into the electrum console, then there is no point of letting them do that.  

PS: Sorry for closing/opening... wrong buttton...

src/cryptoadvance/specter/server_endpoints/devices.py
relativisticelectron: Ok, I'll do that :-)

moneymanolis:
> Well, if 99% of all users (adding an electrum device) just need to go into the electrum wallet info section and don't need to type anything into the electrum console, then there is no point of letting them do that.

I know too little about Electrum, their defaults, their derivation paths, their BIP39 suppport etc. If 99% just need this "master" public key, then you are completely right.  

Probably the most pragmatic way would be to hide the root fingerprint stuff in the jinja template and make a comment that it might be used in the future? But we can also just remove it.

relativisticelectron:
> I know too little about Electrum, their defaults, their derivation paths, their BIP39 suppport etc. If 99% just need this "master" public key, then you are completely right.

All  wallets created with electrum (afaik) have the m/0h derivation path. However electrum offers to import any BIP39 seed and derivation path. I guess some people did that, and it would be great to keep the support for them in the jinja template, since the code works already.

> Probably the most pragmatic way would be to hide the root fingerprint stuff in the jinja template and make a comment that it might be used in the future? But we can also just remove it.

If it is OK with you, I would put the root fingerprint and derivation path (including how to get them) into an collapsible "advanced / BIP39" panel.  No logic has to be changed.  Should I rearrange the fields?

moneymanolis:
RE https://github.com/cryptoadvance/specter-desktop/pull/1622/commits/130a2fffb7af698e9b6c3e186d76209fcb6f1759: This made the Cypress tests green in local dev for me.

moneymanolis:
> However electrum offers to import any BIP39 seed and derivation path. I guess some people did that, and it would be great to keep the support for them in the jinja template, since the code works already.

Sorry, need to clarify first:
So there are people who generated a BIP39 mnemonic and with it a seed (either manually or with some sort of HWW) and imported this into Electrum and they are using Electrum as an airgapped device then?

So far, I had only imagined the use case of someone having an "Electrum signing device" in the form of having Electrum on an airgapped laptop or so (and using all the Electrum defaults regarding derivation paths).

relativisticelectron:
> Sorry, need to clarify first: So there are people who generated a BIP39 mnemonic and with it a seed (either manually or with some sort of HWW) and imported this into Electrum and they are using Electrum as an airgapped device then?

Yes, or that is possible. But I though more of people who have created a seed with another device (e.g. HWW) and then they downgraded their security by importing it into a hot-wallet electrum.  (If electrum becomes perhaps part of a newly created multisig (with specter)  it wouldn't even be downgrading security).  In any case electrum offers this BIP39 import possibility, and I don't know how or how often people use it.

moneymanolis:
I don't understand the "Electrum hot wallet" in the Specter context, why not just import the BIP39 mnemonic in Specter (thus using a Bitcoin Core Hot wallet)? Then, you could also actually use the hot wallet as a signer. 

There is no functionality to sign with an Electrum hot wallet.

I would rather better / more explicitly implement in the front end a BIP39 watch only wallet then putting this in the Electrum device category.

In sum: I would just delete the Derivation path and Root fingerprint fields for the Electrum device here:

grafik

relativisticelectron:
> I would rather better / more explicitly implement in the front end a BIP39 watch only wallet then putting this in the Electrum device category.

I agree with this point.

> In sum: I would just delete the Derivation path and Root fingerprint fields for the Electrum device here:

OK.

moneymanolis:
Alright, I made a new issue out of the last point. Will finalise it as discussed and add a little Cypress test for it.

moneymanolis:
@relativisticelectron please have a final look, thanks.

relativisticelectron (commented):
Thanks for the work. I tested it and it also looks great.

moneymanolis:
@k9ert please check whether you are happy with the Cypress changes: the Electrum device test is in a new spec file dedicated to (probably special?) devices.

src/cryptoadvance/specter/server_endpoints/devices.py
```diff
@@ -40,6 +42,8 @@ def new_device_type():
 @devices_endpoint.route("/new_device_keys/<device_type>/", methods=["GET", "POST"])
 @login_required
 def new_device_keys(device_type):
+    device_class = get_device_class(device_type)
+    template = device_class.template
```
k9ert, line 46:
```suggestion
    template = get_device_class(device_type).template
```

src/cryptoadvance/specter/server_endpoints/devices.py
k9ert: That's done, no?

k9ert (approved):

k9ert:
LGTM. Had one tiny recommendation. Apart from that, ready to merge.

src/cryptoadvance/specter/server_endpoints/devices.py
moneymanolis: But, then we'd have no `device_class` variable to pass in `render_template()`, no?

k9ert: Sorry, was too quick with that.

PR #1623, opened 2022-03-03
Service class refactoring

k9ert:
Some changes to the extension-framework:
* Specter now has a dictionary called `ext` which contains all the extensions. This might be a semi-good idea.
* As we now might have external extensions in separate repos and therefore directories like `src/cryptoadvance`, we need a more sophisticated method to detect, whether we're in the specter-desktop repo-folder.
* There is now a way to schedule events with the help of the `Flask-APScheduler` module. A scheduler instance will be created and passed to the `after_serverpy_init_app` callback. The dummyextension has some example code on how to use this.
* Now there is also an `ServiceUnencryptedStorage` class available in order to store service specific stuff unencrypted.
* Some bugfixes


moneymanolis:
Opportunity to change Services to Plugins in this PR
image

src/cryptoadvance/specter/services/callbacks.py
```diff
@@ -11,5 +11,6 @@
     
 """
 
-afterServiceManagerInit = "afterServiceManagerInit"
+afterServiceManagerInit = " afterServiceManagerInit"
```
moneymanolis, line 14:
```suggestion
afterServiceManagerInit = "afterServiceManagerInit"
```

src/cryptoadvance/specter/services/service_encrypted_storage.py
```diff
@@ -98,6 +99,14 @@ def get_service_data(self, service_id: str) -> dict:
         return service_data
 
 
+class ServiceUnencryptedStorage(ServiceEncryptedStorage):
```
moneymanolis, line 102: Is that really enough? No "encrypted=False" etc.

k9ert: I've added some sanity-checking and a documentation for that class.

PR #1626, opened 2022-03-07
Adding Foundation Passport as supported device

jonathancross:
Per https://github.com/Foundation-Devices/passport-firmware/issues/6#issuecomment-946897266


k9ert:
Thanks for the addition.
Did you tested the passport support? QR code and SD-card? Which version did you used exactly?

jonathancross:
Unfortunately I do not have a Passport device to test.
I was researching support for various wallets (like Specter) as a prerequisite towards ordering one.

k9ert:
Sorry but i don't think we should merge this if we can't even find someone who confirms that it works. We have to pray that there isn't someone who complains that it's not working.

jonathancross:
No problem, makes sense.
Hopefully I can get a device and test in the future.

PR #1633, opened 2022-03-22
Bugfix: versionChecker and downloadloc

k9ert:
fixes:
```
2022-03-21T23:01:38.097Z [info] : stderr-SPECTERD: ERROR in version: name 'self' is not defined
  File "cryptoadvance\specter\util\version.py", line 148, in _get_latest_version_from_github
NameError: name 'self' is not defined
SPECTERD: INFO in version: version checked. upgrade: False
```
and also fixed downloadlocation.


moneymanolis (approved):
LGTM

PR #1648, opened 2022-04-01
Feature: Menu extensions can extend menus

k9ert:
With this PR, extensions are able to contribute tabs to settings and wallet-extensions.
Here is an example for the wallet:

image

There is a description how to do that in the docs-folder with an example for the settings_tabs:
image


k9ert:
@wombat6 maybe you can review this?

wombat6:
Yes, sure thing @k9ert. I should be able to get to it this weekend.

moneymanolis:
Tested: Adding a tab with a mini jinja in the settings menu with the rubberuck extension: 

grafik

@wombat6 perhaps you could test adding a tab in the wallet menu?

docs/extensions.md
```diff
@@ -248,19 +248,70 @@ Unfortunately, the two unencrypted classes are derived from the encrypted one ra
...
+```
+
+In order for this to work, you need to define `setting_exts`. Those are the tabs from all the extensions which are contributing tabs to the settings. So you have to specify them like this in your controller:
+
+```python
+@myext_endpoint.context_processor
+def inject_common_stuff():
+    """Can be used in all jinja2 templates"""
+    setting_exts = app.specter.service_manager.execute_ext_callbacks(
+        callbacks.callback_add_settingstabs
```
wombat6, line 308: This should be

```python
callbacks.add_settingstabs
```

docs/extensions.md
```diff
@@ -248,19 +248,70 @@ Unfortunately, the two unencrypted classes are derived from the encrypted one ra
...
 
 ### `controller.py`
-The minimal url routes for `Service` selection and management. As usualy in Flask, `templates` and `static` resources are in their respective subfolders. Please note that there is an additional directory with the id of the extension which looks redundant at first. This is due to the way blueprints are loading templates and ensures that there are no naming collisions. Maybe at a later stage, this can be used to let plugins override other plugin's templates.
+The minimal url routes for `Service` selection and management. As usual in Flask, `templates` and `static` resources are in their respective subfolders. Please note that there is an additional directory with the id of the extension which looks redundant at first. This is due to the way blueprints are loading templates and ensures that there are no naming collisions. Maybe at a later stage, this can be used to let plugins override other plugin's templates.
+
+### Extending the settings dialog
+You can extend the settings dialog with your own templates. To do that, create a callback method in your service like:
```
wombat6, line 269: It's unclear to me which settings dialog we're talking about. As I'm trying to get something to show up, I'm not sure whether I should be looking at wallet settings or the settings for my extension. Am I just expecting the tabs to change, or am I expecting the contents of one of the existing settings tabs to change? So far I'm not having luck getting something to change and I'm not sure if it's because I'm looking in the wrong place or I just implemented the changes incorrectly.

It would also be helpful to give a concrete example or two as to what sort of thing you might be able to do with this feature.

docs/extensions.md
```diff
@@ -117,18 +117,18 @@ class DiceService(Service):
...
 `isolated_client` should not be used yet. It is determining where in the url-path tree the blueprint will be mounted. This might have an impact on whether the extension's frontend client has access to the cookie used in Specter. Check `config.py` for details.
-`devstatus` is one of `devstatus_alpha`, `devstatus_beta` or `devstatus_prod` defined in `cryptoadvance.specter.services.service`. Each Specter instance will have a config variable  called `SERVICES_DEVSTATUS_THRESHOLD` (prod in Production and alpha in Development) and depending on that, the plugin will be available to the user.
+`devstatus` is one of `devstatus_alpha`, `devstatus_beta` or `devstatus_prod` defined in `cryptoadvance.specter.services.service`. Each Specter instance will have a config variable called `SERVICES_DEVSTATUS_THRESHOLD` (prod in Production and alpha in Development) and depending on that, the plugin will be available to the user.
 
 ## Frontend aspects
 
 As stated, you can have your own frontend with a blueprint. If you only have one, it needs to have a `/` route in order to be linkable from the `choose your plugin` page.
```
wombat6, line 129: Not clear to me what a blueprint is, or why I would want more than one. Is it a UI? What would be different between the blueprints? (Theme/design? Platform? Features?)

docs/extensions.md
wombat6: Ah, I finally figured it out by just manually typing the route in the browser. It's the top-level setting dialog. Definitely suggest making that clearer, and maybe including a small screenshot to help the developer understand what they're doing.

Can you clarify the relationship between the settings I would define in the global settings vs. the extension settings? Is the new settings tab in the global settings also for configuring the extension?

wombat6 (changes requested):
OK, it took me a while to interpret exactly what I was doing here, but now I get it. I started out adding a tab to the global settings just because that's what the instructions were describing. Now I understand that what you were asking me to do was basically the same thing, but with the wallet tabs. I assume the approach is exactly analogous, so I'll give that a shot too. Anyway I have a fair amount of feedback for the settings tab too.

src/cryptoadvance/specter/templates/wallet/components/wallet_menu.jinja
```diff
@@ -6,12 +6,17 @@
...
-{% macro wallet_menu(active_menuitem, wallet_alias) -%}
+{% macro wallet_menu(active_menuitem, wallet_alias, ext_wallettabs) -%}
 	<nav class="row collapse-on-mobile">
 		{{ wallet_menu_item('history', _('Transactions'), wallet_alias, active_menuitem, isLeft=true, id='btn_transactions') }}
 		{{ wallet_menu_item('addresses', _('Addresses'), wallet_alias, active_menuitem) }}
 		{{ wallet_menu_item('receive', _('Receive'), wallet_alias, active_menuitem, id='btn_receive') }}
 		{{ wallet_menu_item('send', _('Send'), wallet_alias, active_menuitem, id='btn_send') }}
+		{% for ext_id, wallettabs in ext_wallettabs.items() %}
+			{% for wallettab in wallettabs %}
+				{{ wallet_menu_item(wallettab["endpoint"], _(wallettab["title"]), active_menuitem, isRight=false, blueprint=ext_id+"_endpoint") }}
```
wombat6, line 17: Are you missing `wallet_alias` here?

wombat6: Yeah, it's missing. Once I added it, I was able to get the tab to render as active.

docs/extensions.md
wombat6: Ah, I see that this is a Flask concept.

src/cryptoadvance/specter/templates/wallet/components/wallet_menu.jinja
wombat6: Also, after clicking on my wallet tab, the wallet alias comes in as a request parameter rather than coming in on the request path. So when I click back on one of the other tabs, the wallet alias is no longer part of the request path, and I get the "Wallet history/addresses/receive/send does not exist" alert.

I haven't looked into the cause but I assume that the menu item URLs are being constructed using relative paths.

docs/extensions.md
```diff
@@ -2,7 +2,7 @@
 
 A developer's guide for the Specter Desktop `Extension` framework. 
 
-We currently rework the naming of extensions/plugins/services. If not otherwise stated, you can see those three terms as the same, for now.
+We currently rework the naming of extensions/plugins/services. If not otherwise stated, for now, they are used interchangeably.
 
 ## TL;DR
```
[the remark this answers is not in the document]
k9ert: see #e165afe265876f24be87dec9e94e85ad2340ded6

docs/extensions.md
```diff
@@ -248,19 +248,70 @@ Unfortunately, the two unencrypted classes are derived from the encrypted one ra
...
 In order to separate the service-configuration from the main-configuration, you can specify your config in a file called `config.py`. It's structure is similiar to the specter-wide `config.py`, e.g.:
-```
+```python
 class BaseConfig():
     SWAN_API_URL="https://dev-api.swanbitcoin.com"
 
 class ProductionConfig(BaseConfig):
     SWAN_API_URL="https://api.swanbitcoin.com"
 ```
 In your code, you can access the correct value as in any other flask-code, like `api_url = app.config.get("SWAN_API_URL")`. If the instance is running a config (e.g. `DevelopmentConfig`) which is not available in your service-specific config (as above), the inheritance-hirarchy from the mainconfig will get traversed and the first hit will get get configured. In this example, it would be `BaseConfig`.
```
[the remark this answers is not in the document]
k9ert: committed in #1887 (as i've splitup the docu there and don't want to get too horrible merge-conflicts)

docs/extensions.md
k9ert: clarified, committed in #1887 (as i've splitup the docu there and don't want to get too horrible merge-conflicts)

src/cryptoadvance/specter/templates/wallet/components/wallet_menu.jinja
k9ert: > Also, after clicking on my wallet tab, the wallet alias comes in as a request 
> parameter rather than coming in on the request path.

Not sure why that is. I've added the `wallet_alias` here as well and i've reworked the documentation. I could get rid of the `inject_common_stuff` method and so we only have three modifications left:
* adding the callback_add_-method
* adding the controller-method
* adding the template
So, should work now.

docs/extensions.md
k9ert: Ok, screenshot added, check the preview:
https://deploy-preview-1648--specter-desktop-docs.netlify.app/extensions

docs/extensions.md
k9ert: Yes, indeed. Fixed

moneymanolis:
Tested with the wallet tabs:

grafik

Click from the active "Secret rubberduck wallet" anywhere else results in an error, no idea where wallet "admin" should come from:

grafik

moneymanolis:
I think ` from ..services.service import callbacks` is missing in `src/cryptoadvance/specter/server_endpoints/wallets.py`

moneymanolis:
Also tested with two (settings-) tabs:

grafik

src/cryptoadvance/specter/templates/settings/components/settings_menu.jinja
```diff
@@ -5,6 +5,8 @@
     Parameters:
     - active_menuitem: Current active tab. Options: 'general', 'bitcoin_core', 'auth'.
     - current_user: The current user.
+    - ext_settingstabs: Dictionary gathered by execute_ext_callbacks() in the service manager. Looks like this:
+    	{'rubberduck': [{'title': 'Special rubberduck settings', 'endpoint': 'special_rubberduck_settings'}]}
```
k9ert, line 9: Thanks, pretty good point!

Issue #1661, opened 2022-04-08
Active Node is not User Aware


k9ert:
Currently, the active node which is managed by the [NodeManager](https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/managers/node_manager.py#L96-L97) is valid for all users. This means that using more than one node and more than one user at a time is practically not usable.

We need to make the node-configuration an attribute of the session. Making that needs careful consideration which Managers are User/Session aware and which are not.

k9ert:
@skwp making you aware of this issue. Key sentence here is:

> This means that using more than one node and more than one user at a time is practically not usable.

Fixing that is probably a bigger change. The description also revealed a part of the issue: The stupid check/update architectural pattern: Calling that centralized method every now and then which is effectively then recursively resetting subsystems in order ... you know ... "just to be sure".
We need to get rid of this and e.g. in #1920 we cut it a bit down but it still is one of the biggest issues we have in specter.

PR #1667, opened 2022-04-12
Chore: Refactor_wallet_manager

k9ert:
This is a preparation of making the `WalletManager ` aware of the active node via a session-attribute.

relates to #1661


k9ert:
The experimental session-feature used in the `spec_multiuser_sessions.js` from cypress doesn't seem to work for us. I've created:
https://github.com/cypress-io/cypress/issues/21138

k9ert:
This is ready for review. I'd like to get this in part by part. I could continue here but i think it's better to merge it as is.

cypress/integration/spec_multiuser_sessions.js
```diff
@@ -0,0 +1,92 @@
+describe('Test the actions in UTXO list', () => {
```
moneymanolis, line 1: This is probably from copy&pasting, change this to provide more info on what the test is about.

cypress/integration/spec_multiuser_sessions.js
```diff
@@ -0,0 +1,92 @@
...
+    // See https://github.com/cypress-io/cypress/issues/21138
+    if (Cypress.env("CI")) {
+        const login = (name,password) => {
```
moneymanolis, line 4: Perhaps I am overlooking sth. but where is the name variable set?

moneymanolis (approved):

djpnewton:
So this does not yet make the wallet manager session aware? It is just refactoring to use less instance variables such that when the wallet manager is session aware things like the `wallets_update_list` dict will not contain wallets from multiple different sessions?

k9ert:
> So this does not yet make the wallet manager session aware? 
No. It's also not planned to make it session-aware.

> It is just refactoring to use less instance variables such that when the wallet manager is 
> session aware things like the wallets_update_list dict will not contain wallets from multiple different sessions?

This was just a side_effect of that change. The main point from my point of view is that the `WalletManager` no longer throws away all it's data when the Node gets changed. This is a prerequisite to requesting data from an instance which is session-aware.

cypress/integration/spec_multiuser_sessions.js
k9ert: It's defining a function and the name is the parameter here. It's initialized with the arguments at call time, e.g. in line 39.

PR #1670, opened 2022-04-20
Bugfix: Fees for TXs with two recipients if send-max is placed on wrong recipient fixes #1663

### LINKED ISSUE #1663: Send max with multiple recepients
When I send money to several addresses and click "Send max" not on the first one, Specter subtracts fee from the first output.

Expected behaviour: fee should be subtracted from the output where I clicked "Send max".
Previously we had an input that was specifying which output to subtract from, not sure why it was removed.

image

image

devesh-002:
I believe the issue was with visibility and I have tried to fix it by this PR.

Any reviews would be helpful.
#1663

### DIFF
--- a/src/cryptoadvance/specter/templates/includes/fee-selection.html
+++ b/src/cryptoadvance/specter/templates/includes/fee-selection.html
@@ -14,7 +14,7 @@
                 </p>
             </div>
         </div>
-        <span id="subtract_from" style="display: none">
+        <span id="subtract_from" style="display: none" >
             <br>{{ _("Subtract from recipient number:") }} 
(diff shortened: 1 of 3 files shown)


k9ert:
I tested it and it doesn't work for me. It seems that you made some unrelated changes and also screwed up so formatting. Can you please check again and tidy it up so that you only have changes which are related to fixing the bug?

Hint:
@moneymanolis  mentioned a field which got missing. Here is a screenshot from `v1.7.2`:
image
The field seem to be no longer there. It got updated with the correct field if you click "send max".

k9ert:
Closing this for now. Don't hesitate to reopen if you want to continue working on this.

PR #1683, opened 2022-04-27
Fix: Button alignment issue 

b30wulffz:
# Description
- Remove float property used for aligning the buttons, which was causing this issue
- Switch to flexbox for responsive alignment
- Remove hard-coded right alignment, instead centered the buttons to achieve same results

# Visuals
space-fix-specter

# TODO
- [x] Should I handle the same case for mobile layout also? As of now, these buttons will disappear in the mobile window, since display will be set to "none" by optional class.

### DIFF
--- a/src/cryptoadvance/specter/templates/includes/tx-row.html
+++ b/src/cryptoadvance/specter/templates/includes/tx-row.html
@@ -107,18 +107,18 @@
         </td>
         <td id="column-time" class="time"></td>
         <td id="column-confirmations">
-            <div style="display: flex; align-items: center;">
+            <div style="display: flex;">
                 <span class="confirmations"></span>
-                <div style="flex:1; display: flex; justify-content: center; align-items: center; flex-wrap: wrap;">
-                    <button class="rbf btn optional hidden" style="width: 130px; flex-grow: 0; margin: 5px;" type="button">{{ _("Speed up") }}</button>
-                    <button class="rbf-cancel danger btn optional hidden" style="width: 130px; flex-grow: 0; margin: 5px;" type="button">{{ _("Cancel transaction") }}</button>
-                </div>
             </div>
         </td>
         <td class="hidden optional blockhash"></td>
         <td id="column-action" class="highlight-box">
             <input class="select-tx-value" type="hidden" value="">
             <img style="vertical-align: middle;" class="select-tx-img" src="{{ url_for('static', filename='img/checkbox-untick.svg') }}" width="25px">
+            <div style="display: flex; justify-content: center; align-items: center; flex-wrap: wrap;">
+                <button class="rbf btn optional hidden" style="width: 130px; margin: 2px;" type="button">{{ _("Speed up") }}</button>
+                <button class="rbf-cancel danger btn optional hidden" style="width: 130px; margin: 2px;" type="button">{{ _("Cancel transaction") }}</button>
+            </div>
         </td>
     </tr>
 </template>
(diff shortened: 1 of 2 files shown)


b30wulffz:
fixes #1467 : alignment bug

src/cryptoadvance/specter/templates/includes/tx-row.html
```diff
@@ -106,10 +106,14 @@
             <span class="amount-price note hidden">()</span>
         </td>
         <td id="column-time" class="time"></td>
-        <td id="column-confirmations" >
-            <span class="confirmations"></span>
-            <button class="rbf btn optional hidden" style="width: 130px; float: right;" type="button">{{ _("Speed up") }}</button>
-            <button class="rbf-cancel danger btn optional hidden" style="width: 130px; float: right; margin-right: 10px;" type="button">{{ _("Cancel transaction") }}</button>
+        <td id="column-confirmations">
```
moneymanolis, line 109: Looks already really good, just played around in the firefox dev tools and those little changes appeared to be a nice fine-tuning, what do you think @b30wulffz?

grafik

b30wulffz: I tested this, both designs look nice to me. Can't decide.
image


Also, I feel that we can avoid margin-right, until very much required. I get an OCD  xD, as it looks like we are disturbing the  central alignment of the div beside it.
We are doing this because last td in table for each row, is empty and thus it looks like a block before that is misaligned a bit.

But what if last td gets filled, then it will be a little misaligned.

b30wulffz: Also, spotted a bug, we need to supply a fixed width of confirmation "first" container, else button's positions will get disturbed if buttons are open for multiple transactions.

moneymanolis:
With the rbf buttons in the action column it would look like below.

grafik

src/cryptoadvance/specter/templates/includes/tx-table.html
```diff
@@ -939,7 +939,14 @@ <h1>{{ _("Export transactions to CSV") }}</h1>
                 } else {
                     pending_psbts = []
                 }
-                for (let txRow of this.shadowRoot.querySelectorAll('tx-row')) { 
+                for (let txRow of this.shadowRoot.querySelectorAll('tx-row')) {
```
moneymanolis, line 942: I've added a check here to ensure that the rbf buttons are hidden on the UTXO overview. See comment for why this is not strictly necessary but still more future proof in case the UTXO api endpoint return values change.

src/cryptoadvance/specter/templates/includes/tx-row.html
```diff
@@ -106,15 +106,19 @@
...
+            <div style="display: flex;">
+                <span class="confirmations"></span>
+            </div>
         </td>
         <td class="hidden optional blockhash"></td>
         <td id="column-action" class="highlight-box">
```
b30wulffz, line 115: I have a small doubt, isn't this highlight-box supposed to show some specific element? If so what is it, and when that element is visible, will it affect our current button's positions?

b30wulffz, line 115: I have a small doubt, isn't this highlight-box supposed to show some specific element? If so what is it, and when that element is visible, will it affect our current button's positions?

moneymanolis: We don't need this div / flex container around the span element anymore, I've just removed it.
As far as I can see, the action column wasn't used under the "History" or txlist tab and for the UTXO tab we use the column for the checkboxes. So, I don't see any conflicts.

k9ert (approved):

Issue #1635, opened 2022-03-25
Jade not supported in some cases with hwi-2.0.2


k9ert:
Solution is to upgrade to hwi-2.1.0
see https://github.com/cryptoadvance/specter-desktop/issues/1480#issuecomment-1076837538 for details.

evliu:
Underlying cause is pyserial 3.4 cannot detect `pid` and `vid` but pyserial 3.5 can. Specter can still interact fine with the newer Jades if the check in the `if` statement is removed for the correct `comport` in https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/devices/hwi/jade.py#L865

If HWI is updated, then python 3.10 can be used, but it will also require updates to conform to the newer HWI API across all current specter-supported hardware code (I'm guessing everything in `src/cryptoadvance/specter/devices`

k9ert:
Fixed

PR #1693, opened 2022-05-02
migrating to hwi 2.1.0

k9ert:
Jade is currently not fully supported, see #1635 and this get fixed via migrating to hwi 2.1.0

### DIFF
--- a/tests/test_hwi_rpc.py
+++ b/tests/test_hwi_rpc.py
@@ -1,0 +1,27 @@
+import logging
+from cryptoadvance.specter.hwi_rpc import HWIBridge
+from cryptoadvance.specter.util.descriptor import Descriptor
+
+
+def test_enumerate(caplog):
+    caplog.set_level(logging.DEBUG)
+
+    hwi = HWIBridge()
+    # bla = hwi.detect_device()
+
+    res = hwi.enumerate(passphrase="")[0]
+
+    # seems to be normal
+    assert res["type"] == "trezor"
+    assert res["model"] == "trezor_1"
+    assert res["path"] == "webusb:003:1:1:1:2"
+    assert res["error"].startswith(
+        "Could not open client or get fingerprint information: Passphrase needs to be specified before the fingerprint information can be retrieved"
+    )
+    assert res["code"] == -12
+    assert res["fingerprint"] == "1ef4e492"
+
+    results = hwi.extract_xpubs(chain="test", device_type="trezor").split("\n")
+    assert len(results) == 9
+    assert results[0].startswith("[")
+    assert results == None
--- a/requirements.in
+++ b/requirements.in
@@ -8,9 +8,8 @@ Flask-Cors==3.0.10
 Flask-Login==0.5.0
 Flask-RESTful==0.3.9
 Flask-HTTPAuth==4.4.0
-hwi==2.0.2
+hwi==2.1.0
 importlib_metadata==4.8.1
-pyserial==3.4
 python-dotenv==0.13.0
 requests==2.26.0
 pysocks==1.7.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -40,13 +40,15 @@ base58==2.1.0 \
     --hash=sha256:171a547b4a3c61e1ae3807224a6f7aec75e364c4395e7562649d7335768001a2 \
     --hash=sha256:8225891d501b68c843ffe30b86371f844a21c6ba00da76f52f9b998ba771fb48
     # via bitbox02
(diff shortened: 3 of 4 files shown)


tests/test_hwi_rpc.py
```diff
@@ -1,27 +1,77 @@
...
+
+    hwi = HWIBridge()
+    # bla = hwi.detect_device()
+
+    res = hwi.enumerate(passphrase="")[0]
+
+    # seems to be normal
+    assert res["type"] == "jade"
+    assert res["model"] == "jade"
+    assert res["path"] == "/dev/ttyUSB0"
```
evliu, line 65: when I use pyserial or hwi or blockstream green, my newer jade always shows this path:
```
> ./hwi enumerate
[
  {
    "type": "jade",
    "model": "jade",
    "path": "/dev/cu.usbserial-0224A7E3",
    "needs_pin_sent": false,
    "needs_passphrase_sent": false,
    "fingerprint": "..."
  }
]
```

relativisticelectron:
I found in https://github.com/cryptoadvance/specter-desktop/pull/1688 that  the eerror message has changed and I need to replace https://github.com/cryptoadvance/specter-desktop/blob/b5bf6f19d2014c253651899fc44d1adf45dc631c/tests/test_hwibridge.py#L114
by
```
    assert {
            'error': {
                'code': -32000,
                'message': "Internal error: HWIBridge.enumerate() got an unexpected keyword argument 'non_existing_parameter'."
            },
            'id': 1,
            'jsonrpc': '2.0'
            } == json.loads(req.data)
```

Do I assue correctly that this error is due to the newer version of hwi?

k9ert:
> Do I assue correctly that this error is due to the newer version of hwi?

I would assume so.

tests/test_hwi_rpc.py
k9ert:
```
[
  {
    "type": "jade",
    "model": "jade",
    "path": "/dev/ttyUSB0",
    "needs_pin_sent": false,
    "needs_passphrase_sent": false,
    "fingerprint": "...."
  }
]
```
This is mine, an older one. I don't think that the path is specific to the version. But i think these two numbers are:
```
$ ls -l /dev/ttyUSB0                 
crw-rw----+ 1 root plugdev 188, 0 Mai 18 14:35 /dev/ttyUSB0
$
```
(188 / 0).

k9ert:
Ok, so i don't have a ledger and for some reason my old keepkey doesn't work anymore for testing. I would love to at least test the keepkey, though as i removed the keepkey specific code which is no longer needed.
Anyone with a keepkey can test that?

k9ert:
I didn't get any feedback from keepkey folks. Also no feedback from their support in order to get my keepkey back to work. I'm a bit annoyed, i have to say.

evliu:
looks like hwi 2.1.1 has been released with small updates. might as well jump ahead
https://github.com/bitcoin-core/HWI/releases/tag/2.1.1

k9ert:
Doesn't look like it get any better with Keepkey. Not sure about HWI 2.1.0. I'll definitely move this forward this week.

k9ert:
Ok, i think we should merge this and maybe shortly after this is released upgrade to `HWI 2.1.1`. I've tested it with `2.1.0`, it has some major changes on our side and it's beneficial to ask people if they have issues to switch versions in order to figure out where an issue is coming from.

I think i've discussed the details of this with @moneymanolis and it seems that @relativisticelectron looked into it but we don't have a formal approval yet. Can one of you approve this?

relativisticelectron:
I am surprised you did not run into the seemingly changed error message:
https://github.com/cryptoadvance/specter-desktop/pull/1693#issuecomment-1119309462



> I think i've discussed the details of this with @moneymanolis and it seems that @relativisticelectron looked into it but we don't have a formal approval yet. Can one of you approve this?

Sorry, I did not test hwi 2.1.0. However I could test hwi 2.1.0 with Ledger, as soon as I have time.

k9ert:
> I am surprised you did not run into the seemingly changed error message:

I spent some time trying to understand why that might be fut failed. No idea. But i also can't see the relevance for merging this.

> Sorry, I did not test hwi 2.1.0. However I could test hwi 2.1.0 with Ledger, as soon as I have time.

I don't think Ledger is that critical. This PR explicitely changed a lot for keepkey and Trezor and enabled the new jade's. So keepkey testing would be more important.

I'll merge that tomorrow and will create a ReleaseCandidate so that people from the chat can potentially test the ledger and you don't have to. There are some folks in there who might be able to help here (again).

PR #1696, opened 2022-05-03
Docker

AaronDewes:
This adds Docker builds to the repo, based on the lncm container


relativisticelectron:
Wouldn't it be nicer if the Dockerfile would be in a folder like `docker/github-workflow` ?

k9ert:
All the images in the docker subfolder are supporting images. They are not containing specter itself. I think there is the convention that if you have only one image which is representing the app, it should be directly in the main folder of the project. For the reference, we had that discussions about that also here: https://github.com/cryptoadvance/specter-desktop/pull/480#issuecomment-706127293

PR #1706, opened 2022-05-06
Feature: Optimized uncle jim pdf for the Debasafebags.

f9ert:
image


moneymanolis:
The displayed character length of 42 in the pdf for the addresses is fine: P2WPKH addresses are always 42 characters long (for more details see here: https://bitcoin.stackexchange.com/questions/100508/can-you-break-down-what-data-is-encoded-into-a-bech32-address). With regtest addresses we are adding two characters ("rt"), thus two characters are missing then, just keep this in mind. Perhaps it makes sense to add this non-trivial background info to the tool-tip belonging to the pdf, @k9ert?

src/cryptoadvance/specter/templates/wallet/components/wallet_pdf.jinja
```diff
@@ -174,47 +174,47 @@
...
             doc.text({{ wallet.name | tojson | safe }}.replace(/(.{55})/g, "$1\n"), 105, currentHeight, {align: 'center'});
             doc.setFontSize(9);
-            doc.text('{{ _("Scan this QR code with Specter-Desktop") }}', 65, currentHeight + 6, {align: 'left'});
-            doc.text("The below addresses are from index 1. ", 65, currentHeight + 12)
-            doc.text("Feel free to use the checkbox as a proof", 65, currentHeight + 18)
-            doc.text("to have verified the address via the Hardwarewallet.", 65, currentHeight + 24)
+            doc.text('{{ _("Scan this QR code with Specter-Desktop") }}', 50, currentHeight + 6, {align: 'left'});
+            doc.text("The below addresses are from index 1. ", 50, currentHeight + 12)
+            doc.text("Feel free to use the checkbox as a proof", 50, currentHeight + 18)
+            doc.text("to have verified the address via the Hardwarewallet.", 50, currentHeight +24)
```
moneymanolis, line 190:
```suggestion
            doc.text("to have verified the address via a hardware wallet.", 50, currentHeight +24)
```

src/cryptoadvance/specter/templates/wallet/components/wallet_pdf.jinja
```diff
@@ -174,47 +174,47 @@
...
-            doc.addImage(qrCode, 'PNG', 5, currentHeight, 50, 50)
+            doc.addImage(qrCode, 'PNG', 5, currentHeight, 30, 30)
             doc.text({{ wallet.name | tojson | safe }}.replace(/(.{55})/g, "$1\n"), 105, currentHeight, {align: 'center'});
             doc.setFontSize(9);
-            doc.text('{{ _("Scan this QR code with Specter-Desktop") }}', 65, currentHeight + 6, {align: 'left'});
-            doc.text("The below addresses are from index 1. ", 65, currentHeight + 12)
-            doc.text("Feel free to use the checkbox as a proof", 65, currentHeight + 18)
-            doc.text("to have verified the address via the Hardwarewallet.", 65, currentHeight + 24)
+            doc.text('{{ _("Scan this QR code with Specter-Desktop") }}', 50, currentHeight + 6, {align: 'left'});
+            doc.text("The below addresses are from index 1. ", 50, currentHeight + 12)
```
moneymanolis, line 188:
```suggestion
            doc.text("The addresses below start from index 1. ", 50, currentHeight + 12)
```

k9ert:
> Perhaps it makes sense to add this non-trivial background info to the tool-tip belonging to the pdf, @k9ert?

Great finding! Maybe it also make sense to print it directly on the pdf but just if the network is regtest. Do we also have char-loss for testnet?

moneymanolis:
No, testnet is fine, too, it's just tb:

image

moneymanolis:
I think I would only enable the Uncle Jim PDF feature if the node is set to mainnet, it only poses risks to send to testnet addresses (if the wallet used is not validating correctly ..., see: https://bitcoin.stackexchange.com/questions/18028/what-happens-if-you-send-btc-to-a-testnet-address)

k9ert:
> I think I would only enable the Uncle Jim PDF feature if the node is set to mainnet,

Makes development and testing out stuff a lot more complicated. Don't like the idea so much. I'll try to print it at the bottom of the page...

src/cryptoadvance/specter/templates/wallet/components/wallet_pdf.jinja
```diff
@@ -242,8 +242,14 @@
                 xpubLines = "{{ key.xpub }}".split(/(.{36})/g);
                 xpubLines = xpubLines.filter(line => line !== '');
                 currentHeight += 6
+                doc.setLineDash([2, 2], 0);
+                doc.line(181,0,181,214)
+                doc.line(0,214,181,214)
```
k9ert, line 247: You're drawing those lines for every key in the wallet. Please move it out of the loop!

k9ert (approved):

PR #1707, opened 2022-05-06
Fix: Change specter compatibility to ">=3.7,<3.10" due to dependency

relativisticelectron:
The specter requirements include
https://github.com/cryptoadvance/specter-desktop/blob/8e6c374c110aca9639aa4b6cc3f3e2d5b0605a6e/requirements.in#L3
which has the requirement  https://pypi.org/project/click/8.1.1/
image

This then requires for specter also to have python>= 3.7

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,7 +61,7 @@ setup(
         "Operating System :: OS Independent",
         "Framework :: Flask",
     ],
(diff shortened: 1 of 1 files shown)


moneymanolis:
Sounds logical to me.

k9ert (approved):

PR #1716, opened 2022-05-11
Chore: Adjusted releasing process

k9ert:
In order to be able to ship specter extensions which are located outside the specter-desktop repo, the release-process needed to adjustements refactoring it on the way. There won't be any more separate CI-scripts but options on the `build-(linux|macos|win)` scripts.

This is still work in progress as the Macos build is not yet completed. The linux/win is finished, though. Test Artifacts can be found at https://github.com/k9ert/specter-desktop/releases
* No longer using `pip3 install -e .` for an environment which is used for the release. So we're now using `.buildenv` to generate a virtualenv, capable for releasing


src/cryptoadvance/specter/util/reflection.py
```diff
@@ -74,7 +74,21 @@ def get_classlist_of_type_clazz_from_modulelist(clazz, modulelist):
     """
     class_list = []
     for fq_module_name in modulelist:
-        module = import_module(fq_module_name)
+        try:
+            module = import_module(fq_module_name)
+        except ModuleNotFoundError as e:
+            # ToDo: make it somehow clear where specific extensions are coming from: external or within same repo
+            # Then we can remove the hardcoding of the first extra-repo extension: liquidissuer
```
k9ert, line 81: comment does not fit anymore. The hardcoding is already removed.
```suggestion
```

moneymanolis:
Change PR title to include liquid issuer addition

moneymanolis:
Tested by running v1.10.0-pre23 MacOS build - works fine.

moneymanolis (approved):

PR #1719, opened 2022-05-13
Bugfix: wrong arguments for requests_session

k9ert:
You cannot call it with the explorer as argument. Instead, we're now checking within the UtxoScanner for inconsitent arguments.


k9ert:
Two things are still missing in this PR:
* [ ] A user choosing a tor-onion-address for the explorer should get a tor-capable session
* [ ] A user who does the above but doesn't have a reasonable tor-setup, needs to get a reasonable error-message

k9ert:
Two things are still missing in this PR:
* [ ] A user choosing a tor-onion-address for the explorer should get a tor-capable session
* [ ] A user who does the above but doesn't have a reasonable tor-setup, needs to get a reasonable error-message

k9ert:
The second one was mitigated by reasonable error-logging.

PR #1729, opened 2022-05-23
Chore: non-standard ports for test-cypress.sh

k9ert:
It's beneficial to have a regtest-server running on the default-port but still be able to run the tests in parallel.
This PR makes that possible.


src/cryptoadvance/specter/cli/cli_noded.py
```diff
@@ -132,8 +132,8 @@ def bitcoind(
 )
 @click.option(
     "--port",
-    default=7040,
-    help="Specify a port the elementsd should run on. Default is 7040.",
+    default=18555,
```
moneymanolis, line 135: Is this really the default? 
Any changes to the default liquid RPC ports should also be reflected in the front end:
grafik

moneymanolis (approved):
Works! Great PR, makes working with the Cypress tests a LOT easier. Now, we can run Cypress and a Specter instance in parallel, each with it its own bitcoind instance. 🎉

moneymanolis:
@relativisticelectron just tagging you to make you aware of this.

src/cryptoadvance/specter/cli/cli_noded.py
k9ert: The default in the cli is not the default in specter. That's imho exactly the point because otherwise, you still have conflicts.
You should be able to long-run your (elements-) node how you would like. If you deal with the cli, it should not interefere, especially not with its default parameters.
Makes sense?

moneymanolis: No, it doesn't really make sense to me tbh. We've settled on default ports for elementsd and those should be reflected here as well. Non-default ports to avoid conflicts when running Cypress and Specter with bitcoind/elementsd are handled in test-cypress.sh.

PR #1730, opened 2022-05-24
Basic auth in electron

djpnewton:
add a 'HTTP Basic authentication' setting in electron so that a user can access their remote specter instance that is protected by http basic auth

WIthout this PR it is not possible to access a remote specter server behind a proxy server that uses basic auth (as documented in the reverse proxy guide https://github.com/cryptoadvance/specter-desktop/blob/master/docs/reverse-proxy.md#adding-basic-authentication-via-nginx)

Some questions:
 - would it be better to not have a setting in preferences and instead popup a dialog box like a standard browser ("The server at https://XXX requires a password...")?
 - do we want to save the session data so that a user does not have to repeat the password for specter in built auth when restarting the app?


k9ert:
I'm not sure i understand why this is necessary. You can switch on one of the three auth methods in specter. Doesn't that do the trick?

djpnewton:
If the server you are connecting to has basic auth then this is the result with no obvious way for the user to fix it:
image

Users might want to use basic auth for a extra layer of security and/or to be able to use a service like fail2ban in order to block IPs that are using a brute force attack

k9ert:
Ah, ok, now i got it.

djpnewton:
added a popup for basic auth if missing or incorrect:
image

k9ert:
Thank you very much!

Issue #1710, opened 2022-05-09
build-unix.sh  failes in Ubuntu 22.04


relativisticelectron:
**Describe the bug**
Running `cd pyinstaller` and `./build-unix.sh`  gives the error message:
```
Traceback (most recent call last):
  File "/home/<user>/programs/anaconda3/envs/specter/lib/python3.8/site-packages/PyInstaller/building/build_main.py", line 464, in assemble
    ctypes_binaries = scan_code_for_ctypes(co)
  File "/home/<user>/programs/anaconda3/envs/specter/lib/python3.8/site-packages/PyInstaller/depend/utils.py", line 110, in scan_code_for_ctypes
    binaries = __recursively_scan_code_objects_for_ctypes(co)
  File "/home/<user>/programs/anaconda3/envs/specter/lib/python3.8/site-packages/PyInstaller/depend/utils.py", line 172, in __recursively_scan_code_objects_for_ctypes
    libname = ctypes.util.find_library(libname)
  File "/home/<user>/programs/anaconda3/envs/specter/lib/python3.8/ctypes/util.py", line 351, in find_library
    _get_soname(_findLib_gcc(name) or _findLib_ld(name))
  File "/home/<user>/programs/anaconda3/envs/specter/lib/python3.8/ctypes/util.py", line 148, in _findLib_gcc
    if not _is_elf(file):
  File "/home/<user>/programs/anaconda3/envs/specter/lib/python3.8/ctypes/util.py", line 100, in _is_elf
    with open(filename, 'br') as thefile:
FileNotFoundError: [Errno 2] No such file or directory: b'liblibsecp256k1.a'

```

**Expected behavior**
It should find the library and continue with building a binary.  **In Ubuntu 20.04 it builds fine.**  

If I go into "/home/<user>/programs/anaconda3/envs/specter/lib/python3.8/ctypes/util.py" and catch the library name and change it to b'libsecp256k1.a' it can find it and continues flawlessly with the built.

_Could this be a Pyinstaller bug?  Or embit?_

**Desktop (please complete the following information):**
 - Ubuntu 22.04, Pyinstaller 4.9
 - Specter Version: current master

relativisticelectron:
In Ubuntu 20.04  
![image](https://user-images.githubusercontent.com/60378539/169772257-eb8c1c1f-0623-4889-a2ae-eb9e7594388a.png)
[...]

PR #1735, opened 2022-05-28
Corrected readme

relativisticelectron:
The build only works for virtuelenv and a specfic env name (see https://github.com/cryptoadvance/specter-desktop/issues/1710#issuecomment-1140179748 ). 

- Updated the docs to give correct instructions.
- Made that the build env can have any name


k9ert:
Good point. Works great. Will merge soon.

relativisticelectron:
Thanks.

PR #1741, opened 2022-06-01
Bugfix: PDF backup does not open for multisig

### LINKED ISSUE #1738: Export wallet's PDF backup doesn't open in current `master` branch
**Describe the bug**
Cannot open a wallet's backup PDF file.

**To Reproduce**
1. Navigate to a wallet (multisig in my case)
2. Click on its Settings tab
3. Click on the Export sub tab
4. Click on "Open backup PDF"

Nothing happens. If you open a console window, there's no error. Viewing the source I don't see anything wired up to the button `id="pdf-wallet-download"`. So I think the button is literally doing nothing when clicked.

**Expected behavior**
Should open the wallet's backup PDF file.

 - Where is your node running: local regtest
 - Where is your specter running: local from current (2022-05-28) `master` branch
 - If you're running a binary, which OS: Not a binary, but on macOS
 - Browser: Chrome 101.0.4951.64
 - Specter Version: n/a (current `master`)

k9ert:
fixes #1738

### DIFF
--- a/src/cryptoadvance/specter/templates/wallet/settings/wallet_settings.jinja
+++ b/src/cryptoadvance/specter/templates/wallet/settings/wallet_settings.jinja
@@ -170,9 +170,9 @@
 				<button id="pdf-paperxpub-download" type="button" class="btn centered padded" style="margin: 30px auto;">
 					Get Uncle Jim PDF
 				</button>
+			{% endif %}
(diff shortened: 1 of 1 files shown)


b30wulffz (approved):

b30wulffz:
lgtm

moneymanolis:
Good catch, the misplaced endif jinja tag was the problem.

Issue #1722, opened 2022-05-15
unncecessary scrolling


k9ert:
**Describe the bug**
Currently, the list of transactions is only 6 TX long. You have to scroll even though you might have lots of space on your screen.
![scrolling](https://user-images.githubusercontent.com/117085/168476060-e5c58601-af98-44a6-8633-6b75893f13c3.gif)

ankur12-1610:
I would like to work on it😅

ankur12-1610:
@k9ert I was planning to it this way - 

https://user-images.githubusercontent.com/76884959/168595187-f3ca5c94-2d8b-4445-8905-7e4ef12c77dd.mp4

Or do we need a scrollable table?

moneymanolis:
@ankur12-1610
It was a user request to implement a scrollable table, mainly to keep this part of the table (`<div class="search-container">`) visible when scrolling down:
<img width="835" alt="image" src="https://user-images.githubusercontent.com/70536101/168836873-bd162d58-5fcb-4472-a058-fbedd7e95ad3.png">

For now, the requirement would be to keep that functionality.
I think one way to make the height of the table-container more adaptive would be by setting the height of the table container using the viewport's (visible page area) height (using a percentage of the viewport height). See here: https://stackoverflow.com/questions/39190742/css-fill-height-of-screen-when-window-is-zoomed-100

ankur12-1610:
> @ankur12-1610 It was a user request to implement a scrollable table, mainly to keep this part of the table (`<div class="search-container">`) visible when scrolling down: <img alt="image" width="835" src="https://user-images.githubusercontent.com/70536101/168836873-bd162d58-5fcb-4472-a058-fbedd7e95ad3.png">
> 
> For now, the requirement would be to keep that functionality. I think one way to make the height of the table-container more adaptive would be by setting the height of the table container using the viewport's (visible page area) height (using a percentage of the viewport height). See here: https://stackoverflow.com/questions/39190742/css-fill-height-of-screen-when-window-is-zoomed-100
> 
> As a side note, I would also think about changing the default page limit from 50 to sth. lower.

Sure, working on it :)

I was thinking to set the default page limit to 25, does that sound right?

PR #1746, opened 2022-06-05
UIUX: Improving the tx table scrollbar

moneymanolis:
### Contents
- Using viewport height for the table container in scrollbar mode which on its own should already fix https://github.com/cryptoadvance/specter-desktop/issues/1722
- In addition: Toggle for tx table scrollbar - so the scrollbar can be turned off, if some unusual screen size and / or zoom is used
- Some styling to differentiate between the tx- / UTXO- list in scrollbar mode and without
- Page limit default set to 25

#### Firefox
https://user-images.githubusercontent.com/70536101/172051329-6374a3c4-560d-40c2-a536-ab277fe71121.mov

#### Chrome
https://user-images.githubusercontent.com/70536101/172052777-8dd022be-105e-44ec-8a46-9712c17598c2.mov

### DIFF
--- a/src/cryptoadvance/specter/templates/includes/tx-table.html
+++ b/src/cryptoadvance/specter/templates/includes/tx-table.html
@@ -175,12 +175,10 @@
         }
         .table-container {
             width: 100%;
-            height: 50vh;
             margin-bottom: 25px;
         }
         .scrollbar {
-            min-height: 350px;
-            overflow-y: scroll;
+            overflow-y: auto;
             scrollbar-color: hsl(217, 10%, 80%) var(--cmap-border); 
             scrollbar-width: thin;
             border-radius: 15px;
@@ -210,6 +208,16 @@
         }
         input:checked + .slider {
             background-color: hsl(217, 30%, 50%);
+        }
+        :host {
+            display: flex;
+            flex-direction: column;
+            overflow: hidden;
+            height: 100%;
+        }
+        .table-container {
+            background: var(--cmap-bg-lighter);
+            height: 100%;
         }
     </style>
 
--- a/src/cryptoadvance/specter/templates/includes/tx-table.html
+++ b/src/cryptoadvance/specter/templates/includes/tx-table.html
@@ -33,7 +33,9 @@
     blockhash="true" 
     hide-switcher="true"
     wallet="{{ wallet.alias }}"
-    scrollbar-height="333px"
+    table-height="333px"
+    table-max-height="333px"
+    table-border-radius="10px"
     scrollbar-default="off"
     scrollbar-switch="true">
 </tx-table>
@@ -181,7 +183,7 @@
             overflow-y: auto;
             scrollbar-color: hsl(217, 10%, 80%) var(--cmap-border); 
             scrollbar-width: thin;
-            border-radius: 15px;
+            border-radius: 10px;
             border: 1.5px solid var(--cmap-border);
         }
         .table-container::-webkit-scrollbar {
@@ -212,13 +214,14 @@
         :host {
             display: flex;
             flex-direction: column;
-            overflow: hidden;
             height: 100%;
         }
+        /* For giving additional height to table which fills the empty space */
+        /*
         .table-container {
             background: var(--cmap-bg-lighter);
             height: 100%;
-        }
+        } */
     </style>
 
     <nav class="switcher row">
@@ -596,7 +599,7 @@
         }
 
         static get observedAttributes() {
-            return ['blockhash', 'btc-unit', 'price', 'symbol', 'type', 'hide-sensitive-info', 'wallet', 'service-id', 'selected-coins', 'scrollbar-height', 'scrollbar-default', 'scrollbar-switch'];
+            return ['blockhash', 'btc-unit', 'price', 'symbol', 'type', 'hide-sensitive-info', 'wallet', 'service-id', 'selected-coins', 'table-height', 'table-max-height', 'table-border-radius', 'scrollbar-default', 'scrollbar-switch'];
         }
 
         connectedCallback() {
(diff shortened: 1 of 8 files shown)


ankur12-1610:
@moneymanolis did we do the same in `Addresses` section?

moneymanolis:
@djpnewton could you let me know what you think, you were also active in the related PR that worked on that (and that I just closed).

moneymanolis:
> @moneymanolis did we do the same in `Addresses` section?

No, I let this section at it is. There is no real need there for any scrollbar there, since there is no functionality that should be kept visible like we have it with the tx list and utxo list.

moneymanolis:
@k9ert does this solve your original issue?

src/cryptoadvance/specter/templates/includes/tx-table.html
```diff
@@ -269,7 +277,7 @@ <h4 slot="title">{{ _('Freezing UTXOs') }}</h4>
             <a id="manage-psbt-btn" type="button" class="btn manage-psbt-btn hidden" href="">{{ _("Manage unsigned PSBT") }}</a>
         </div>
         <div id="scrollbar-checkbox-div" class="scrollbar-checkbox-div row hidden">
-            {{ _("Compact table") }} &nbsp;&nbsp;
+            {{ _("Scrollbar table") }} &nbsp;&nbsp;
```
djpnewton, line 280: `"Scrollbar Table"` is a bit awkward. This would be better just "Scrollbar" perhaps?

moneymanolis: I wanted to make clear, that it only refers to the scrollbar of the table not the general scrollbar.

k9ert: Maybe "Table scrollbar" is better then?

src/cryptoadvance/specter/templates/includes/tx-table.html
```diff
@@ -174,21 +174,26 @@
             border: 1px solid var(--main-color);
         }
         .table-container {
-            width:  100%;
+            width: 100%;
+            height: 50vh;
```
b30wulffz, line 178: Hardcoding 50vh doen't looks like a perfect fix for all displays. There can be a case when the elements above this table are together occupying space << 50vh. Hence again, the empty space the end can be present.
Or it might overflow if display is small.

b30wulffz (commented):
After turning scroll bar switch on and off, the round corners of the table are appearing and disappearing. For better UI/UX, i feel that we can either let it be square or circular in both modes.

k9ert:
> For better UI/UX, i feel that we can either let it be square or circular in both modes.

Marginal improvement. However, i'm not a UI/UX perfectionist ;-)

src/cryptoadvance/specter/templates/includes/tx-table.html
moneymanolis: That is the reason why I chose to also add the toggle for the table scrollbar, so in weird edge cases, people can just deactivate it altogether. Once we have put the user persistence system on a more solid footing and if many users want to have the scrollbar deactivated permanently, we can make the switching off persistent.

Having said this, if you have a better short term solution, let me know!

b30wulffz: What about using flex on the outer div with height 100vh, and then flex:1 on tx table, 
or getting remaining height using javascript?

GBKS:
I'd keep the rounding consistent. It's a small detail that users are very unlikely to even notice.

Overall, I would not have this toggle and also no scroll bar in the table. The page can just expand and get as tall as it needs to be. If there's an element that you want to be visible further down (like a table header with options), using [position: sticky](https://caniuse.com/css-sticky) is a more elegant solution.

moneymanolis:
@GBKS - thanks for your input, we'll give it a shot with the sticky option. It was a feature request by an user to have this scrollbar for the table, but perhaps your solution is indeed more elegant.

b30wulffz:
Hii, I tried working on this PR tweaking the properties with just flexbox. This implementation is independent of providing any custom height, and takes the available screen space. Here are two responsive versions of the same. 

Version 1:

https://user-images.githubusercontent.com/31105595/175104682-f403cbc2-b546-4bfb-a115-80d872468e9f.mp4

Version 2:

https://user-images.githubusercontent.com/31105595/175104708-b081f917-4d60-4fd3-96e3-caf9e0f11206.mp4

Which one looks better? Also, should I keep that toggle now?

src/cryptoadvance/specter/templates/includes/tx-table.html
```diff
@@ -197,12 +202,26 @@
...
         input:checked + .slider {
-            background-color: var(--cmap-border);
+            background-color: hsl(217, 30%, 50%);
         }
+        :host {
+            display: flex;
+            flex-direction: column;
+            height: 100%;
+        }
+        /* For giving additional height to table which fills the empty space */
```
moneymanolis, line 219: This can be deleted. This is done by height: 100vh; in main - no?

moneymanolis:
I've implemented the suggestion of @GBKS (using `position: sticky`) in https://github.com/cryptoadvance/specter-desktop/pull/1746/commits/d7ad2b2f4280af545d889ba5ff0c709359efc58f
It would look like this (I've also tried to make the UX a bit cleaner here and there)
https://user-images.githubusercontent.com/70536101/175778291-c1aa5008-db0b-461a-aec5-cde7ba024ee1.mov

Scrollbar within the table only kept for the coin selection:
https://user-images.githubusercontent.com/70536101/175778793-70f1b075-e72a-4a4d-8c56-2515195b4094.mov

@djpnewton, @b30wulffz, @ankur12-1610, @relativisticelectron - could someone of you guys test it? The sticky option is relatively new and it would be good to check whether the browser support is good enough. Worked fine for me on Firefox and Chrome so far.
@GBKS let me know what you think and if you have an further suggestions.

TODO
- [ ] Check browser support
- [ ] Optional: Fix a little bug with the pagination

PR #1744, opened 2022-06-04
Show and set subtract from field on send max

ahv15:
fixes #1663

PR #1747, opened 2022-06-05
UI Feature: Drag and drop of recipients

relativisticelectron:
I found the create transation dialog very complicated, and aimed to make it better. It uses https://github.com/SortableJS/Sortable. Let me know what you think:

**Todo:**

- [x] Add test for rearranged order
- [x] Keep always 1 recipient
- [x] Added green coloring if the address belongs to the wallet 
- [x] test with narrow screen size 
- [x] Needs more testing to find edge cases
- [x] Add/Change a cypress test that tests the reordering of the inputs AND selecting where to deduct the fee.  I gave up on drag and drop in cypress. This seems impossible, because `getBoundingClientRect` is not available; [see also](https://stackoverflow.com/questions/55361499/how-to-implement-drag-and-drop-in-cypress-test#55436989) . Cypress tests added 
- [x] Change the [dev help text](https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specterext/devhelp/templates/devhelp/html/fee-selection.jinja)
- [x] This will need rebasing after https://github.com/cryptoadvance/specter-desktop/pull/1744 is merged. Implemented as select  
image

**Add, delete, and rearrange recipients**

- own addresses are highlighted in green
- delete any and rearrange freely

Peek 2022-06-15 21-11

**Breaking API changes:**
Changed [`subtract_from` to start from `0`]( https://github.com/relativisticelectron/specter-desktop/blob/f20ca566a2fccdd13d68e1e234067174a44fe06a/src/cryptoadvance/specter/commands/psbt_creator.py#L301), relevant for `paymentinfo_from_json`  see  https://github.com/cryptoadvance/specter-desktop/blob/5f82e2a8f89f5532f2f609f75d8d7fd0152f02bd/src/cryptoadvance/specter/commands/psbt_creator.py#L198 .  Original:
https://github.com/cryptoadvance/specter-desktop/blob/5f82e2a8f89f5532f2f609f75d8d7fd0152f02bd/src/cryptoadvance/specter/commands/psbt_creator.py#L299


relativisticelectron:
@moneymanolis : If you have time, could you look at this?

relativisticelectron:
I added that addresses, which belong to the wallet,  are colored green:
Peek 2022-06-06 11-53

k9ert:
Wow, this looks cool! If @moneymanolis can't do it, maybe @b30wulffz ?!

relativisticelectron:
> Wow, this looks cool! 

Thanks, I think it gives the send dialog a more modern feeling, and hopefully encourages users  to create tx with more than 1 recipient.

b30wulffz:
> Wow, this looks cool! If @moneymanolis can't do it, maybe @b30wulffz ?!

Yeah sure, i'll test this.

relativisticelectron:
I don't understand why cypress fails.

relativisticelectron:
Sorry, I need a little more time to sort out another issue.....

relativisticelectron:
@moneymanolis :  Somehow I am completely lost with how to develop with cypress.  I simply tried copying a test and modifying it:  https://github.com/relativisticelectron/specter-desktop/commit/9a91518d9cb78f91afe46d5a4baf59a82b5c2538  but it doesnt get far
Test sending transactions -- check drag and drop reordering of recipients and check the fee es deducted corrrectly (failed)

moneymanolis:
@relativisticelectron I have to work through some other PRs first and will then focus on this one. No worries about the Cypress tests, I'll fix them (hopefully 😇).

relativisticelectron:
> @relativisticelectron I have to work through some other PRs first and will then focus on this one. No worries about the Cypress tests, I'll fix them (hopefully innocent).

Thank you so much.

k9ert:
@b30wulffz don't hesitate to have a look here as well.

relativisticelectron:
> @relativisticelectron I really like the ease of adding a recipient now, this is excellent! Before, delving deeper into the review, I just wanted to understand better what the actual idea / use case is to drag and drop the recipients / to move them up and down? It adds quite a bit of "stuff" (buttons etc.) which is also not under an advanced toggle anymore, so I want to make sure that it has a purpose (which I perhaps just don't get straight away).

The idea is to make it easier to:
- make adding more recipients more user friendly (good for fees and privacy)  (This was the original motivation, since I wanted to add more recipients and found it quite difficult)
- allow swapping the order of the recipients, e.g. :
  - to allow an obfuscation of recipient and change address (The obfuscation of recipient and change address could be done via other methods (does bitcoin core already do that?))
  - as a visual aid for the user when there are many recipients
  - when for whatever reason the user wants a different order, he doesn't have to delete all recipients and start fresh    

The up/down buttons on the right hand side are not strictly necessary. I added them for cypress, since cypress doesn't seem to support drag&drop in flask. However they could actually be necessary for users with limited mouse control (voice or keyboard control alone), though I didn't test this.

relativisticelectron:
If there is only 1 recipient, one could hide all 3 buttons on the right (via css and only child)
image

moneymanolis:
Yeah, the green colour is nice. For the manual change address to work (and I really like this idea of putting it somewhere in between of many recipients!) the send max functionality must work (I just played around with it and it didn't). You would select a UTXO from coin selection to cover your recipients and then make send max (in this case it would actually be nice for the button text to change to "send rest" or so). 

The only real argument to have drag and drop and the up and down buttons is this IMO (which can be enough of an argument of course, but we'd need to discuss)
> when for whatever reason the user wants a different order, he doesn't have to delete all recipients and start > fresh. There is no real reason to limit the ordering to the "creation order".

There is a lot of potential in this PR but we need to get the UX right. I personally would like to open up the complexity of the batch transaction only to more advanced users otherwise average users will be a bit overwhelmed I think.

relativisticelectron:
> the send max functionality must work (I just played around with it and it didn't

ok, I will fix that.

> I personally would like to open up the complexity of the batch transaction only to more advanced users otherwise average users will be a bit overwhelmed I think.

Would hiding the 3 buttons for css only-child work for you?

relativisticelectron:
> > the send max functionality must work (I just played around with it and it didn't
> 
> ok, I will fix that.

It should be fixed now:
Peek 2022-06-23 06-29

k9ert:
I'm not that sure where the value of this PR is coming from. Why do we want to shuffle the recipients? If it's about privacy, why not randomly shuffle every TX on the backend by default? That would get a better result from a methematical point of view than doing it manually? On the other hand, the `wallet_send` page is already the most complex piece of code we have and its maintainability is currently already not the best. Complicating it even further should be done for a very good reason and i'm struggling a bit to see that reason.

relativisticelectron:
In my opinion, this UI (each recipient is in a visually clearly separated box) is a step forward for creating transactions with 2 or more recipients. Adding recipients should be done as easy as possible  1) to encourage more batched transactions, 2) to allow more complex transactions to be created in the UI instead of text based

After I had the  UI (each recipient is in a box) and added several, allowing the order to be changed was the next logical step, because there is no reason (for the user) to restrict the order. For me the main motivation is a visual aid, similar to the green color being a visual aid.

In hindsight the old UI has not an ideal flow. Example: You could only delete the last recipient. Why? What if I actually want to delete the first of my 10 recipients?  I would have to basically start fresh, or give up on the UI and do it text based.

k9ert:
I agree with most of your points, explicitely:
* the list of recipients should look more like a list of recipients where one is distinguished from the other
* encourage batch txs
* Deletion of any recipient should be possible
But that doesn't answer why shuffling is necessary. Can't we do that without the shuffling UI, keep the green background, separate boxes for recipients and the possibility to delete any Recipient?
I think we also should think hard about how to compose that with webcomponents. As said, the complexity of this is already quite horrible.

relativisticelectron:
I agree this PR should structure, e.g. the recipient boxes, in webcomponents to increase maintainability.   I can start with that as soon as the other details are clear.

I agree that adding dependencies is not ideal.  I think the shuffling is not a crucial part of this PR, if you think the use-cases (right now) are not strong enough.  I would like to see in in Specter sooner or later however. Reasons:

- Adding manual change addresses (1 or more) last and being able to change the order to not have them as the last outputs, I think is a benefit (At least I constructed tx like that in the past, and this shuffling would simplify the tx construction). As you point out a random (or otherwise information removal) process would be better. The only way I can think of presenting such a random shuffling to the user is actually to keep the SortableJS and add a dice 🎲 symbol in the top, so the user can randomize the order. The user would then actually see (via animation), that just the order changed, and nothing else.
- A visual aid: 1. output  to Friend 1,  2. output to Friend 2,  3. output to  parent 1 -->  Ohhh I forgot Friend 3, let me drag it under Friend 2 -->  Ok double check Friend 1,2,3, and Parent 1. Looks good --> Send.  




> * The send-dialog is on a good way to getting to such a component and i'm thinking how to improve the situation here.

2 ideas:  
1. The Transaction editors  **UI** , and  **text** should be separated in completely different tabs to allow independent development and features.
2. UI: Apart from that,  I think the main complexity comes in through the advanced section, that has influence on some of the outputs (e.g. send_max, deduct_fee_here).  A cleaner solution would be to carry these properties (like  send_max, deduct_fee_here) in the recipient webcomponent itself.  Then it will be clearer what interacts with what.

Though the 2 ideas would make this PR too big, and would belong in a different PR.

moneymanolis:
I'd have a suggestion, I think we should first gather the requirements / objectives and then take it from there (deductive) instead of dissecting this PR (inductive). We could then tackle them one by one via ideally individual PRs when we have rough consensus. 

I try to kick it off and gather the points so far, but feel free to edit / change:

### Sending transactions in Specter - new features or better UX of existing functionality

1. Make it easier to do batch transactions via the UI (e.g. with plus and minus buttons to add and remove recipients)
2. Shuffle the recipients (outputs) in a PSBT for better privacy (main question here: via front or back end)
3. Enable setting a manual change address

relativisticelectron:
> 1. Make it easier to do batch transactions via the UI (e.g. with plus and minus buttons to add and remove recipients)

Do you mean by "minus buttons", that it is possible to remove any recipient, and not just the last?

> 2. Shuffle the recipients (outputs) in a PSBT for better privacy (main question here: via front or back end)

The order of the recipients should be identical in `/wallets/wallet/seg/send/new` and `/wallets/wallet/seg/send/pending/` view, otherwise the user will be confused after creating a tx. That means a shuffling should be done already in `/wallets/wallet/seg/send/new`.

> Enable setting a manual change address

If 1) and 2) are done, the user can do this manually already.

moneymanolis:
>Do you mean by "minus buttons", that it is possible to remove any recipient, and not just >the last?

Yes.

> The order of the recipients should be identical in /wallets/wallet/seg/send/new and /wallets/wallet/seg/send/pending/ view, otherwise the user will be confused after creating a tx. That means a shuffling should be done already in /wallets/wallet/seg/send/new.

Well, of course the shuffling in the back end wouldn't be done in stealth mode - I'd prefer a toggle for that with a dice or so and we could even add dice symbol in the pending view as well to be completely sure.

relativisticelectron:
> Well, of course the shuffling in the back end wouldn't be done in stealth mode - I'd prefer a toggle for that with a dice or so and we could even add dice symbol in the pending view as well to be completely sure.

I don't think that is sufficient to make it clear to a novice user (2 outputs). I am worried about a novice user sending some small amount, specter swapping the order, and then at hardware wallet presenting the bigger change output first, letting the user confused. (this is why I let the reordering as an optional feature via drag and drop).

I think if we want to automate the random shuffling we have to show an intermediate step / page telling the user how the outputs were shuffled.

relativisticelectron:
> What's the difference whether the ordering is friend1, friend2, change, friend3 or friend 2, friend3, change, friend 1? This would only make sense for people who send regularly to the same recipients in a batch tx (perhaps employer?) - for this rather edgecasey scenario I'd suggest a shuffle toggle for the backend (default off) with user feedback that the shuffling had been applied.

OK, I can see the benefit of shuffling visibly does (probably) not outweigh the downside of bringing in the extra depenency.  I will close this PR and open a new one without SortableJS.

k9ert:
Why not removing it in this PR? Ok, maybe we have so many comments which are targeting that so that we can't see the comments targeting something else ...

PR #1748, opened 2022-06-06
fix starting server with hardware wallet plugged in

djpnewton:
fix issue where the server stalls on startup waiting for a hardware wallet pin to be entered if such a hardware wallet is plugged in when starting the server

there is no feedback to the user that the startup process is waiting for the hardware wallet pin to be entered (or unplugged) so it is confusing for the user why specter desktop has stalled

I havent experienced the issue documented in the removed code comment below.. perhaps it is not an issue with current HWI version? I guess it will need some more testing to verify:
```
        # Running enumerate after beginning an interaction with a specific device
        # crashes python or make HWI misbehave. For now we just get all connected
        # devices once per session and save them.
```

### DIFF
--- a/src/cryptoadvance/specter/hwi_rpc.py
+++ b/src/cryptoadvance/specter/hwi_rpc.py
@@ -75,11 +75,6 @@ class HWIBridge(JSONRPC):
             "extract_master_blinding_key": self.extract_master_blinding_key,
             "bitbox02_pairing": self.bitbox02_pairing,
         }
-        # Running enumerate after beginning an interaction with a specific device
-        # crashes python or make HWI misbehave. For now we just get all connected
-        # devices once per session and save them.
-        print("Initializing HWI...")  # to explain user why it takes so long
-        self.enumerate()
 
     @locked(hwilock)
     def enumerate(self, passphrase="", chain=""):


k9ert:
I have a similiar experience when i do `pytest` but i think i've fixed it with #1720 . But maybe that's also a different issue.
I tested with a trezor one and a bitbox2 and i did not experience any issue with startup. So the enumerate call doesn't seem to hurt?! 
So what is the setup where this bugs you? However the print-statement should definitely rather be a debug-statement.

djpnewton:
it only seems to happen with bitbox02 (I have only tested bitbox02 and ledger nano x so far).

I have raised an issue with HWI https://github.com/bitcoin-core/HWI/issues/613

k9ert:
Ok, i get it now. I think the HWI-issue should solve that issue and potentially, we could also solve it by removing the `enumerate` at startup. However, on the other hand, we might soon have a notification system with #1766. So i'd suggest to wait on that one and then do the enumeration in a thread maybe with a timeout if that's possible, informing the user about the result?! Would that make more sense?

k9ert:
Discussed with @moneymanolis , should not be that complicated than i first thought. He'll look and then we can probably merge this even without the notification-stuff.

moneymanolis:
I am experiencing the problem mentioned in the comment, Python crashes for me (tested with Trezor One) - so, we can't merge. @djpnewton I am not sure whether the issue was placed best with HWI, perhaps makes sense to at least also open an issue with the Bitbox guys? I'd close the PR for now, we can reopen if we have news from HWI and / or the Bitbox team.

moneymanolis (commented):
Can't merge, see comment in the conversation.

k9ert:
I think we could still improve the experience for the users:
* We should give a hint in the logs that if the bootup is not proceeding, the user should unplug their HWW or even other USB-Equipment (had the experience with a gaming controller).
* The electron-app is effectively starting another process. The reason why we even have the (endless-pacman-page)[https://docs.specter.solutions/desktop/endless-pacman/] is, the it's difficult to give the user specific feedback from an app not properly starting up. But it's not impossible, e.g. in this case. If a timeout occurs, we could check the lines of the log and realize that (currently) "Initializing HWI ..." is the last line. We could the give the user the hint to unplug. The ingredients to these changes are basically here: https://github.com/cryptoadvance/specter-desktop/blob/master/pyinstaller/electron/main.js#L370-L398

Not sure whether @djpnewton is up for such a change but reopening the PR anyway. Will close after a week or so and effectively turn this into an issue if we don't have acitivity here.

k9ert:
Ok, let's close this. Feel free to reopen if necessary.

PR #1749, opened 2022-06-06
Feature: combined test script for cypress and pytest

relativisticelectron:
Since I switch frequently between different consoles, conda and virtualenv it was cumbersome to get cypress and pytest to work each time.  This script should work (at least in linux) and run all tests.


k9ert (approved):

k9ert:
LGTM. Will merge soon.

PR #1752, opened 2022-06-09
added text-explainer webcomponent

b30wulffz:
Currently we have multiple instances of copy pasting code of text-explainer in the codebase.
Hence to avoid this, and in order to make code more modular, I have added a text-explainer webcomponent.

`<text-explainer>` is a custom element for showing extra information on the webapp. It makes use of a collapsible design, where some descriptive data is hidden from the user, and can be accessed by clicking on the one-liner title present after 📚 icon.

## Usage
```html
<text-explainer>
        <span slot="title">One-Liner Text which is visible</span>
        <div slot="content">
            Content that is hidden by default
        </div>
</text-explainer>
```

### Instance 1
wc-1
### Instance 2
wc-2

### DIFF
--- a/cypress/integration/spec_devices.js
+++ b/cypress/integration/spec_devices.js
@@ -2,6 +2,7 @@ describe('Test adding different devices', () => {
     before(() => {
         cy.viewport(1200,660)
         cy.visit('/')
+        Cypress.config('includeShadowDom', true)
     })
 
     // Keeps the session cookie alive, Cypress by default clears all cookies before each test
@@ -15,8 +16,8 @@ describe('Test adding different devices', () => {
         cy.get('#electrum_device_card').click()
         cy.get('#device_name').type("Electron's Electrum Device")
         // Open and close the explainer
-        cy.get("text-explainer").shadow().find('#toggle-explainer').click({force: true})
-        cy.get("text-explainer").shadow().find('#toggle-explainer').click({force: true})
+        cy.get('#toggle-explainer').find('#drop-icon').click()
+        cy.get('#toggle-explainer').find('#drop-icon').click()
         cy.get('#master_pub_key').type("vpub5VGXXixD2pHLFtcKtCF57e8mx2JW6fie8VydXijC8sRKAL4RshgjEmzbmV915NeVB9pd23DVYem6zWM7HXFLNwaffNVHowdD9SJWwESyQhp")
         cy.get('.small-card > .btn').click()
         cy.contains('New device was added successfully!')
--- a/cypress/integration/spec_devices.js
+++ b/cypress/integration/spec_devices.js
@@ -15,8 +15,8 @@ describe('Test adding different devices', () => {
         cy.get('#electrum_device_card').click()
         cy.get('#device_name').type("Electron's Electrum Device")
         // Open and close the explainer
-        cy.get('#toggle-explainer').click()
-        cy.get('#toggle-explainer').click()
+        cy.get("text-explainer").shadow().find('#toggle-explainer').click({force: true})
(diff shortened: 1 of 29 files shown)


cypress/integration/spec_devices.js
```diff
@@ -15,8 +15,8 @@ describe('Test adding different devices', () => {
         cy.get('#electrum_device_card').click()
         cy.get('#device_name').type("Electron's Electrum Device")
         // Open and close the explainer
-        cy.get('#toggle-explainer').click()
-        cy.get('#toggle-explainer').click()
+        cy.get("text-explainer").shadow().find('#toggle-explainer').click({force: true})
```
moneymanolis, line 18: I'd prefer to include **Cypress.config('includeShadowDom', true)** as below, then you don't need `.shadow()`
```
describe('Test adding different devices', () => {
    before(() => {
        cy.viewport(1200,660)
        cy.visit('/')
        Cypress.config('includeShadowDom', true)
    })
```

src/cryptoadvance/specter/templates/base.jinja
```diff
@@ -59,6 +59,7 @@
 	{% endif %}
 	{% include "includes/message-box.html" %}
 	{% include "includes/tool-tip.html" %}
+	{% include "includes/text-explainer.html" %}
```
moneymanolis, line 62: I'd rather not include it in the base template but only in those jinja templates where we need it, especially since we only use it twice so far.

b30wulffz: Ahh agree, I initially thought that text-explainer can be used at multiple places so did like that. I’ll update its import to specific files since sample space is small right now.

src/cryptoadvance/specter/templates/includes/text-explainer.html
```diff
@@ -0,0 +1,91 @@
...
+
+    <link rel="stylesheet" type="text/css" href="{{ url_for('static', filename='styles.css') }}">
+    <style>
+        :host {
+            display: block;
+        }
+
+        #toggle-explainer {
+            cursor: pointer;
+            font-size: 1.2em;
```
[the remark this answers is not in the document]
b30wulffz: In new_device_keys/electrum, 1em is not matching font-size of nearby elements. So I am planning to add a custom attribute to handle the styling of this dropdown title, where user can provide additional properties depending on the use case.

cypress/integration/spec_devices.js
moneymanolis: Do you want to change it? @b30wulffz I have the feeling that above is more stable.

src/cryptoadvance/specter/templates/includes/text-explainer.html
```diff
@@ -0,0 +1,88 @@
...
+<script type="text/javascript">
+    class TextExplainerElement extends HTMLElement {
+        constructor() {
+            super();
+            // Create a shadow root
+            let shadow = this.attachShadow({mode: 'open'});
+            let template = document.getElementById('text-explainer');
+            let clone = template.content.cloneNode(true);
+
+            this.$toggleExplainer = clone.querySelector("#toggle-explainer");
```
moneymanolis, line 67: Why did you add the dollar signs? I don't think they are necessary and I think it only creates confusion. Can we remove?

cypress/integration/spec_devices.js
moneymanolis: I've simplified the test in https://github.com/cryptoadvance/specter-desktop/pull/1752/commits/2acb643fb6fd1e60e4711b3966976cbac773ebab
What do you think?

b30wulffz: Ahh I see, understood where I went wrong.

PR #1755, opened 2022-06-10
hwi_rpc.py issues

djpnewton:
-change print statement to log statement
-fix issue in enumerate() where wrong branch executed due to comparing 'passphrase' param which defaults to an empty string to 'None'


k9ert (approved):
Thanks for the PR. LGTM.

PR #1772, opened 2022-06-20
Bugfix: Check also for duplicate xpubs in existing singlesig wallets

### LINKED ISSUE #1724: Creating a Wallet with Same Keys Creates Duplicate Wallet
When creating a wallet, if you define a wallet with the same device pool and same parameters (x of y multi-sig), Specter will create a duplicate of an existing wallet if a wallet with the same parameters had already been created previously.

**To Reproduce**
1. Create a wallet with some specific devices and parameters (2 of 3 for example)
2. Transfer a balance into the wallet
3. Create a second wallet with identical parameters
4. Note that the second wallet has the same balance and transaction history as the first wallet.

**Expected behavior**
I expected Specter to create the second wallet as a unique and different wallet.

 - Where is your node running: local Bitcoin Core
 - Where is your specter running: local binary
 - If you're running a binary, which OS: Windows 11
 - Browser Chrome
 - Specter Version 1.9.2

**Additional context**
This issue can likely be fixed by allowing the user to increment the derivation paths of one or more device's keys so that the same set of keys is not used for the second wallet even though it is the same pool of signing devices.

k9ert:
fixes #1724


relativisticelectron:
If I understand correctly, it would prevent using 1 xpub in different multisig setups.  This should however **NOT** be prevented. An xpub can safely be reused in multisig setups.

k9ert:
In theory, i'll agree with you. However, in practice, it survived exactly this test. I feel a bit guilty creating this PR not knowing exactly why ...

relativisticelectron:
I am not quite sure what you are saying.  Reusing xpubs in multisig setups makes perfect sense. Examples:

- 1 device/xpub is the recovery key (possibly on the other side of the world) and is used in many multisig setups
- 2 or more devices/xpubs are the same in multisig setups of different size, giving different recover-ability and complexity tradeoffs

k9ert:
I fully agree with you! The issue is that my manual tests went well with your specification and i don't know why. I'll definitely make more explicit what i want to achieve here and also how this is implemented.

moneymanolis:
The current implementation would (among other things) disallow to use an xpub for a multisig wallet if it is already used in a singlesig. This should definitely NOT be the case.

The original issue was about an exact duplicate of a wallet. I think it would be better to start from this level - the wallet - and not from individual xpubs. Is there some attribute / property (like a hash of a wallet) that unambiguously identifies a wallet? If so, we could either disallow to exactly replicate a wallet or give the user feedback via flashing a message that he is about to replicate a wallet.

k9ert:
I've created a test and it seems that my manual tests were somehow flawed. Anyway, now i have a test to work against. Probably the fingerprint of the wallet is the best way to check whether something is already in use.

k9ert:
Next steps would be:
* Resolving the conflicts
* git cherry-pick top port the PR to `uiux-revamp`
* Finish it

Issue #1759, opened 2022-06-12
Error in consecutively uploading same file twice


b30wulffz:
**Describe the bug**
The bug is in file-uploader webcomponent where I can't upload same file consecutively twice.

**To Reproduce**
1. Goto any wallet of your choice.
2. Click on "Send". Then click on "Import" tab. This will open PSBT uploader.
3. Now upload a text file.
4. Clear the text content in the textarea.
5. Upload the same text file again. No text content will be uploaded.

PS: You can try the same "Settings > Advanced" of the wallet for importing address labels.

**Expected behavior**
Ideally, the text content should be updated to the one stored in the latest file.

**ScreenRecord**
![ezgif com-gif-maker](https://user-images.githubusercontent.com/31105595/173221578-acad9096-8473-48c0-af9a-1e545692358c.gif)
In this demo, when I tried uploading the same file consecutively, nothing got uploaded. When I tried uploading a different file, tand hen the same file, the upload was successful.

**Desktop (please complete the following information):**
 - Where is your node running: local(internal)
 - Where is your specter running: locally (development build)
 - If you're running a binary, which OS: Linux
 - Browser: Brave

relativisticelectron:
- In Firefox (Linux) this error does not occur.
- In Chromium (Linux)  I can reproduce this error.

PR #1776, opened 2022-06-24
Bugfix: Consecutively uploading same file twice

relativisticelectron:
Fix for https://github.com/cryptoadvance/specter-desktop/issues/1759


relativisticelectron:
@b30wulffz : Can you please check if it solves the issue also for you?

moneymanolis (approved):
Nice fix, thank you!

moneymanolis:
I've just changed the comment to make clear that this is an issue for Chrome only, since Chrome is not treating it as a change event if the input value is the same whereas Firefox for example does treat it as a change event. Good idea to use the click event and a clearing of the input value!

b30wulffz:
tested, its working very well, thanks for the fix

PR #1778, opened 2022-06-28
Bugfix: Pagination in transactions and coin selection

moneymanolis:
### Content
- Remove the pagination counter (Page 1 **_**of 3**_**) if there is only 1 page
- Also set the default value of 25 entries in tx table
- Prettify the page limit select element

### Visuals
Old:
https://user-images.githubusercontent.com/70536101/176268630-7950b6cf-0b55-44ee-bdb7-3d21343a22de.mov

New:
https://user-images.githubusercontent.com/70536101/176268451-6f2134a1-21ea-4f8a-9822-37dedcf137aa.mov

### DIFF
--- a/src/cryptoadvance/specter/templates/includes/tx-table.html
+++ b/src/cryptoadvance/specter/templates/includes/tx-table.html
@@ -87,10 +87,8 @@
             border: 1.5px solid var(--cmap-border);
             z-index: 9;
         }
-        .page-limit-div {
-            flex-grow: 0.3;
-            display: flex;
-            justify-content: space-around;
+        .page-limit-label {
+            margin-left: 5px;
         }
         .search {
             margin: auto;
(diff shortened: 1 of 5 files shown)


src/cryptoadvance/specter/templates/includes/tx-table.html
```diff
@@ -942,10 +942,14 @@ <h1>{{ _("Export transactions to CSV") }}</h1>
                 // If there are multiple pages show the pagination bar
                 if (this.pageCount > 1) {
                     this.paginationContainer.classList.remove('hidden');
+                    this.paginationCounter.classList.remove('hidden');
                     this.paginationCounter.innerText = `of ${this.pageCount}`;
                     this.paginationIdxInput.max = this.pageCount;
                 } else {
+                    console.log("we are running in pageCount is less or equal to 1")
```
b30wulffz, line 949: This can be commented.
Also, shouldn't we hide the entire pagination counter when it is just a single page?

b30wulffz:
Pagination bar gets hidden when scrolling, a sticky fix can be implemented.

src/cryptoadvance/specter/templates/includes/tx-table.html
moneymanolis: That is just from debugging, will delete it.
Was thinking about hidding the pagination completely, but it somehow looked smoother to keep Page 1 - but I don't have strong opinion here.

b30wulffz (approved):
Nice work, lgtm

PR #1779, opened 2022-06-29
Feature: Rescan button for empty wallet

yogendra-17:
Rescan-button added when no-transaction appear.

### Visual
For a visual see here below:
https://github.com/cryptoadvance/specter-desktop/pull/1779#issuecomment-1170374596


moneymanolis (approved):
Nice implementation, I've directly pushed style changes.

moneymanolis:
@yogendra-17 in https://github.com/cryptoadvance/specter-desktop/pull/1779/commits/663df878888335218aa59f9b66248171e93d4559 I've also added a little animation for better orientation for the user to find the relevant rescan part (in addition to your scrollintoview). Let me know what you think.
It looks like this:
https://user-images.githubusercontent.com/70536101/176514604-970bfb51-8c6c-4d4f-981f-99412319821c.mov

yogendra-17:
> @yogendra-17 in [663df87](https://github.com/cryptoadvance/specter-desktop/commit/663df878888335218aa59f9b66248171e93d4559) I've also added a little animation for better orientation for the user to find the relevant rescan part (in addition to your scrollintoview). Let me know what you think. It looks like this: https://user-images.githubusercontent.com/70536101/176514604-970bfb51-8c6c-4d4f-981f-99412319821c.mov

Thanks for implementing changes,They are working fine.

moneymanolis (changes requested):
The issue with this implementation is that the rescan button is also shown for a completely fresh wallet with no tx (where a rescan doesn't make any sense and could make for a bad UX). Ideally, we find a heuristic to distinguish between a fresh wallet and a wallet with pre existing tx / funds.

yogendra-17:
> 

I think at this moment this is the best approach that we can go for, we can merge now ig.

moneymanolis (approved):

k9ert:
Just discussed with @moneymanolis about the failing cypress-tests: We have no idea why the tests don't pass but if the button would be inside the table, it would be much simpler as no eventing logic would be necessary. The "no transactions" text is not failing as the button is.

So let's change the location of that button and make it a part of the WebComponent and it should be much simpler and the test has less chance to fail.

moneymanolis:
> Just discussed with @moneymanolis about the failing cypress-tests: We have no idea why the tests don't pass but if the button would be inside the table, it would be much simpler as no eventing logic would be necessary. The "no transactions" text is not failing as the button is.
> 
> So let's change the location of that button and make it a part of the WebComponent and it should be much simpler and the test has less chance to fail.

Implemented in https://github.com/cryptoadvance/specter-desktop/pull/1779/commits/baeeeb61fe2158d7aa5662b173439d657b55217d

k9ert (approved):

PR #1780, opened 2022-06-29
Chore: added extended import test

k9ert:
Tests an import of a wallet with lots of TXs

This sets up a node and funds a default-wallet w0
This wallet will then create lots of TXs to a wallet w1 which got created
and imported a descriptor.

After that, a SpecterWallet with the same descriptor get created + rescan.
At the end the txlist should be very similiar with the TXs of w1.


k9ert:
Doesn't fit so well in the rest of the test-framework. Won't merge. Closing.

PR #1782, opened 2022-06-30
UI: Improve Adding/deleting recipients in Send , and move the recipient to a webcomponent

relativisticelectron:
Follow up on: https://github.com/cryptoadvance/specter-desktop/pull/1747

image

**TODO:**
- [x] Restructure the innerHTML code into a web component
- [x] Adding, and deleting of recipients
- [x] sort out tests and cypress
- [x] remove `console.log`s    after review  (this will allow fast debugging if needed)

**Notable changes:**
- The webcomponent  `recipient-box` has a proptery `value`  which contains a dictionary like:
```
{
    unit: "btc",
    amount: 2,
    recipient_id: 1,
    address: "bcrt1qfvkcy2keql72s8ev87ek93uxuq3xxsx9l0n03r",
    label: "",
    amount_input: "2",
    btc_amount: 2
}
```
- All   `value`s of all `recipient-box`es are then combined before form submission to a dictionary, jsonified and stored in an hidden input `recipient_dicts` and then available in request.form (Other methods of making the fields available for request.form did not work and are less elegant.).    This will allow in the future to unify some code from `PsbtCreator.paymentinfo_from_ui` and  `PsbtCreator.paymentinfo_from_json`

**Breaking API changes:**
Changed [`subtract_from` to start from `0`]( https://github.com/relativisticelectron/specter-desktop/blob/f20ca566a2fccdd13d68e1e234067174a44fe06a/src/cryptoadvance/specter/commands/psbt_creator.py#L301), relevant for `paymentinfo_from_json`  see  https://github.com/cryptoadvance/specter-desktop/blob/5f82e2a8f89f5532f2f609f75d8d7fd0152f02bd/src/cryptoadvance/specter/commands/psbt_creator.py#L198 .  Original:
https://github.com/cryptoadvance/specter-desktop/blob/5f82e2a8f89f5532f2f609f75d8d7fd0152f02bd/src/cryptoadvance/specter/commands/psbt_creator.py#L299


relativisticelectron:
There is a cypress error that should have nothing to do with this PR.  Any ideas how to solve this?
image

relativisticelectron:
> EDIT: Not yet clear how to fix this: image

This workaround https://github.com/cryptoadvance/specter-desktop/blob/4d2abc75f518ff2d2901b002772a3d97ec53e79a/cypress/integration/spec_elm_single_segwit_wallet.js#L41   was not necessary (the psbt disappeared as intended). So the workaround itself lead to the cypress error.  I fixed it by removing the workaround.

src/cryptoadvance/specter/commands/psbt_creator.py
```diff
@@ -138,29 +138,31 @@ def paymentinfo_from_ui(cls, specter, wallet, request_form):
...
         amounts = []
         amount_units = []
-        while "address_{}".format(i) in request_form:
-            addresses.append(request_form["address_{}".format(i)])
+
+        recipient_dicts = json.loads(request_form["recipient_dicts"])
+        print(recipient_dicts)
+        for recipient_dict in recipient_dicts:
+            addresses.append(recipient_dict["address"])
             amount = 0.0
```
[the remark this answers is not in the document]
relativisticelectron: I guess this line could be moved into the except part...

moneymanolis:
@relativisticelectron I've simplified the approach for the subtract_from value setting in https://github.com/cryptoadvance/specter-desktop/pull/1782/commits/71444ab9dcdb2df5e896e5a52f1f14d90d732e61 - it was also a bit anti pattern to set an input value for this outside of the web component (fee-selection.html in this case). Let me know whether this is fine by you. I am also having trouble with the Cypress tests locally, so I am "misusing" the CI to test whether the changes broke anything. Manually testing was fine.

relativisticelectron:
> @relativisticelectron I've simplified the approach for the subtract_from value setting in [71444ab](https://github.com/cryptoadvance/specter-desktop/commit/71444ab9dcdb2df5e896e5a52f1f14d90d732e61) - it was also a bit anti pattern to set an input value for this outside of the web component (fee-selection.html in this case). Let me know whether this is fine by you. I am also having trouble with the Cypress tests locally, so I am "misusing" the CI to test whether the changes broke anything. Manually testing was fine.
> 
> Update: Of course, Cypress fails us. I will investigate. Doesn't look my change is the culprit, though, rather some general problems to access the shadow dom of the new recipient web component.

[This broke the ](https://github.com/cryptoadvance/specter-desktop/commit/71444ab9dcdb2df5e896e5a52f1f14d90d732e61)  
removed the important line
` 			document.getElementById('subtract_from').value =  get_recipient_ids().indexOf(parseInt(subtract_from_recipient_id_select.value, 10)); `

that should be why cypress fails.

relativisticelectron:
> I think it is unrelated to the failing Cypress tests, they already fail before this becomes relevant in the tests. I'll try to fix them, let's see.

At e59d2d65  I also get cypress erros that "should" be unrelated to this PR:
image

moneymanolis:
The question to me is: Isn't it more intuitive that if you have 5 recipients and you delete recipient 3 and 5, that the remaining recipient automatically just become recipient 1, recipient 2 and recipient 3? Why do we keep them like recipient 1, 2 and 4?

relativisticelectron:
> The question to me is: Isn't it more intuitive that if you have 5 recipients and you delete recipient 3 and 5, that the remaining recipient automatically just become recipient 1, recipient 2 and recipient 3? Why do we keep them like recipient 1, 2 and 4?

I think of them more like ids. And relabeling the numbers could create confusion of what just happened.  A missing number should not be weird, if you just deleted the middle one.

An alternative is just to remove the numbers completely off the UI.

moneymanolis:
@relativisticelectron I couldn't come up with a better solution for the ordering of the recipients, so I tried to make the previous approach work again while still using mostly the fee-selection.html component for handling it. Could you test it? Will focus on the Cypress tests and styling then.

relativisticelectron:
After looking through the javascript code, I need to do some fixing and reorganizing to make it clearer.  Please pause the review, until I finished that.

relativisticelectron:
In hindsight the webcomponent had lots of issues. I think I resolved them now all and the structure should be way clearer.

@moneymanolis : Thanks for the time you put into the review.  Please also push the style changes

moneymanolis:
Contents of https://github.com/cryptoadvance/specter-desktop/pull/1782/commits/835423eec405a92ab302ca7910aa9ace773185ed:
- I've refactored `command.js` to reduce code duplication in `spec_wallet_send.js `
- I had to include` {force:true}` for everything related to typing and the recipient web component. I ran into this flakiness all the time: https://github.com/cypress-io/cypress/issues/5830 - seems to be more stable now. This was also the reason why I upgraded the Cypress version. I am not completely sure whether this was necessary but the version seems to be less flaky, and @k9ert said it was a good idea to upgrade here on master as well, so I've put it in.
- I added some small tests in `spec_wallet_send.js`

If the Cypress tests are green, I will move on to the styling part.

moneymanolis:
Improved styling in https://github.com/cryptoadvance/specter-desktop/pull/1782/commits/a16c9663ae9457f8a0c1220eaac98970524c2a82:

grafik

@relativisticelectron let me know, what you think. I'd also like to get rid of the x / delete button if there is only one recipient, would that be fine by you? I'd make another commit for that as well.

relativisticelectron:
The styling is nice!   Yeah I tried getting rid of the button in css and only-child, but couldn't get it to work.

moneymanolis:
> The styling is nice! Yeah I tried getting rid of the button in css and only-child, but couldn't get it to work.

I think I might have a rough idea, I'll give it a shot then.

moneymanolis:
https://github.com/cryptoadvance/specter-desktop/pull/1782/commits/267bf75c962a5c8adda2b1d98feb221539f06481 looks like this:

https://user-images.githubusercontent.com/70536101/180530951-323fe442-f19e-4424-9d66-12cdcfaad54c.mov

@relativisticelectron if you agree with the approach, I'd adapt the Cypress test accordingly tomorrow. We are nearing the finish line then. Have a look at the open comments that I made, so we can merge soon.

relativisticelectron:
Yes looks nice. I simplified the  (hidden Button) code in https://github.com/cryptoadvance/specter-desktop/pull/1782/commits/255eaddd6f3d5f951d4f66c8e66ac6d03709f93d .

relativisticelectron:
> **Breaking API changes:** Changed [`subtract_from` to start from `0`](https://github.com/relativisticelectron/specter-desktop/blob/f20ca566a2fccdd13d68e1e234067174a44fe06a/src/cryptoadvance/specter/commands/psbt_creator.py#L301), relevant for `paymentinfo_from_json` see
> 
> https://github.com/cryptoadvance/specter-desktop/blob/5f82e2a8f89f5532f2f609f75d8d7fd0152f02bd/src/cryptoadvance/specter/commands/psbt_creator.py#L198
> . Original:
> 
> https://github.com/cryptoadvance/specter-desktop/blob/5f82e2a8f89f5532f2f609f75d8d7fd0152f02bd/src/cryptoadvance/specter/commands/psbt_creator.py#L299

Before we merge I wanted to bring the point above again to attention and ask if `paymentinfo_from_json` is already used somewhere. If so it would need to be changed.

moneymanolis:
> paymentinfo_from_json

I couldn't find anything, tagging @k9ert to double check.

moneymanolis:
> > **Breaking API changes:** Changed [`subtract_from` to start from `0`](https://github.com/relativisticelectron/specter-desktop/blob/f20ca566a2fccdd13d68e1e234067174a44fe06a/src/cryptoadvance/specter/commands/psbt_creator.py#L301), relevant for `paymentinfo_from_json` see
> > https://github.com/cryptoadvance/specter-desktop/blob/5f82e2a8f89f5532f2f609f75d8d7fd0152f02bd/src/cryptoadvance/specter/commands/psbt_creator.py#L198
> > 
> > . Original:
> > https://github.com/cryptoadvance/specter-desktop/blob/5f82e2a8f89f5532f2f609f75d8d7fd0152f02bd/src/cryptoadvance/specter/commands/psbt_creator.py#L299
> 
> Before we merge I wanted to bring the point above again to attention and ask if `paymentinfo_from_json` is already used somewhere. If so it would need to be changed.

I couldn't find anything, tagging @k9ert to double check.

moneymanolis:
Ah, I think my review comments are pending unless I submit the review, will do in a second.

moneymanolis:
> I simplified the (hidden Button) code in [255eadd](https://github.com/cryptoadvance/specter-desktop/commit/255eaddd6f3d5f951d4f66c8e66ac6d03709f93d) .
> 
Nice, much better.

https://github.com/cryptoadvance/specter-desktop/pull/1782/commits/b9c5d601791460299ce7516af99463e21f546679 is also a great improvement. I didn't understand how it worked beforehand to be honest without the CustomEvents that can be picked up by the functions defined in `wallet_send.` So, good that you changed it!

moneymanolis:
I'll update the Cypress test then. We have an issue on Cirrus regarding the (higher) Cypress version. Not sure yet, what is going on there. Seems like it is not caching the higher version properly.

moneymanolis (commented):
Here are my comments. The most important is regarding the arguments `addRecipient() `and what would happen if we want to prepopulate with multiple recipients.

src/cryptoadvance/specter/templates/wallet/send/new/wallet_send.jinja
```diff
@@ -190,181 +228,109 @@
...
+			let recipientIds = getRecipientIds();
+			recipientId = 0;
+			if (recipientIds.length > 0) {
+				recipientId = Math.max(...recipientIds)+1
+			}; 
+
+			// This adds the recipient
+			let recipients = document.getElementById('recipients');			
+			let newRecipient = new RecipientBox();
+			newRecipient.myatt = 100;
```
moneymanolis, line 242: What is this?

relativisticelectron: that was debug code. Can be deleted.

src/cryptoadvance/specter/templates/includes/recipient-box.html
```diff
@@ -0,0 +1,486 @@
...
+            this.shadowRoot.appendChild(this.clone);
+			this.addAssetSelectorToShadowRoot('', 'btc')			
+		}
+
+
+
+		// add the html code for the assetSelector
+		addAssetSelectorToShadowRoot(amount, amountUnit) { 
+			if (amount >= 0){
+				this.shadowRoot.getElementById('amount').value = amount;
```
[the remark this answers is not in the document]
relativisticelectron: This is due to my lack of understanding why we need clones.   AFAIK a ShadowRoot is simply a sub-html-structure, in which the elements (and their ids) are nicely encapsulated. That way I can choose simple ids like `id="remove"` and don't have to do `id="remove_0"`.  
Why would we need clones?

tests/test_commands_psbt_creator.py
```diff
@@ -43,10 +36,15 @@ def test_PsbtCreator_ui(caplog):
...
+        "bcrt1qfvkcy2keql72s8ev87ek93uxuq3xxsx9l0n03r",
     ]
-    assert psbt_creator.amounts == [0.1, 0.00111211]
-    assert psbt_creator.labels == ["someLabel", "someOtherLabel"]
-    assert psbt_creator.amount_units == ["btc", "sat"]
+    assert psbt_creator.amounts == [0.1, 0.00111211, 0.003]
+    assert psbt_creator.labels == [
+        "someLabel",
+        "someOtherLabel",
+        "<script>console.log('I escaped')</script>",
```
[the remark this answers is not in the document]
relativisticelectron: yeah...  :-)  I just wanted to see if even a weird string is handled correctly.

cypress/support/commands.js
```diff
@@ -234,9 +234,10 @@ Cypress.Commands.add("mine2wallet", (chain) => {
 // Quick and easy way to fill out the send form and create a psbt
 Cypress.Commands.add("createPsbt", (address, label="a_label", amount=0.01) => { 
   cy.get('#btn_send').click()
-  cy.get('#address_0').type(address)
-  cy.get('#label_0').type(label)
+  // it is not clear why .shadow(), or { includeShadowDom: true } is needed here to find the elements in the ShadowDOM, but not in the other cypresss tests
```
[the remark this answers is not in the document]
relativisticelectron: Thanks for explanation. Now it is clear.

tests/test_commands_psbt_creator.py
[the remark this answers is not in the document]
moneymanolis: OK

src/cryptoadvance/specter/templates/includes/recipient-box.html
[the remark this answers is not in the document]
moneymanolis: I mean you also did it like this here: https://kinsta.com/blog/web-components/#html-templates
The only difference is how to query HMTL elements when assigning them as properties of the HTML element. I think in your approach you were fine with using the Shadow Root since you had already appended the clone of the template to the Shadow Root and in other web components in the code base the querying had to be done through the clone of the template because the Shadow Root had not been appended yet. 

This is a good and simple example from `dnd-textarea.html` (querying takes place before the `appendChild(clone)`):

```
constructor() {
  let shadow = this.attachShadow({mode: 'open'});
  let template = document.getElementById('dnd-textarea');
  let clone = template.content.cloneNode(true);
  this.textarea = clone.querySelector("textarea");
  shadow.appendChild(clone);
}
```

I suppose the main drawback is more lengthy code. Here for example you could have used `this.amount` instead of writing always `this.shadowRoot.getElementById('amount') `if you had defined` this.amount` in the constructor.

grafik

We can leave it like this but we'd need at least an explanation / comment that this web component uses a little bit of different approach in this regard. Tbh, I can't say for sure what the best practice is.

src/cryptoadvance/specter/templates/wallet/send/new/wallet_send.jinja
moneymanolis, line 242:
```suggestion
```

src/cryptoadvance/specter/templates/includes/recipient-box.html
```diff
@@ -0,0 +1,486 @@
...
+		// creates the entire html code (without event listeners)  mainly from the template
+		// and it also populates the dictionary this.inputs
+		buildHTML(){
+			// Create a shadow root
+			this.attachShadow({
+				mode: 'open'
+			});
+
+			var template_content = document.getElementById('recipient-box').content;
+            this.clone = template_content.cloneNode(true);
```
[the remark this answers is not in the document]
relativisticelectron: ` this.clone` is also used.  Can you replace both occurrences?

moneymanolis:
Note: Cypress tests are green in local dev

src/cryptoadvance/specter/templates/includes/recipient-box.html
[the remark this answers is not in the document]
moneymanolis: Done in https://github.com/cryptoadvance/specter-desktop/pull/1782/commits/efb579b72355678112264f25485d28812d036b66

src/cryptoadvance/specter/templates/wallet/send/new/wallet_send.jinja
```diff
@@ -190,183 +220,103 @@
 			}
 		})
 
+
 		function addRecipient(addr, amount, amount_unit, label) {
```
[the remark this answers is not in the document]
k9ert: This functionality would be relevant for extensions which integrate with exchanges. So if you want to sell some bitcoin (stupid idea anyway) you could fill a form automatically with the address of the exchange.
However it's currently not used and that already since month and so it might make sense to make it anew when it's necessary again.

So feel free to purge this as needed.

src/cryptoadvance/specter/templates/wallet/send/new/wallet_send.jinja
```diff
@@ -190,181 +228,104 @@
...
-			`
+			let recipientIds = getRecipientIds();
+			recipientId = 0;
+			if (recipientIds.length > 0) {
+				recipientId = Math.max(...recipientIds)+1
+			}; 
+
+			// This adds the recipient
+			let recipients = document.getElementById('recipients');			
+			let newRecipient = new RecipientBox();
```
k9ert, line 241: Good, Looks much cleaner than what is was before!

src/cryptoadvance/specter/templates/includes/recipient-box.html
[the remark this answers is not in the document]
k9ert: I would agree here to @moneymanolis. Javascript as such is already quite flexible and can be misused in a lot of ways. Having consistency in the project is a value in itself. 
I also see value in clear explicit logical (buiseness-) references: e.g. `amount` getter and setter.
However, referring to internal representation of these (business-) properties via some search-algorithm (like `shadowRoot.getElementById`) doesn't feel right. 
So ideally maybe we can have additional members in the class like `amountInput` or `amountElement` which are defined in the constructor and used instead of `shadowRoot.getElementById("amount")`?

src/cryptoadvance/specter/commands/psbt_creator.py
```diff
@@ -138,29 +138,31 @@ def paymentinfo_from_ui(cls, specter, wallet, request_form):
         """calculates the correct format needed by wallet.createpsbt() out of a request-form
```
k9ert, line 138: Can you have a look in line 31. Is that description of how the `request_form` look like? Can you update that? 
@stepansnigirev might be happy as he has created his exfund extension using this API:
https://github.com/stepansnigirev/specterext-exfund/blob/master/src/cryptoadvance/specterext/exfund/controller.py#L68

stepansnigirev: I tried with json, but it was giving me troubles with liquid assets, so I switched to the request_form.
I should probably try with json again, add a few tests and report back if something doesn't work especially on Liquid. I hope I can get to it on Saturday.

src/cryptoadvance/specter/templates/includes/recipient-box.html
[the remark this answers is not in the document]
relativisticelectron: @k9ert , @moneymanolis  : Thanks for the advice. I followed `  this.textarea = clone.querySelector("textarea");`  in 5d3c188781c21d5afd35b686d02ad324cb4980ed  (which also removes old code).

src/cryptoadvance/specter/commands/psbt_creator.py
k9ert: @moneymanolis took a screenshot from the browser. This is how it looks like in this PR: 
image

src/cryptoadvance/specter/templates/includes/recipient-box.html
```diff
@@ -105,20 +102,11 @@
 			var template_content = document.getElementById('recipient-box').content;
             let clone = template_content.cloneNode(true);
 
-
-			// fill this.inputs with the input and textarea fields.
```
moneymanolis: Was this part redundant code?

moneymanolis:
I like the structure in https://github.com/cryptoadvance/specter-desktop/pull/1782/commits/d6c8689a85c1a483eb2f0a3edff3dfe97a5f8bac - excellent work overall, @relativisticelectron.

src/cryptoadvance/specter/templates/includes/recipient-box.html
relativisticelectron: yes.

relativisticelectron:
> I like the structure in [d6c8689](https://github.com/cryptoadvance/specter-desktop/commit/d6c8689a85c1a483eb2f0a3edff3dfe97a5f8bac) - excellent work overall, @relativisticelectron. I just have one comment (see above).

Thanks. It was my first web component. Thanks for all the feedback; it helped me a lot.

src/cryptoadvance/specter/templates/wallet/send/new/wallet_send.jinja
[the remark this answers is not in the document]
moneymanolis: @relativisticelectron I think this is the only remaining task. 

This 
```
		{% if fillform %}
			{% for addr, amount, amount_unit, label in recipients %}
			addRecipient("{{ addr }}", {{ amount }}, "{{ amount_unit }}", "{{ label }}");
			{% endfor %}
		{% else %}
			addRecipient("", 0, "btc", "");
```
probably will reduce to just` addRecipient()`, no? Please also adjust the other occurrences ` addRecipient()` then (and it particular the definition of `add Recipient()`in `wallet_send`).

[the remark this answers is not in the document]
relativisticelectron: The functionality of `fillform` should be independent of this PR.  Would it not be better to do the removal/replacement  in a different PR?  `fillform` also occurs in  https://github.com/cryptoadvance/specter-desktop/blob/7cf5b7637dedb40ee5f77b45270d5e1d351c2012/src/cryptoadvance/specter/server_endpoints/wallets.py#L597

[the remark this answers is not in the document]
moneymanolis: There is a comment, there, too: `# TODO: Not yet used. Remove if the use case doesn't happen.`

We can handle the `fillform` stuff in a different PR, but IMO this PR is defining `addRecipient()` in a new way and should have the correct arguments. In Python the code would fail, but JS is using "null" if undefined arguments are passed.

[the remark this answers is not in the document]
moneymanolis: OK, we should check how / whether fillfrom=True is really used [there](https://github.com/cryptoadvance/specter-desktop/blob/7cf5b7637dedb40ee5f77b45270d5e1d351c2012/src/cryptoadvance/specter/server_endpoints/wallets.py#L558) and whether it breaks sth. with rbf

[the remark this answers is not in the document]
relativisticelectron: Yes it is, I just checked. And I am currently implementing that the arguments in `addRecipient` are used.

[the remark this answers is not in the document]
relativisticelectron: 85a4777c81ed213435ad3240014c7395d3b19d88 enables the use of the arguments in `addRecipient`

src/cryptoadvance/specter/templates/includes/recipient-box.html
```diff
@@ -109,52 +124,48 @@
...
+				this.setUnit =  (newValue) => {
+					this.amountUnitBtcElement.checked =  newValue == 'btc';
+					this.amountUnitSatElement.checked =  newValue == 'sat';
+					this.updateUnitLabelAndStep();
+				}
+				this.getUnit =  () => {
+					return this.amountUnitBtcElement.checked  ? this.amountUnitBtcElement.value : this.amountUnitSatElement.value ;
+				}
+
+				// set the initial unit at creation
```
moneymanolis, line 164: I think this comment should rather be below at in the same file:

```
		// unit
		set unit(newValue){
			// this function is different for bitcoin and liquid 
			this.setUnit(newValue);
		}			
		get unit(){
			// this function is different for bitcoin and liquid 
			return this.getUnit();
		}	
```

This is what is called from "the outside", e.g. in `wallet_send` now with:
`newRecipient.unit = amount_unit;`

src/cryptoadvance/specter/templates/wallet/send/new/wallet_send.jinja
```diff
@@ -190,181 +227,108 @@
...
-				document.getElementById('remove-recipient').style.display = 'block';
-				document.getElementById('coin-selection-row').style['margin-top'] = '90px';
+				document.getElementById('vertical-space-for-subtract-fees-from-amount-checkbox').style['margin-top'] = '60px';
 				if (document.getElementById('subtract').checked) {
 					FeeSelectionComponent.showSubtractFrom(true)
 				}
 			}
-			calculateConvertedUnit(i);
+
+			document.getElementById(`recipient_${recipientId}`).calculateConvertedUnit();
```
moneymanolis, line 289: Why is this needed?

src/cryptoadvance/specter/templates/includes/recipient-box.html
```diff
@@ -0,0 +1,486 @@
...
+		}		
+
+		// amount
+		set amount(newValue){
+			this.amountElement.value = newValue;
+			this.calculateConvertedUnit()
+		}			
+		get amount(){
+			var value = parseFloat(this.amountElement.value );
+			if (!value) {value = 0}
```
moneymanolis, line 357: If you always set the value to 0 here, this will never lead to this part in `calculateConvertedUnit()`
```
			if (convertedAmount == 'NaN') {
				convertedAmount = '-'
			}
```
This also leads to confusing error messaging when submitting:

grafik

We have a pretty encompassing `validateAmount()` in `wallet_send`that is taking care of all weird input. I would just delete` if (!value) {value = 0}`.

src/cryptoadvance/specter/templates/wallet/send/new/wallet_send.jinja
```diff
@@ -501,35 +468,39 @@
...
+						if(recipient.unit in assetAmounts){
+							assetAmounts[recipient.unit] += recipient.amount;
 						}else{
-							assetAmounts[unit] = amount;
+							assetAmounts[recipient.unit] = recipient.amount;
 						}
 					}
-					if (document.getElementById("amount_" + i).value == '') {
-						amount = 0;
+					if (document.getElementById(`recipient_${recipient.recipientId}`).shadowRoot.getElementById('amount').value == '') {
```
moneymanolis, line 492: I would remove this, it is better feedback to the validation fail without setting the amount to zero beforehand.

moneymanolis:
@relativisticelectron I made a few comments for final optimisations. Great solution with https://github.com/cryptoadvance/specter-desktop/pull/1782/commits/85a4777c81ed213435ad3240014c7395d3b19d88. No we can just keep fillform as it is. This will definitely be used in an extension etc. Thanks for all the work!

src/cryptoadvance/specter/commands/psbt_creator.py
moneymanolis: @relativisticelectron, if this is fine how I updated it in this commit https://github.com/cryptoadvance/specter-desktop/pull/1782/commits/033baa6007638e72274834c19fac2697ea4f85b3 we can resolve this.

src/cryptoadvance/specter/templates/wallet/send/new/wallet_send.jinja
relativisticelectron: Ok.

src/cryptoadvance/specter/templates/includes/recipient-box.html
relativisticelectron: I am not sure I understand.  Do you mean the comment `// set the initial unit at creation` should be in `recipient_box.html`?

src/cryptoadvance/specter/templates/wallet/send/new/wallet_send.jinja
relativisticelectron: It is not needed. Thanks for pointing that out.

src/cryptoadvance/specter/templates/includes/recipient-box.html
relativisticelectron: Ok.

relativisticelectron:
I tested the `fillform` with 
Peek 2022-07-28 16-30

@moneymanolis : Could you test the liquid unit changer?

moneymanolis:
> I tested the `fillform` 

Looks good!
> @moneymanolis : Could you test the liquid unit changer?

OK, will do!

src/cryptoadvance/specter/templates/includes/recipient-box.html
moneymanolis: Or put it in `wallet_send` where `newRecipient.unit = amount_unit`; it just looks a bit lost here. Can IMO also just be deleted.

relativisticelectron: Oh I missed that, yes in https://github.com/cryptoadvance/specter-desktop/pull/1782/commits/b3ff4d88ed0bbff8ce1585b94700ab191c35e6f2

moneymanolis:
I've tested with a Elements Regtest node: The asset selector works if we take the highlighted second condition out of this line:
{% if specter.is_liquid **and wallet.balance.get("assets", {})** %}

Looks like this then:

grafik

I suppose the condition means that there are no other "assets" than LBTC in the wallet (because if there are there would be more to select from). I will double check with Stepan on how to proceed, he has implemented all the Liquid stuff and I still don't (want to? :-)) grasp it.

Not sure whether we need 10-8 here:

grafik

relativisticelectron:
OK. :-) That is as far as I got too...   I was able to setup an elements node, but "generatetoaddress" just didn't work. So I was unable to create funds or other assets.  Also linking elements to the bitcoin regtest node failed for me.....

moneymanolis:
> OK. :-) That is as far as I got too... I was able to setup an elements node (in regtest), but "generatetoaddress" just didn't work. So I was unable to create funds or other assets. Also linking elements to the bitcoin regtest node failed for me.....

You can run elements "in isloation" (not connected to Bitcoin Core) if you use `validatepegin=0` in the elements.conf.
Yeah, generatetoaddress doesn't work for me either. But, there are also just faucets for liquidtestnet to play around with assets and LBTC / to have some of them.

I am just posting a sample conf here (the testnet part is from a blockstream link):

```
# Running elementsd without chain parameter will run regtest
chain=elementsregtest

validatepegin=0
server=1
rpcuser=michaelsaylor
rpcpassword=mstr

# Regtest settings
[elementsregtest]
rpcport=18884

# Testnet settings:
[liquidtestnet]
rpcport=18891
txindex=1
anyonecanspendaremine=0
initialfreecoins=2100000000000000
con_dyna_deploy_start=0
con_max_block_sig_size=150
checkblockindex=0
addnode=liquid.network:18444
addnode=liquid-testnet.blockstream.com:18892
addnode=liquidtestnet.com:18891
fallbackfee=0.00000100
con_has_parent_chain=0
parentgenesisblockhash=NULL
pubkeyprefix=36
scriptprefix=19
blindedprefix=23
bech32_hrp=tex
blech32_hrp=tlq
pchmessagestart=410edd62
dynamic_epoch_length=1000
signblockscript=51210217e403ddb181872c32a0cd468c710040b2f53d8cac69f18dad07985ee37e9a7151ae
evbparams=dynafed:0:::
```

moneymanolis:
After talking with Stepan, we need to ensure that the we keep the status quo on master, which is:

Show BTC/sat if there are **no** assets, and show select box with **all** assets if they are any, which looks like this:

grafik

relativisticelectron:
Thanks for the elements config.   I can now send tractors and the unit selector works well.
image

moneymanolis:
Yes, the PR works in the same way as on master, but I am wondering whether it would not be better to have at least a separate unit logic if a Liquid wallet has ONLY LBTC (probably the most likely case for a user currently), I just hacked it to show:

grafik

Basically an elif (`{% elif specter.is_liquid and wallet.balance["assets"] == {} %}`) in `addAssetSelectorToShadowRoot()`
plus adjusting `updateUnitLabelAndStep() `

IMO this could also be done in a separate PR. Just checking whether it makes sense at all.

moneymanolis:
@relativisticelectron regarding https://github.com/cryptoadvance/specter-desktop/pull/1782/commits/267cb9942dec78fc422006da27c9d8b639ead9ed initiating `addRecipient`with amount = 0 made the Cypress type command basically append `X` to have an amount `0X` in the tests. This made the tests flaky / failing. Also, this made the placeholder of 0 pointless. Just double-checking with you whether you somehow wanted to have the 0.

relativisticelectron:
> @relativisticelectron regarding [267cb99](https://github.com/cryptoadvance/specter-desktop/commit/267cb9942dec78fc422006da27c9d8b639ead9ed) initiating `addRecipient`with amount = 0 made the Cypress type command basically append `X` to have an amount `0X` in the tests. This made the tests flaky / failing. Also, this made the placeholder of 0 pointless. Just double-checking with you whether you somehow wanted to have the 0.

Thanks for fixing that.

PR #1785, opened 2022-07-01
WIP: Implement JWT authentication in REST API

ankur12-1610:
This PR basically involves the implementation of JWT authentication in Specter's REST API, I've divided this PR into different milestones:

- [x] Add `jwt_token` field in the `UserMixin`
- [x] Create basic endpoints( `/v1alpha/token`) which return the  `jwt_token` as the JSON response.
- [x] Add `BasicHTTPAuth` to the endpoints
- [x] Implement `TokenHTTPAuth` on top of the `BasicHTTPAuth`  
- [x] Add token life functionality, so that user can decide the life of the token generated
- [x] One time access functionality, which basically will limit user to see the token only once after generation

Demo (latest progress):

https://user-images.githubusercontent.com/76884959/186760671-3f5e06e3-d5dc-4fbd-8160-8e4d012d1526.mp4


src/cryptoadvance/specter/api/rest/api.py
```diff
@@ -16,6 +16,7 @@
 from ...util.fee_estimation import get_fees
 
 from .. import auth
+from .jwt import TokenResource
```
k9ert, line 19: Please apply the naming conventions. Check the other resources for examples. In that case you need to rename the file.

src/cryptoadvance/specter/api/rest/jwt.py
```diff
@@ -0,0 +1,58 @@
...
+        "exp": datetime.datetime.utcnow() + datetime.timedelta(days=1),
+    }
+    return jwt.encode(payload, app.config["SECRET_KEY"], algorithm="HS256")
+
+
+@rest_resource
+class TokenResource(AdminResource):
+    endpoints = ["/v1alpha/token/"]
+
+    def get(self):
```
k9ert, line 31: The REST standard defines that a get-request does not change the state of the resource but in that case, you're creating a token. So please make change it to a post-request. As the token should only be handed out once. The corresponding get-reuqest should probably return a 404 if the token does not exist and an empty response if it does.

src/cryptoadvance/specter/api/rest/jwt.py
```diff
@@ -0,0 +1,58 @@
...
+            jwt_token = return_dict["jwt_token"]
+            user_details.save_jwt_token(jwt_token)
+            return {
+                "message": "Token generated",
+                "username": user_details.username,
+                "jwt_token": jwt_token,
+            }
+        return {"message": "Token already exists", "jwt_token": jwt_token}
+
+    def delete(self):
```
k9ert, line 51: Also here you should return proper http-codes. Get yourself a bit familiar with the main codes. If the token does not exist, return 404. See the healthz-endpoints for examples to specify the http-code while returning.

src/cryptoadvance/specter/api/rest/jwt.py
ankur12-1610: > The REST standard defines that a get-request does not change the state of the resource but in that case, you're creating a token. So please make change it to a post-request. As the token should only be handed out once. The corresponding get-reuqest should probably return a 404 if the token does not exist and an empty response if it does.
> 
> I think it makes sense to store more than one token, just right from the beginning. So maybe you design the REST endpoint for that. On the user-side, maybe store them in a hash so that each token gets an ID which needs to be specified by the user with the post-reuqest.

@k9ert I've resolved the first part of the review and have added the second part in the recent push, kindly check ;)

src/cryptoadvance/specter/api/rest/resource_jwt.py
```diff
@@ -0,0 +1,95 @@
...
+        "user": user.username,
+        "exp": datetime.datetime.utcnow() + datetime.timedelta(days=1),
+    }
+    return jwt.encode(payload, app.config["SECRET_KEY"], algorithm="HS256")
+
+def generate_token_id():
+    return str(uuid.uuid4())
+
+@rest_resource
+class ResourceJWT(SecureResource):
```
k9ert, line 29: Can you please document all classes, e.g. here is an example:
```suggestion
class ResourceJWT(SecureResource):
    ''' A Resource to manage JWT tokens in order to authenticate against the REST-API
      Other then the other Resources, this endpoint uses BasicAuth to avoid the chicken egg problem
    '''
```

src/cryptoadvance/specter/user.py
```diff
@@ -70,13 +70,19 @@ def __init__(
         password_hash,
         config,
         specter,
+        jwt_token_id,
+        jwt_token,
```
k9ert, line 74: Just a reminder that this can be deleted. It can even be done now, no?

src/cryptoadvance/specter/user.py
```diff
@@ -70,13 +70,19 @@ def __init__(
...
+        tokens = {},
```
k9ert, line 75: can you give it a more specific name like `jwt_tokens` ?!

src/cryptoadvance/specter/user.py
ankur12-1610: > Just a reminder that this can be deleted. It can even be done now, no?

By deleting mean do you mean the `delete` route?

ankur12-1610:
@k9ert can you review the changes?

src/cryptoadvance/specter/user.py
k9ert: No, i mean that you delete line 73 and 74 and all the occurrences of those. You now have a `jwt_tokens` and you no longer need them, right?!

src/cryptoadvance/specter/api/rest/resource_jwt.py
```diff
@@ -0,0 +1,102 @@
...
+    ''' 
+      A Resource to manage JWT tokens in order to authenticate against the REST-API
+      Other then the other Resources, this endpoint uses BasicAuth to avoid the chicken egg problem
+      This one is only to create a token. The other Resource for getting and deleting.
+      This violates the REST principles but only on the implementation-side. Happy for improvements here.
+    '''
+    endpoints = ["/v1alpha/token/"]
+
+    def post(self):
```
k9ert, line 38: this is suppose to be a get-request, no?! Returning `jwt_tokens` directly also expose the secret, no?!

src/cryptoadvance/specter/api/rest/resource_jwt.py
```diff
@@ -0,0 +1,102 @@
...
+                "created_jwt_token": jwt_token,
+            }, 201
+
+@rest_resource
+class ResourceJWTById(SecureResource):
+    ''' 
+        A Resource to manage individual JWT token
+    '''
+    endpoints = ["/v1alpha/token/<jwt_token_id>/"]
+
```
k9ert, line 81: Maybe for the completeness, you should have a get-request here as well? It would return the description (the human readable one) only but that's still worth to do imho.

src/cryptoadvance/specter/user.py
```diff
@@ -200,6 +209,28 @@ def delete_user_secret(self, autosave: bool = True):
         if autosave:
             self.save_info()
 
+    def get_all_tokens(self):
+        self.jwt_tokens = json.dumps(self.jwt_tokens.copy())
+        self.jwt_tokens = json.loads(self.jwt_tokens)
+        return self.jwt_tokens
+        
+    def append_token(self, jwt_token_id, jwt_token):
+        self.jwt_tokens = self.jwt_tokens.copy()
```
k9ert, line 218: What's the purpose of that copy-statement?

src/cryptoadvance/specter/api/rest/resource_jwt.py
```diff
@@ -0,0 +1,102 @@
...
+            "id": user_details.id,
+            "jwt_token_id": jwt_token_id,
+            "jwt_token": jwt_token,
+        }
+        return_dict["jwt_token_id"] = generate_token_id()
+        return_dict["jwt_token"] = generate_jwt(user_details)
+        jwt_token_id = return_dict["jwt_token_id"]
+        jwt_token = return_dict["jwt_token"]
+        user_details.save_jwt_token(jwt_token_id, jwt_token)
+        user_details.append_token(jwt_token_id, jwt_token)
```
k9ert, line 67: The `id` is part of the dict so you don't need to pass the id in a parameter, right?!

src/cryptoadvance/specter/api/rest/resource_jwt.py
```diff
@@ -0,0 +1,102 @@
...
+        if len(jwt_tokens) == 0:
+            return {"message": "Token does not exist"}, 404
+        return {"message": "Tokens exists", "jwt_tokens": jwt_tokens}, 200
+
+    def post(self):
+        user = auth.current_user()
+        user_details = app.specter.user_manager.get_user(user)
+        jwt_token = user_details.jwt_token
+        jwt_token_id = user_details.jwt_token_id
+        return_dict = {
```
k9ert, line 56: Why not creating a function which completely create the dict and call that?

src/cryptoadvance/specter/api/rest/resource_jwt.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def post(self):
+        user = auth.current_user()
+        user_details = app.specter.user_manager.get_user(user)
+        jwt_tokens = user_details.jwt_tokens
+        return_dict = {
+            "jwt_tokens": jwt_tokens,
+        }
+        return_dict["jwt_tokens"]=user_details.get_all_tokens()
+        jwt_tokens = return_dict["jwt_tokens"]
```
k9ert, line 46: This looks very strange. You create a dict, fill it with some key and then assign the value for that exact key to another variable?!

k9ert:
I've added some more comments. Please also fix the tests. Check the development documentation for something called `pip-compile`.

src/cryptoadvance/specter/api/rest/resource_jwt.py
ankur12-1610: Done via nested hashmap

src/cryptoadvance/specter/api/rest/resource_jwt.py
ankur12-1610: Done

ankur12-1610:
> I've added some more comments. Please also fix the tests. Check the development documentation for something called `pip-compile`.

I've done the same thing but I don't know why all the tests are failing due a `backport` error

ankur12-1610:
@k9ert @moneymanolis I've updated the PR, the changes made are:
- Removed irrelevant variables like `jwt_token` and `jwt_token`id` , all the info is stored in a nested hashmap
- The hashmap has the following structure:
```
 "jwt_tokens": {
       "jwt_token_id": {
            "jwt_token": "a human readable value",
            "jwt_token_description": "a description give to the token"
      }
}
```

k9ert:
```
 "jwt_tokens": {
       "jwt_token_id": {
            "jwt_token": "a human readable value",
            "jwt_token_description": "a description give to the token"
      }
}
```
This is quite confusing to me.  I've mentioned this example of a structure can not match my example to your description here:
```
{
  "e93ef5d0-e35c-431e-b711-ddf5e6966e6c": {
    "description": "a human readable description",
    "token": "a very long token here"
  },
  "162ec91c-3cdd-4d87-9e67-44a6c4ac1453": {
    "description": "yet another human readable description",
    "token": "a very long other token here"
  }
}
```

ankur12-1610:
> ```
>  "jwt_tokens": {
>        "jwt_token_id": {
>             "jwt_token": "a human readable value",
>             "jwt_token_description": "a description give to the token"
>       }
> }
> ```
> This is quite confusing to me.  I've mentioned this example of a structure can not match my example to your description here:
> ```
> {
>   "e93ef5d0-e35c-431e-b711-ddf5e6966e6c": {
>     "description": "a human readable description",
>     "token": "a very long token here"
>   },
>   "162ec91c-3cdd-4d87-9e67-44a6c4ac1453": {
>     "description": "yet another human readable description",
>     "token": "a very long other token here"
>   }
> }
> ```

It is same, can you watch video please, the structure is same, I've just declared the parent hashmap a name which is `jwt_tokens`

ankur12-1610:
> ```
>  "jwt_tokens": {
>        "jwt_token_id": {
>             "jwt_token": "a human readable value",
>             "jwt_token_description": "a description give to the token"
>       }
> }
> ```
> 
> This is quite confusing to me. I've mentioned this example of a structure can not match my example to your description here:
> 
> ```
> {
>   "e93ef5d0-e35c-431e-b711-ddf5e6966e6c": {
>     "description": "a human readable description",
>     "token": "a very long token here"
>   },
>   "162ec91c-3cdd-4d87-9e67-44a6c4ac1453": {
>     "description": "yet another human readable description",
>     "token": "a very long other token here"
>   }
> }
> ```

The map looks like this:
```
        "jwt_tokens": {
            "7e3e9f44-d867-4573-9058-1f00399f1fdd": {
                "jwt_token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VyIjoiYWRtaW4iLCJleHAiOjE2NTkwNjI5OTF9.-EhCintG94jGf3DZGnke4Rs_v1cUlocJQ2QmZuWPGts",
                "jwt_token_description": "test2"
            },
            "e51f4fe7-f5a8-4a4b-92d8-e404bb613951": {
                "jwt_token": "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJ1c2VyIjoiYWRtaW4iLCJleHAiOjE2NTkwODY3Nzl9.3HshQQGQ-eDes9SkFE7oNnDXud76aymyV2Z6QS8VJRI",
                "jwt_token_description": "test3"
            }
        }
```

requirements.txt
```diff
@@ -1,8 +1,8 @@
 #
-# This file is autogenerated by pip-compile
+# This file is autogenerated by pip-compile with python 3.10
```
moneymanolis, line 2: Python 3.10 is not supported yet in Specter, this is probably why all your tests are failing. Use pyenv to manage python versions.

requirements.txt
```diff
@@ -1,8 +1,8 @@
...
 # To update, run:
 #
-#    pip-compile --generate-hashes requirements.in
+#    pip-compile --allow-unsafe --generate-hashes requirements.in
```
moneymanolis, line 5: This flag was not used so far. Check the current requirements.txt file.

requirements.txt
```diff
@@ -16,26 +16,6 @@ babel==2.9.1 \
     --hash=sha256:ab49e12b91d937cd11f0b67cb259a57ab4ad2b59ac7a3b41d6c06c0ac5b0def9 \
     --hash=sha256:bc0c176f9f6a994582230df350aa6e05ba2ebe4b3ac317eab29d9be5d2768da0
     # via flask-babel
-backports.zoneinfo==0.2.1 \
```
moneymanolis, line 19: This whole transitive  dependency got deleted, no wonder the tests don't work.

moneymanolis:
Tested this draft in local dev, works fine. One point I came across is that the used request parser of Flask-RESTful will be removed, see this [link](https://flask-restful.readthedocs.io/en/latest/reqparse.html) - we might think about already using sth. else here.

ankur12-1610:
@k9ert @moneymanolis in the latest push I've added the `token_verification` functionality now we can fully implement the jwt authorization to any endpoint. The only part which remains now is to add the lifespan of the token and to add a check if the token is expired and if it is expired then we'll warn the user and eventually delete the token.

I faced an issue implementing token verification which was - mapping a token with its id i.e. we want to make sure that the user only gets access to that token to which the corresponding route is entered. For eg:

The route is `/v1alpha/token/tokenid1` so it should only give the user `token1` but in `token_verification` we can't pass 2 variables as `Bearer Authorization` takes only 1 variable as an input which is the jwt_token. 

So I came up with the solution : I'm passing token_id as well as token encoded in the single variable the format will be:
```
Bearer Token: <token_id>:token
```
I've added a check to verify the token_id and token are correct and mapped to each other if not the check throws an error :)

src/cryptoadvance/specter/api/security.py
```diff
@@ -35,6 +35,72 @@ def verify_password(username, password):
...
+            and g.user
+        ):
+            logger.info(f"Rest-Request for user {payload['user']} PASSED JWT-test")
+            return True
+    except jwt.ExpiredSignatureError:
+        return "Signature expired. Please create a new one"
+    except jwt.InvalidTokenError:
+        return "Invalid token. Please create a new one"
+
+
```
moneymanolis, line 103: This is exactly the same as above, remove.

src/cryptoadvance/specter/api/security.py
```diff
@@ -9,7 +9,7 @@
 
 # from flask_httpauth import HTTPBasicAuth
```
moneymanolis, line 10: Remove

moneymanolis:
First of all, I don't really understand why users can have multiple tokens in the first place? Why would they need that? Might be that @k9ert had sth. in mind here which I don't see, though. 

> The route is /v1alpha/token/tokenid1 so it should only give the user token1

Does not make sense to get a token with a token. If you mean to get the description (the relevant line is not clear here: ` {"message": "Tokens exists", "jwt_token_description": jwt_token}`) why wouldn't it give the corresponding description (if you use the correct key for the dictionary at least)? 

I would definitely avoid this `Bearer Token: <token_id>:token` construction since it is very counter-intuitive, I don't know of any API that uses such an approach. 

Also, the verification of the token
```
@token_auth.verify_token
def verify_token():
```
 is not implemented correctly.

Please, have look at the documentation here again:
https://flask-httpauth.readthedocs.io/en/latest/index.html

Crucially: "As with the verify_password, the function should return the **user object** if the token is valid."

I had this for example during testing:
`ERROR in user_manager: Could not find user Invalid token. Please create a new one`

which stems from this line returning "Invalid token. Please create a new one" as user.

```
   except jwt.InvalidTokenError:
        return "Invalid token. Please create a new one"
```

ankur12-1610:
> First of all, I don't really understand why users can have multiple tokens in the first place? Why would they need that? Might be that @k9ert had sth. in mind here which I don't see, though.
> 
> > The route is /v1alpha/token/tokenid1 so it should only give the user token1
> 
> Does not make sense to get a token with a token. If you mean to get the description (the relevant line is not clear here: ` {"message": "Tokens exists", "jwt_token_description": jwt_token}`) why wouldn't it give the corresponding description (if you use the correct key for the dictionary at least)?
> 
> I would definitely avoid this `Bearer Token: <token_id>:token` construction since it is very counter-intuitive, I don't know of any API that uses such an approach.
> 
> Also, the verification of the token
> 
> ```
> @token_auth.verify_token
> def verify_token():
> ```
> 
> is not implemented correctly.
> 
> Please, have look at the documentation here again: https://flask-httpauth.readthedocs.io/en/latest/index.html
> 
> Crucially: "As with the verify_password, the function should return the **user object** if the token is valid."
> 
> I had this for example during testing: `ERROR in user_manager: Could not find user Invalid token. Please create a new one`
> 
> which stems from this line returning "Invalid token. Please create a new one" as user.
> 
> ```
>    except jwt.InvalidTokenError:
>         return "Invalid token. Please create a new one"
> ```

@moneymanolis, it gives the corresponding description of the given token public id and token, in which token id is the key of the stored token object.

And regarding the token verification implementation, we can discuss it in more detail, coz I first referred to the approach mentioned in the doc but it didn't verify the user it only verified the token and I thought doing it as password verification is done made more sense:sweat_smile:

src/cryptoadvance/specter/api/rest/resource_jwt.py
```diff
@@ -0,0 +1,133 @@
...
+    return jwt.encode(payload, app.config["SECRET_KEY"], algorithm="HS256")
+
+
+def generate_token_id():
+    # Generates a unique token id
+    return str(uuid.uuid4())
+
+
+@rest_resource
+class ResourceJWT(BasicAuthResource):
```
k9ert, line 46: All the Resource-classes are called `SomethingResource` but this one is called `ResourceJWT`. Can you rename it to `JwtResource` ?

src/cryptoadvance/specter/api/rest/resource_jwt.py
```diff
@@ -0,0 +1,133 @@
...
+                "jwt_token_life": jwt_token_life,
+            }, 201
+        else:
+            return {
+                "message": "Token description already exists, please enter a new one"
+            }, 400
+
+
+@rest_resource
+class ResourceJWTById(SecureResource):
```
k9ert, line 94: Ah, same here?

ankur12-1610:
@k9ert @moneymanolis 
I've completed all the milestones mentioned in the PR. The details of the PR is as follows:
### Endpoints
- There are 2 endpoints one is `/v1alpha/token` and other is `/v1alpha/token/<token_id>`. 
- Token can be created at `/v1alpha/token` with the body by a 'POST' request as follows: 
```
"jwt_token_description": "dummy data",
"jwt_token_life": integer,
"jwt_token_life_unit": "days or minutes or seconds"
```
-  `/v1alpha/token` also has 'GET' request which gives info of all the tokens created by the user
-  `/v1alpha/token/<token_id>` endpoint when given a `POST` request gives information of the particular token with `<token_id>` passed in the endpoint URL
- /v1alpha/token/<token_id>` endpoint when given a `DELETE` request deletes the respective token of the `<token_id>` entered.

### Auth
There are two types of authentication resources, one is `BasicAuthResource` which includes basic httpauth and other one is `SecureResource` which includes token based authentication. So any endpoint who implements SecureResource` can only be accessed by token based authentication.

### Expiry time parser
- One problem I noticed while setting sexpiry time was that user could only set expiry time in either seconds or days or months and not all of them at one time, the solution I came up with to give user the flexibility is by taking `jwt_token_life_unit` as an additional input from user which then is concatenated with the integral time and the parsed using [pytimeparse](https://pypi.org/project/pytimeparse/).

- Working:
For eg:
```
"jwt_token_description": "token1",
"jwt_token_life": 6,
"jwt_token_life_unit": "days"
```
the above example has integral value as `6` and unit as `days`, they are concatenated and converted to string to give the result as `6 days` . According to https://pypi.org/project/pytimeparse/, `6 days` is then converted to seconds and hence all the units are converted to seconds and the user can happily set the token expiry time in days, minutes, seconds or even weeks.

BEFORE TESTING THE PR PLEASE WATCH THE DEMONSTRATION VIDEO MENTIONED IN THE PR FOR BETTER VISUALISATION AND UNDERSTANDING ;)

moneymanolis:
@ankur12-1610 tests are red.

ankur12-1610:
> @ankur12-1610 tests are red.
> 
> Also: Specter Resource and TXlist Resource are still AdminResources which has those decorators: `method_decorators = [error_handling, require_admin, auth.login_required] `
> 
> Thus, no `token_auth.login_required` there. Shall we change the AdminResource?

@moneymanolis regarding the tests we discussed in the meeting, I tried to resolve them but were failing, can you look into it?
Yeah we can add `token_auth.login_required` to the `AdminResource` as well, should I add it?

k9ert:
> I tried to resolve them but were failing, can you look into it?

Obviously the tests are still using BasicAuth and need to be adjusted to the new Authentication mechanism, right?! I'm niot sure what's the issue for you to fix these tests?

Also please review the documentation for Rest and adjust it as necessary.

Please also adjust the storage in the json-file: It shouldn't be necessary to parse something and make calculations in order to figure out when the token will expire. So ideally you just store a unix-timestamp of the expiry date in there or the number of seconds it is valid (and you calculate validity via the creation-date). I think this is the best option.

Also please adjust the `current_status` field. It should be machine readable without too much parsing and without redundancy. So simply `valid` or `expired` should imho be the two reasonable values for that field.

k9ert:
The rest-Tests are still failing?! Please fix the tests. Will continue with the review after that!

k9ert:
Still no documentation in the docs folder for this PR, right?!

ankur12-1610:
> Still no documentation in the docs folder for this PR, right?!

Yeah, I need to add the details of the new endpoints, I'll them:sweat_smile:

ankur12-1610:
@k9ert I've added the docs although I'm not sure if we can describe two endpoints in one file itself but the two endpoints were similar so I added them to a single file.

Can you verify it?

k9ert:
> Can you verify it?

That's completely fine! However the [basic-usage section](https://deploy-preview-1785--specter-desktop-docs.netlify.app/api/#basic-usage) is still outdated. Can you bring that up-to-date and add the endpoint to the list there as well?

docs/api/README.md
```diff
@@ -8,18 +8,37 @@ export SPECTER_API_ACTIVE=True
...
+```
+
+## Token Based Usage
+> Works for the endpoints with `SecureResource` and `AdminResource`
+
+Curl:
+
+```bash
+curl --location --request GET 'http://127.0.0.1:25441/api/v1alpha/specter' \
+--header 'Authorization: Bearer <token>' | jq .
```
k9ert, line 34: Please also add the second request which is the one the developer is interested about. Only obtaining the Token adds no Business Value and knowing how to exactly using the token is really critical (and potentially quire time-consuming).

docs/api/README.md
```diff
@@ -8,18 +8,37 @@ export SPECTER_API_ACTIVE=True
...
+```
+
+Python:
+
+```python
+import requests
+response = requests.get('http://127.0.0.1:25441/api/v1alpha/specter', headers={'Authorization': 'Bearer <token>'})
 json.loads(response.text)
```
k9ert, line 42: Same here! Please explain how to use the token, not only how to obtain it.

docs/api/README.md
ankur12-1610: > Please also add the second request which is the one the developer is interested about. Only obtaining the Token adds no Business Value and knowing how to exactly using the token is really critical (and potentially quire time-consuming).

@k9ert in this request I'm actually using token to access the `/v1alpha/specter` route.
In the `Basic Usage` section mentioned the user gets the token and in the `Token Usage` section the token is being used.

Isn't this complete:sweat_smile: info?

k9ert:
If there is no hiccup, this should now be green. I've also refined the documentation. @ankur12-1610 currently no need to do anything on your side other than maybe reviewing my changes in the documentation.
@moneymanolis This is ready for review.

ankur12-1610:
Oh! @k9ert you updated the doc:sweat_smile:, I was doing it rn ;)
Thanks for it ;)

Also in the updated version, we've not described python requests any reason for that:thinking:?

ankur12-1610:
> Yeah, i was a bit impatient sweat_smile
> 
> > Also in the updated version, we've not described python requests any reason for thatthinking?
> 
> What's missing here ? https://deploy-preview-1785--specter-desktop-docs.netlify.app/api/#python

Oh yeah it is correct I thought we the same curl should be in python requests as well didn't see the line you mentioned which says in python requests we won't use `BasicAuth`.

Also there is a minor typo over here https://deploy-preview-1785--specter-desktop-docs.netlify.app/api/#python:~:text=%3Ctoken%3E%20and-,%3Ctoekn_id%3E, it should be `<token_id>`.

ankur12-1610:
> Typo fixed.
> 
> It might need an explanation: You don't "need" BasicAuth in python because we assume here, that people who are doing python are developing serious integration-solutions which are configurable. Configuration with a token which has been created upfront via ... remember: Ideally we should have done a frontend in order to create the token. We wanted to make it as simple as possible and therefore we skipped the UI and instead created an API endpoint for that as a first fix.
> 
> However this (BasicAuth) endpoint should only be used by an administrator of Specter in order to create the token which they then configure in some other (python-) application. But an admin will use curl for that, not python.
> 
> So that's the reason why we don't need the BasicAuth endpoint example in python (even though it would be possible).
> 
> Or looking it from the other point of view: If you WOULD implement a configuration where you configure Username/Password into the python application and the python application is getting a token and using that one, then the whole exercise would have been completely useless and no security gains would be there. Therefore: Implementing the BasicAuth example in python is even bad practice.

Got it, thanks for the info I didn't have any idea about this ;)

k9ert:
We can soon merge this PR. I don't think there is anything left other than doing a normal bugfix release, probably on monday. Next week tuesday afternoon, we can merge this.

moneymanolis:
There were some little changes of the rest api test in this https://github.com/cryptoadvance/specter-desktop/pull/1920, I'd suggest to merge this PR first and then check whether the tests are still green (I guess so!) and then merge.

k9ert (approved):

k9ert:
@ankur12-1610 merged, thank you !

moneymanolis:
Congrats! @ankur12-1610

ankur12-1610:
Yaaayy! :tada:  it was a great learning experience, thanks @k9ert and @moneymanolis :smile:

PR #1786, opened 2022-07-01
Chore: fix broken cirrus build cache for bitcoind

k9ert:
The fingerprint_script is not dependent on what exactly the populate_script is or how it is executed. Both is a problem:
in #1780 i've created a new task where i used `--compile` instead of `--binary`.  As a result, it seems that the cache got overridden by a compiled version which affects all branches.
This:
```
./tests/install_noded.sh --debug --bitcoin binary
```
is very different from this:
```
./tests/install_noded.sh --debug --bitcoin compile
```

This should be reflected in the cache key, especially if you want to use both modes at the same time (as in #1780 )

Also there were some issues with the new `--binary` mode of the elementsd version of the install_noded.sh script.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -50,7 +50,6 @@ prep_stuff_template: &PREP_STUFF_TEMPLATE
     - source ./.env/bin/activate
     - pip3 install -r requirements.txt --require-hashes && pip3 install -r test_requirements.txt
   install_script: 
-    - ls -l ./tests/elements/src/ 
     - source ./.env/bin/activate
     - pip3 install -e .
 
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -22,7 +22,7 @@ prep_stuff_template: &PREP_STUFF_TEMPLATE
       - cat tests/elements_gitrev_pinned 2> /dev/null || true
       - cat /etc/os-release | grep VERSION
       - cat ./tests/install_noded.sh
-      - echo "binary" # if the next line is --binary, otherwise compile. Not doing that correctly will break a proper caching switching between both
+      - echo "binary" # if the next line is --elements binary, otherwise use echo "compile" - this ensures different caching keys.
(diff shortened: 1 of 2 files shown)


.cirrus.yml
```diff
@@ -12,19 +12,25 @@ prep_stuff_template: &PREP_STUFF_TEMPLATE
       - cat pytest.ini | grep "addopts = " | cut -d'=' -f2 |  sed 's/--/+/g' | tr '+' '\n' | grep bitcoin |  cut -d' ' -f2 
       - cat tests/bitcoin_gitrev_pinned 2> /dev/null || true
       - cat /etc/os-release | grep VERSION
+      - cat ./tests/install_noded.sh
+      - echo "binary" # if the next line is --binary, otherwise compile. Not doing that correctly will break a proper caching switching between both
```
moneymanolis, line 16:
```suggestion
      - echo "binary" # if the next line is --bitcoin binary, otherwise use echo "compile" - this ensures different caching keys.
```

.cirrus.yml
```diff
@@ -12,19 +12,25 @@ prep_stuff_template: &PREP_STUFF_TEMPLATE
...
     populate_script: ./tests/install_noded.sh --debug --bitcoin binary
   elementsd_installation_cache:
     folder: ./tests/elements
     fingerprint_script: 
       - cat pytest.ini | grep "addopts = " | cut -d'=' -f2 |  sed 's/--/+/g' | tr '+' '\n' | grep elements |  cut -d' ' -f2 
       - cat tests/elements_gitrev_pinned 2> /dev/null || true
       - cat /etc/os-release | grep VERSION
+      - cat ./tests/install_noded.sh
+      - echo "binary" # if the next line is --binary, otherwise compile. Not doing that correctly will break a proper caching switching between both
```
moneymanolis, line 25:
```suggestion
      - echo "binary" # if the next line is --elements binary, otherwise use echo "compile" - this ensures different caching keys.
```

PR #1790, opened 2022-07-05
Fix run the numbers

k9ert:
* Using a lock to prevent concurrent calls (at all)
* Expose errors to the user
* Treat the special case of bitcoind running hot with queue full


src/cryptoadvance/specter/templates/includes/sidebar/components/bitcoin_core_info.jinja
```diff
@@ -50,17 +50,28 @@
...
                             method: 'GET'
                         }
-                    );
+                    ).catch((err) => {
+                        showError(err)
+                        return
+                    });
                     let result = await response.json();
                     console.log(result)
-                    let userBalanceFromTotal = parseFloat((100 / (result.total_amount / totalUserBalance)).toFixed(8));
```
moneymanolis, line 56: `userBalanceFromTotal` isn't defined then anymore, no? There needs to be an if clause to avoid dividing by 0 if `totalUserBalance` ist zero.

Issue #1794, opened 2022-07-11
UI Bug: The Address table header isn't rendered correctly


relativisticelectron:
**Describe the bug**
The table header looks like this
![image](https://user-images.githubusercontent.com/60378539/178193042-a3c61b68-2047-40b9-b33a-1125e208d8f2.png)

There is the filter missing (is that intended?), and the "Page Limit:" is lonely on the left, while the actual page Limit selector is on the right...

**To Reproduce**
1. Go to `/wallets/wallet/<wallet>/addresses/` 
 
**Expected**
![image](https://user-images.githubusercontent.com/60378539/178193379-8ee168c9-98cb-4a63-acce-dbb9032b7319.png)

**Desktop (please complete the following information):**
 - Linux, local, current master

Tagging @moneymanolis

moneymanolis:
There was no filter / search bar on the address list so far, which "makes sense" since it is not implemented as a web component as of yet so it was not some easy `{% include ... %}`.

I've fixed the alignment bug in a PR (#1779) that I will finalise soon.

relativisticelectron:
> There was no filter / search bar on the address list so far, which "makes sense" since it is not implemented as a web component as of yet so it was not some easy `{% include ... %}`.
> 
> I've fixed the alignment bug in a PR (#1779) that I will finalise soon.

Ok, I didn't know that there is no address filter. That might me a good feature in the future.  Thanks for fixing the alignment.

Issue #1593, opened 2022-02-18
UI Feature request: Global search


relativisticelectron:
**Problem**:  I wanted to search an address in specter and found no easy way of doing it.   
**Counter example**: I see friends using gmail. You can search globally all mailboxes for anything and gives a very user friendly experience.

**Feature request**:  A global (all wallets) search over 

- Addresses 
- transactions, id, label, time (date), amount
- device names, wallet names, service/plugin names
- (optionally) settings.  So you could search for "export" or "label import", and you would quickly find the menu where it is hidden

**Mockup**  (and yes this is regtest.... )
![Untitled](https://user-images.githubusercontent.com/60378539/154636741-47e56762-aa84-453f-b497-dad69ebde7b7.png)

**Some design elements:**

- Visual highlighting of the menu buttons under which search results where found, and how many in this sub-category
- Filtering of the tx, utxo, or address list to show only hits

**Replaces**  the current search for the transactions/utxos

moneymanolis:
@agg-pranav want to give it a shot?

relativisticelectron:
I am working on a branch on this https://github.com/relativisticelectron/specter-desktop/tree/20220710_search
![image](https://user-images.githubusercontent.com/60378539/178160421-01b8746c-d03a-4ddf-8b46-20bd26b9e086.png)

relativisticelectron:
PR exists : https://github.com/cryptoadvance/specter-desktop/pull/1832

PR #1795, opened 2022-07-11
UI Feature: Global search across tx, addresses, amounts, xpubs

relativisticelectron:
See: https://github.com/cryptoadvance/specter-desktop/issues/1593

For me most important is the feature to quickly search all wallets for an address.

Peek 2022-07-11 16-33

TODO:
- [x] https://github.com/cryptoadvance/specter-desktop/issues/1794 has to be fixed first, so searches in addresses work
- [ ] send cleaner json via request 
- [ ] Prevent searching in tx fields, that are not visible to the user in tx-table
- [ ] Make the search entry persistent when switching sites,  or maybe have search results in a clickable dropdown list?

If you have **feedback**, e.g., how to make the UI nicer, let me know.  
- Currently I do not know how to make the search_term persist when loading a different page

### DIFF
--- a/src/cryptoadvance/specter/global_search.py
+++ b/src/cryptoadvance/specter/global_search.py
@@ -243,4 +243,8 @@ def do_global_search(search_term, specter):
     if search_term:
         apply_search_on_dict(search_term, HTML_ROOT)
     print(HTML_ROOT.flattened_sub_tree_as_json())
-    return {"tree": HTML_ROOT, "list": HTML_ROOT.flattened_sub_tree_as_json()}
+    return {
+        "tree": HTML_ROOT,
+        "list": HTML_ROOT.flattened_sub_tree_as_json(),
+        "search_term": search_term,
+    }
--- a/src/cryptoadvance/specter/global_search.py
+++ b/src/cryptoadvance/specter/global_search.py
@@ -92,6 +92,14 @@ def build_html_elements(specter):
 
     def add_all_in_wallet(wallet):
         sidebar_wallet = HtmlElement(wallets, id=f"{wallet.alias}-sidebar-list-item")
+        wallet_names = HtmlElement(
+            sidebar_wallet,
+            id="title",
+            function=lambda x: search_in_structure(x, [wallet.alias]),
+            visible_on_endpoints=[
+                url_for("wallets_endpoint.wallet", wallet_alias=wallet.alias)
+            ],
+        )
         transactions = HtmlElement(
             sidebar_wallet,
             id="btn_transactions",
--- a/src/cryptoadvance/specter/global_search.py
+++ b/src/cryptoadvance/specter/global_search.py
@@ -213,6 +213,9 @@ def build_html_elements(specter):
             sidebar_device,
             id="title",
             function=lambda x: search_in_structure(x, [device.alias]),
+            visible_on_endpoints=[
+                url_for("devices_endpoint.device", device_alias=device.alias)
+            ],
         )
         device_keys = HtmlElement(
             sidebar_device,
--- a/src/cryptoadvance/specter/global_search.py
+++ b/src/cryptoadvance/specter/global_search.py
@@ -56,12 +56,12 @@ class HtmlElement:
             node.result = None
 
     def flattened_sub_tree_as_json(self):
-        result_list = [self.to_json()]
+        result_list = [self.json()]
         for child in self.children:
             result_list += child.flattened_sub_tree_as_json()
         return result_list
 
-    def to_json(self):
+    def json(self):
         d = {}
         d["id"] = self.id
         d["children"] = self.children
@@ -75,6 +75,13 @@ HTML_ROOT = None
 
 
 def build_html_elements(specter):
+    """
+    This builds all HtmlElements that should be highlighted during a search.
+    It also encodes which functions will be used for searching.
+
+    Returns:
+        HtmlElement: This is the html_root, which has all children linked inside
+    """
     html_root = HtmlElement(None)
     wallets = HtmlElement(html_root, id="toggle_wallets_list")
     devices = HtmlElement(html_root, id="toggle_devices_list")
@@ -224,6 +231,7 @@ def build_html_elements(specter):
 
 
 def apply_search_on_dict(search_term, html_root):
+    "Given an html_root it will call the child.function for all childs that do not have any children"
     end_nodes = html_root.calculate_end_nodes()
     for end_node in end_nodes:
         end_node.result = end_node.function(search_term)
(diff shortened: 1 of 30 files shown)


relativisticelectron:
Moved the search field to the sidebar
image

k9ert:
WOW, this looks awesome! Will have a look soon!

src/cryptoadvance/specter/global_search.py
```diff
@@ -0,0 +1,260 @@
+import os
+import logging
+
+from flask import url_for
+
+logger = logging.getLogger(__name__)
+
+
+class HtmlElement:
```
k9ert, line 9: This sounds cool but i don't have an idea from that description. Maybe it's a good candidate for some tests? It looks like it's quite easy to write some tests for it and that would give a much better picture how this works.

src/cryptoadvance/specter/helpers.py
```diff
@@ -307,3 +308,17 @@ def get_address_from_dict(data_dict):
     if addr and addr != "Fee":
         return addr
     raise RuntimeError(f"Missing address info in object {data_dict}")
+
+
+def robust_json_dumps(obj):
```
k9ert, line 313: Good point. json parsing is sometimes quite annoying. But can you move that to somewhere in util? `util/json_rpc.py` would do it or a new `util/misc.py`. At some point, all that stuff in the util folder was initially in `helpers.py` and hopefully at some point the `helpers.py` vanish.

Maybe you can add a deprecation warning like 
```
Additions to helpers.py are deprecated. Please place your helpers in utils/somewhere
```

relativisticelectron: Done

src/cryptoadvance/specter/global_search.py
k9ert: Preserving the search-term should be easy via storing it in the session?! I try to get out of the way of UI judgement. Don't have a good track record there. @moneymanolis  @b30wulffz what do you think?

src/cryptoadvance/specter/global_search.py
```diff
@@ -0,0 +1,260 @@
...
+    def result(self):
+        if self._result:
+            return self._result
+        else:
+            return sum([child.result for child in self.children])
+
+    @result.setter
+    def result(self, result):
+        if self.children:
+            raise "Setting results is only allowed for end nodes"
```
k9ert, line 41: Raising a string? Shouldn't you raise an Exception or a descendant from an exception?

src/cryptoadvance/specter/global_search.py
```diff
@@ -0,0 +1,260 @@
...
+    "Given an html_root it will call the child.function for all childs that do not have any children"
+    end_nodes = html_root.calculate_end_nodes()
+    for end_node in end_nodes:
+        end_node.result = end_node.function(search_term)
+    return html_root
+
+
+def do_global_search(search_term, specter):
+    "Builds the HTML Tree if ncessary (only do it once) and then calls the functions in it to search for the search_term"
+    global HTML_ROOT
```
k9ert, line 246: I'm not sure that's a good idea?! Is that possible with multi-user? How expensive is the build-up of the tree? Maybe we can do it every time?

src/cryptoadvance/specter/global_search.py
```diff
@@ -0,0 +1,260 @@
...
+    It also encodes which functions will be used for searching.
+
+    Returns:
+        HtmlElement: This is the html_root, which has all children linked inside
+    """
+    html_root = HtmlElement(None)
+    wallets = HtmlElement(html_root, id="toggle_wallets_list")
+    devices = HtmlElement(html_root, id="toggle_devices_list")
+
+    def search_in_structure(search_term, l):
```
k9ert, line 89: Puh, this function in function-scope is imho quite horrible. Especially evil is the example below where you're not even returning a value but just changing variables which are in the scope. To me, this is a very confusing programming-model.

Shouldn't it be possible to include these methods in the HtmlElement-class?

src/cryptoadvance/specter/global_search.py
relativisticelectron: Yes, you are right, I didn't think about multi-user yet.  Please do not put in too much effort now to review the code. Before I put more work in the code (and have to change a lot), I am seeking feedback, if this  feature is wanted, and how the UI should look like.

src/cryptoadvance/specter/global_search.py
relativisticelectron: I agree, this needs a restructuring.

moneymanolis:
@relativisticelectron I'd need to first understand the goals associated with this PR before we can talk UI options etc. What do you want to search for? When is the global search used / sensible and when the search bar of the tx table?

So far, I could only gather: "feature to quickly search all wallets for an address." - which is fine if you want to double check whether an address belongs to one of your wallets and you don't remember which one. 

My rough first intuition is, that we talk about a two step approach here. (1) The global search tells you in which wallet(s) to look for, (2) and then, say, the search in the tx table takes over

relativisticelectron:
> @relativisticelectron I'd need to first understand the goals associated with this PR before we can talk UI options etc. What do you want to search for? When is the global search used / sensible and when the search bar of the tx table?
> 
> So far, I could only gather: "feature to quickly search all wallets for an address." - which is fine if you want to double check whether an address belongs to one of your wallets and you don't remember which one.
> 
> My rough first intuition is, that we talk about a two step approach here. (1) The global search tells you in which wallet(s) to look for, (2) and then, say, the search in the tx table takes over

The original inspiration is gmail. The search searches everything, is fast and changes the experience/behavior of "I have to remember in which folder the email is" to "just search for it".   **UX: Search (and see results immediately) -->  Click on result --> See result fully.  --> Go back to result list if it was the wrong one**
Similar fantastic search is in VS Code for files and settings. 

In Specter I had the experience: In which wallet was the address and was it used already?  I would like to mirror the gmail UX flow with as few steps as possible (I think the current PR doesn't do that, because there are too many clicks, but it was a first attempt).
And if we do it for addresses, there is no reason to not also search transactions ids, labels, amounts, xpubs, ...

relativisticelectron:
Interesting.  Currently the "History" and "UTXO"   however doesn't show unused addresses belonging to a wallet. And adding it seems kind of wrong, because then it is not the   "History" any more.

I think most users are used to global searches not attached to a certain table but in the upper left/right, e.g.
image

relativisticelectron:
I have to redesign this search. Currently getting to the result needs way too many clicks.... Closing this PR for now.

If anyone knows a good webcomponent for a rich-text dropdown, then please write it here.

PR #1797, opened 2022-07-12
Bugfix: support extracting xpub of master xkey derivation

hodlwave:
# Description

I tried importing an extended key from my hardware wallet (via hwi) with derivation = 'm'; this resulted in an exception due to a python3 `List index out of range` error. The cause of the error is clear from the code, which tries to access the key path after splitting the derivation on "m/". For this particular edge case, there is no index=1 so the process fails.

### DIFF
--- a/src/cryptoadvance/specter/hwi_rpc.py
+++ b/src/cryptoadvance/specter/hwi_rpc.py
@@ -262,6 +262,8 @@ class HWIBridge(JSONRPC):
                     derivation, default="xpub", network=network
                 )
                 xpub = convert_xpub_prefix(xpub, slip132_prefix)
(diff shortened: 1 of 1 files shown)


hodlwave:
Is the error from the `cypress_test` suite related to this PR? It looks like it might be a transient connection issue. If so, is it possible to trigger a re-run?

stepansnigirev:
`m` is a valid derivation path. Ack.

hodlwave:
@k9ert got it, I do not have the "Re-run checks" button so must not be in that permissions group. Thanks for the ACK @stepansnigirev .

moneymanolis (commented):
OK!

PR #1798, opened 2022-07-12
Add nix shell

hodlwave:
# Description

This PR makes it easier to develop and run specter desktop for nix users. Install the dependencies by running `nix-shell` from the project's root directory and then run the server with `python3 -m cryptoadvance.specter server --config DevelopmentConfig --debug`.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -7,6 +7,7 @@
     - [Install dependencies:](#install-dependencies)
       - [Ubuntu/Debian:](#ubuntudebian)
       - [macOS:](#macos)
+      - [Nix:](#nix)
       - [Arch:](#arch)
       - [Fedora/CentOS:](#fedoracentos)
       - [Windows](#windows)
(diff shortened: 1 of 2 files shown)


k9ert:
Great, thanks for this PR! Not using Nix myself but wanted to test it already for years, quite different than any other distro, though. Love to see better support there.
Can't we put the `shell.nix` file in the utils-subdirectory?

hodlwave:
Sure thing. I'm actually pretty new to Nix myself. Definitely check it out when you have some time; it's pretty awesome.

I moved `shell.nix` to the utils directory. Not sure if this is best practice or not as I typically see `shell.nix` at the root, but probably not a big deal either way.

k9ert:
> Not sure if this is best practice or not as I typically see shell.nix at the root, but probably not a big deal either way.

There is no best practice regarding that even if some projects claim that. There is so much stuff in the root of a sophisticated project. All are claiming that they are important want to be placed in the root. Some of the less important stuff in our case:
* `pytest.init` thinks that pytesting is so important to claim space at the root
* `Dockerfile` thinks that docker is so important to claim space at the root
* `.cirrus,yml` thinks that cirrus is so important to claim space at the root
* `cypress.json` thinks that cypress is so important to claim space at the root
* Packagemanagement thinks that it's so important that it claims almost 8 files all in the root-folder (they should have all agreed on a subfolder for that stuff ages ago)

It's a disease! Happy for anything which we can remove from the root-folder.

Anyway, thank you very much for your contribution !

hodlwave:
That's totally reasonable; thanks for reviewing :)

PR #1800, opened 2022-07-13
Chore: reasonable error-messages for cypress within cirrus

k9ert:
The current situation is a sroll-orgy. This should make the situation more sane similiar to what it looks like if we have errors in the pytests.


k9ert:
already merged with some other PR

PR #1804, opened 2022-07-14
UI: Optimize tx-table for mobile screen

relativisticelectron:
**Before**:
before

**After**:
after

While not 100% perfect. It is now usable also on a very narrow screen.

TODOs:

- [x] Rebase on  PR https://github.com/cryptoadvance/specter-desktop/pull/1813


moneymanolis:
@relativisticelectron could you perhaps rebase this PR on this PR https://github.com/cryptoadvance/specter-desktop/pull/1813 (once it is merged) - I've done some work there on the tx table which should also make it a bit more mobile friendly.

relativisticelectron:
> @relativisticelectron could you perhaps rebase this PR on this PR #1813 (once it is merged) - I've done some work there on the tx table which should also make it a bit more mobile friendly.

Sure.

relativisticelectron:
> @relativisticelectron could you perhaps rebase this PR on this PR #1813 (once it is merged)

Done.  Now this PR is much smaller.

relativisticelectron:
@moneymanolis : Could you review this PR?  It is now really small.

moneymanolis:
> @moneymanolis : Could you review this PR? It is now really small.

Will do!

src/cryptoadvance/specter/templates/includes/tx-table.html
```diff
@@ -200,6 +200,7 @@
             width: 100%;
             margin-bottom: 10px;
             border: 1.5px solid var(--cmap-border);
+            overflow-x: scroll;
```
b30wulffz, line 203: I think overflow-x: auto would be better choice, considering it handles both overflowing and non overflowing cases. Just to avoid a scroll bar in case of no-overflow.

b30wulffz (changes requested):
Left a comment. Check it once.

Also, can this overflowing content in table header be fixed? 
For example, the buttons here
image

src/cryptoadvance/specter/templates/includes/tx-table.html
relativisticelectron: Done

relativisticelectron:
> Left a comment. Check it once.
> 
> Also, can this overflowing content in table header be fixed? 

Done
image

b30wulffz (approved):

src/cryptoadvance/specter/templates/includes/tx-table.html
moneymanolis: Sorry for responding so late: I would have chosen a different approach. I would have added the class `optional` to the time column, similarly to what we have done with the txid column (or to another column that seems not so necessary for limited space on a phone). We ca finetune that later in a different PR if you like the idea.

src/cryptoadvance/specter/templates/includes/tx-row.html
```diff
@@ -353,6 +353,7 @@
                         locked: this.tx.locked,
                     } });
                     this.dispatchEvent(event);
+                    e.stopPropagation(); // this prevent the onclick of outer divs to be triggered
```
moneymanolis, line 356: @relativisticelectron could you explain this? Which onclick listeneres picked that up? Where did that bubble to? I thought it would be contained within the web component.

src/cryptoadvance/specter/templates/includes/tx-table.html
moneymanolis: @relativisticelectron Would look like this, let me know what you think, I can quickly make a PR out of it:

grafik

src/cryptoadvance/specter/templates/includes/tx-row.html
relativisticelectron: The entire row becomes clickable https://github.com/cryptoadvance/specter-desktop/blob/bb45ac6347b5fd96817fc8eb86d783f2437e1151/src/cryptoadvance/specter/templates/includes/tx-row.html#L206

src/cryptoadvance/specter/templates/includes/tx-table.html
relativisticelectron: That would have worked too :-)

PR #1813, opened 2022-07-18
UI: Complete overhaul of the tooltips used in Specter Desktop

moneymanolis:
### Contents
- All tooltips are now using the web component `tool-tip.html`
- New look / style 💇‍♂️
- Richer API of this web component, there is also the possibility of responsive widths of the tooltip boxes via JS: 

https://user-images.githubusercontent.com/70536101/179552130-9f0152a9-570a-4237-aa89-e932fc3dc84d.mov

- Via the image attribute other images (currently a note image is available) can be used if the light bulb appears to be too optional to read. For example, I've used it in those places, where the info is important and should be read:
image
image

## TODO
- [x] Finalise the documentation of the web component
- [x] Deleting the old css styles of the tooltips?


src/cryptoadvance/specter/templates/includes/tool-tip.html
```diff
@@ -1,110 +1,181 @@
...
+        connectedCallback() {
+            if (this.getAttribute("width")) {
+                this.tooltipBox.style.width = this.getAttribute("width");
+            }
+            if (this.getAttribute("move-right")) {
+                this.tooltipBox.style.left = "60px";
+            }
+            if (this.getAttribute("image")) {
+                if (this.getAttribute("image") == "note") {
+                    this.tooltipImg.src = "{{ url_for('static', filename='img/note.svg') }}";
```
b30wulffz, line 174: shouldn't we allow user to supply image url, instead of a flag for just 1 image?

b30wulffz (commented):
One minor doubt, else looks pretty good.

src/cryptoadvance/specter/templates/includes/tool-tip.html
moneymanolis: We could, but there needs to be a matching image in the static folder. My idea was to enable more pictures via this if construct over time. Might also be the case that we don't need more than the two.

b30wulffz: Ahh, got it

b30wulffz (approved):

PR #1820, opened 2022-07-23
add faucet and exfund extensions

stepansnigirev:
I wrote two extensions that might be useful for people. This PR adds them to the default Specter distribution.

## Faucet extension

It helps to control your regtest node - generate blocks and fund wallets directly from Specter, without going to command line. Useful for development. As it's a development extension it is available with `--config DevelopmentConfig` flag.

[Repository](https://github.com/stepansnigirev/specterext-faucet)

## Exfund extension

Helps to do transactions with large number of outputs - you can select a CSV file with addresses and amounts, extension will parse it and display outputs in a convenient table form. Then you can tweak the fee rate, select wallet to use etc, and then get to a normal signing flow of Specter-Desktop. Amounts can be changed on all inputs at the same time as well - so if funding multiple addresses with the same amount you don't need to have it in the CSV. Also supports liquid.

[Repository](https://github.com/stepansnigirev/specterext-exfund)


relativisticelectron:
The `Exfund extension` looks like a great replacement for the current txt based input (which lacks labels):
image

k9ert:
It would work on `v1.10.5` but also would break if we would merge #1782 . Needs some rework:
https://github.com/cryptoadvance/specter-desktop/pull/1782#discussion_r929812635

stepansnigirev:
I will migrate the extension to json interface of the PSBT creator, it makes more sense than hacking into ui. I fixed json interface for liquid in https://github.com/cryptoadvance/specter-desktop/pull/1831

moneymanolis:
### Some UX suggestions for the Exfund extension

1. Make those buttons smaller (via max-width etc.)
grafik

2. More width for the tooltip (just adding width=XXX)
grafik

3. Rather use sticky and the overall scrollbar than a scrollable table for this table:
grafik

See the discussion and end result [in this PR](https://github.com/cryptoadvance/specter-desktop/pull/1746)

4. Perhaps a toggle for the fee options?
grafik

moneymanolis:
Parsing of the recipients / destinations from CSV works, but I get this error when trying to continue after that:

grafik

moneymanolis:
> The `Exfund extension` looks like a great replacement for the current txt based input (which lacks labels): image
> 
> I doubt there are users, which would prefer the current text based input over a csv based input.

I agree but before removing the text based stuff in the UI we have to ensure that the exfund extension works solidly.

stepansnigirev:
> Parsing of the recipients / destinations from CSV works, but I get this error when trying to continue after that:
> 
> grafik

Can you provide test data you used to get this error? Then I could reproduce it.

moneymanolis:
> Can you provide test data you used to get this error? Then I could reproduce it.

I used your test data: https://github.com/stepansnigirev/specterext-exfund#example-csv

moneymanolis:
If I use `get_by_alias()` to get the wallet instance it works!

grafik

moneymanolis:
### UX suggestions for the Faucet extension

- _"Get some funds from your default wallet"_ - I think most people (even devs) will not understand that the default wallet in Core ("") is meant here. Either add / explain this or just skip the default wallet part. I thought first there that some sort of default wallet in Specter was set through the extension.
- `<input type="number" value="0.1" name="amount" id="amount">` the value attribute seems to only allow (because of some browser side validation) amounts with 0.1 as decimal, so only 15.1 not 15:
grafik

   Just replace value with a placeholder?

- Why show _"Send back to bcrt1qsrts3q6xwae9whxx8sn957d6pvpj2924uxt5ju"_? Doesn't really matter to which address of the default wallet the `generatetoaddress` is transferring to.

### General questions
- How is the Core default wallet getting funds? Clicking 100 times generate block doesn't seem great.

stepansnigirev:
I addressed all your comments and bumped pip versions.

> Why show _"Send back to bcrt1qsrts3q6xwae9whxx8sn957d6pvpj2924uxt5ju"_? Doesn't really matter to which address of the default wallet the `generatetoaddress` is transferring to.

It's sometimes useful to get a random destination address to test send functionality, so I decided to add this line with a fresh address of the default wallet, so people can grab it and send money to it while testing.

moneymanolis:
> It's sometimes useful to get a random destination address to test send functionality, so I decided to add this line with a fresh address of the default wallet, so people can grab it and send money to it while testing.

Ah okay, I can now see the benefit of that!

moneymanolis:
@stepansnigirev I've tested the exfund plugin again. Looks nice. I think we should add the `@login_required `decorator to the index route otherwise you can run into this error (if you have authentication setup up and enter the URL directly, e.g. if you have it bookmarked etc.):
`AttributeError: 'AnonymousUserMixin' object has no attribute 'wallet_manager'`
The alternative would be to replace the flask User object `current_user` with `user`from the user_manager but then there would be no check whether the user is logged in which is currently not a problem for the plugin but could be an issue later down the line if there are further changes to the plugin.

moneymanolis:
@relativisticelectron I made a PR in the extension repo, in case you want to check it out and test: https://github.com/stepansnigirev/specterext-exfund/pull/4

Issue #1824, opened 2022-07-28
UI: Firefox: No mouse text selection for label edits


relativisticelectron:
Editing labels only works with the keyboard. There is no mouse-text selection possible (in firefox)
![Peek 2022-07-28 20-51](https://user-images.githubusercontent.com/60378539/181615692-7d32e13f-796d-4957-bef5-e1d6cccc099f.gif)
![Peek 2022-07-28 20-53](https://user-images.githubusercontent.com/60378539/181615700-cc4852d8-046e-4720-90a6-6fb0f06d1ef8.gif)

The problem seems to be that the link and `contenteditable` do not go well together. Disabeling the link solves the problem.
![Peek 2022-07-28 21-08](https://user-images.githubusercontent.com/60378539/181618267-60128f63-bae5-42fa-bcae-99621acbf3f9.gif)

PR #1825, opened 2022-07-28
UI BugFix: Firefox:  Allow mouse selection during  address label editing

relativisticelectron:
Fixes: https://github.com/cryptoadvance/specter-desktop/issues/1824

It now works also in firefox:


moneymanolis:
Mhh, works with and without the href attribute for me. I also don't understand the connection between removing this attribute and the mouse selection. I'll let some of the SoB participants test this as well.

moneymanolis:
This would also break this [function](https://github.com/cryptoadvance/specter-desktop/blob/22042f4b12241db3ab0eeeaf9e5f791d08a68ca0/src/cryptoadvance/specter/templates/includes/address-label.html#L215) - even though I am not really sure why this was implemented like this. As of now at least, clicking on the address label serves as link to a explorer resource if an explorer (such as blockstream.info) is selected:

https://user-images.githubusercontent.com/70536101/183246977-e96edc9d-438a-4cc7-9a1a-b5a4cb6ee5f9.mov

relativisticelectron:
The href link is reset every time at updating: https://github.com/cryptoadvance/specter-desktop/blob/972415a762c91528b172068707dac664419bf20f/src/cryptoadvance/specter/templates/includes/address-label.html#L219

The PR only disabled the link during editing.

moneymanolis:
You're right, looks like the href attribute is "recovered" by setting its value in line 219. 

I am still a bit hesitant to merge this since it reduces the incentive to make the edit labelling in a proper way: either with a pop-up or at least using an input field where you input the new label. The span element is not intended for this purpose, that it is why all of this stuff pops up. There is also the issue that the caret in Firefox is just not appearing (only if you click somewhere else and then on the label).

My suggestion, let's give it a week or so and if no PR appears to tackle this until then, I'll merge this PR for the meantime. 
I am tagging @kdmukai for his information.

relativisticelectron:
> I am still a bit hesitant to merge this since it reduces the incentive to make the edit labelling in a proper way

Interesting thought.  I agree that an `input` field (or probably test-area for long labels) would be better.

> My suggestion, let's give it a week or so and if no PR appears to tackle this until then, I'll merge this PR for the meantime.

Ok.

moneymanolis:
I'll merge that for now as a quick improvement for Firefox users and since no PR has showed up.

Issue #1826, opened 2022-07-29
UI Bug Mobile: Some elements are not viewable/accessible  , due to height issues


relativisticelectron:
On mobile (not the browser emulated mobile screen, but a real mobile phone, firefox), it shows that the last entry of the tx-table is not viewable. 
![Screenshot_20220728_220411_org mozilla firefox](https://user-images.githubusercontent.com/60378539/181682704-995e4ff4-d185-4d10-a0a2-33f0f329e670.jpg)

This is however not a problem due the tx-table, but occurs on every page. It is explained here: https://css-tricks.com/the-trick-to-viewport-units-on-mobile/
On mobile I get `window.innerHeight = 600, window.outerHeight = 656`

moneymanolis:
This currently creates huge additional space at the bottom of the tx table for me:

<img width="1066" alt="grafik" src="https://user-images.githubusercontent.com/70536101/183256525-76f60690-9cae-4552-bc91-e2db14ff1e23.png">

I think it would be good to use media queries to restrict the extra spacing to mobile devices, have a look here for example: 
https://wpbuffs.com/media-query-for-mobile/

I could not reproduce the issue in the first place using Firefox but this depends on so much stuff besides the browser.

relativisticelectron:
Fixed by https://github.com/cryptoadvance/specter-desktop/pull/1827

PR #1827, opened 2022-07-29
UI Bugfix Mobile: Add distance at the bottom, if there is a difference between `window.innerHeight` and `window.outerHeight` 

relativisticelectron:
Fix for https://github.com/cryptoadvance/specter-desktop/issues/1826


moneymanolis:
This currently creates huge additional space at the bottom of the tx table for me:

grafik

I think it would be good to use media queries to restrict the extra spacing to mobile devices, have a look here for example: 
https://wpbuffs.com/media-query-for-mobile/

I could not reproduce the issue in the first place using Firefox but this depends on so much stuff besides the browser.

src/cryptoadvance/specter/static/styles.css
```diff
@@ -1244,3 +1244,16 @@ input:checked + .slider:before {
...
     transform: translateX(26px);
 }
+
+
+
+/* The viewable height is often smaller than the height given by the browser. 
+See: https://css-tricks.com/the-trick-to-viewport-units-on-mobile/
+This style is used for a div at the bottom in the <main> element to create a distance at the bottom of the pages
+*/
+.mobile-distance-element-botom {
```
moneymanolis, line 1254: Typo, should be bot**t**om

src/cryptoadvance/specter/templates/device/device.jinja
```diff
@@ -138,33 +138,34 @@
...
-			<button type="submit" style="background: none; color: inherit; border: none; padding: 0; font: inherit; cursor: pointer; outline: inherit;">
-				<div class="small-card radio">
-					<img src="{{ url_for('static', filename='img/key.svg') }}" class="svg-white" style="width: 120px;">
-					{{ _("Create single key wallet") }}
-				</div>
-			</button>
-		</form>
-		{% if specter.device_manager.devices | length > 1 %}
-			<a href="{{ url_for('wallets_endpoint.new_wallet', wallet_type='multisig') }}" style="text-decoration: none;">
+		<h1>{{ _("New device was added successfully!") }}</h1>
```
moneymanolis, line 141: Looks like GitHub is showing a weird diff here, this code is already on master as far as I can see ...

src/cryptoadvance/specter/static/styles.css
```diff
@@ -1244,3 +1244,16 @@ input:checked + .slider:before {
...
+    height: mobileDistanceElementBotomHeight; /* Fallback for browsers that do not support Custom Properties */
```
moneymanolis, line 1255: How is this a fallback? What would such a browser do with "mobileDistanceElementBotomHeight"?

relativisticelectron:
> This currently creates huge additional space at the bottom of the tx table for me:

Strange. Ok. Then this difference calculation is not a good idea.

> 
> I think it would be good to use media queries to restrict the extra spacing to mobile devices, have a look here for more on that: https://wpbuffs.com/media-query-for-mobile/
> 
> Sth. like this:
> 
> ```
> 
> @media only screen and (max-width: 600px) {
> .mobile-distance-element-botom {
>     height: mobileDistanceElementBotomHeight; /* Fallback for browsers that do not support Custom Properties */
>     height: var(--mobileDistanceElementBotomHeight, mobileDistanceElementBotomHeight);
>     flex-shrink: 0;
>     width: 100%;
>   }
> }
> ```
> 
> BTW: I could not reproduce the issue in the first place using Firefox but this depends on so much stuff besides the browser. Update: Could reproduce it with Safari.

Ok, I will try the `@media`.

moneymanolis:
Also: Did you try this approach as well? 
https://css-tricks.com/the-trick-to-viewport-units-on-mobile/
Would be a bit more tailored towards the individual viewports than the difference between `innerHeight` and `outerHeight`.

Does not work for Safari for me, though, and Firefox works fine without anything for me, so can't verify whether this helps as well.

src/cryptoadvance/specter/templates/device/device.jinja
relativisticelectron: There was a `/div`  missing on master, and the inner codeblock was not indented.

relativisticelectron:
> Also: Did you try this approach as well? https://css-tricks.com/the-trick-to-viewport-units-on-mobile/ Would be a bit more tailored towards the individual viewports than the difference between `innerHeight` and `outerHeight`.
> 
> Does not work for Safari for me, though, and Firefox works fine without anything for me, so can't verify whether this helps as well.

This is where I have the idea from to use the `innerHeight`, and then discovered that `outerHeight - innerHeight` fits the height of the navigation bar.  Their JS code is not directly applicable, because we have nothing to shrink. On our  endpoints are various elements that get cut off by the navigation bar.

relativisticelectron:
The `window.outerHeight` was the wrong [approach](https://developer.mozilla.org/en-US/docs/Web/API/Window/outerHeight).   
I now changed it to a fixed height if it is a narrow screen.

src/cryptoadvance/specter/static/styles.css
```diff
@@ -1249,6 +1264,10 @@ input:checked + .slider:before {
     -ms-transform: translateX(26px);
     transform: translateX(26px);
 }
+
```
moneymanolis, line 1267: Why those extra lines? Remove?

src/cryptoadvance/specter/static/styles.css
```diff
@@ -1019,6 +1019,18 @@ nav.collapse-on-mobile a.mobile-nav-icon {
     display: none;
 }
 
+
+/* The viewable height is often smaller than the height given by the browser. 
+See: https://css-tricks.com/the-trick-to-viewport-units-on-mobile/
+This style is used for a div at the bottom in the <main> element to create a distance at the bottom of the pages
+*/
+.mobile-distance-element-botom {
```
moneymanolis, line 1027: Please fix the spelling of bottom

src/cryptoadvance/specter/static/styles.css
```diff
@@ -1192,6 +1204,9 @@ nav.collapse-on-mobile a.mobile-nav-icon {
     .main-page-container{
         background-image: none;
       }
+    .mobile-distance-element-botom {
+        height: var(--mobileDistanceElementBotomHeight, mobileDistanceElementBotomHeight);
```
moneymanolis, line 1208: Doesn't make sense to repeat the variable here, better to provide some px value as fallback.

grafik

src/cryptoadvance/specter/static/styles.css
relativisticelectron: done

src/cryptoadvance/specter/static/styles.css
relativisticelectron: Sorry, I didn't even see the typo until now!   Done.

src/cryptoadvance/specter/static/styles.css
relativisticelectron: Thanks for the reminder. Done.

src/cryptoadvance/specter/static/styles.css
relativisticelectron: I removed it.

src/cryptoadvance/specter/static/styles.css
relativisticelectron: Done. (now)

src/cryptoadvance/specter/templates/base.jinja
```diff
@@ -40,6 +40,9 @@
 			}
 		</style>
 		<script type="text/javascript" src="{{ url_for('static', filename='helpers.js') }}"></script>
+		<script>
```
moneymanolis, line 43: Can we move this to `helpers.js`?

relativisticelectron: Yes, that fits well in there. Done.

moneymanolis: Great, thanks.

moneymanolis (approved):

Issue #1835, opened 2022-08-11
Bug: No Price data in address table


relativisticelectron:
**Describe the bug**
The price (in $ or so) is not shown in the address table in amount.

Cause of the bug in https://github.com/cryptoadvance/specter-desktop/blob/a58afbaeed3335fbdecb145c9db48ed24e379abc/src/cryptoadvance/specter/templates/includes/address-row.html#L139

This statement differs from the tx-table (where the price is shown) https://github.com/cryptoadvance/specter-desktop/blob/a58afbaeed3335fbdecb145c9db48ed24e379abc/src/cryptoadvance/specter/templates/includes/tx-row.html#L243

Is perhaps the `!` in  `!(this.assetsCount == 1 && Object.keys(this.addressData.assets)[0] == "{{ specter.default_asset }}")` wrong?

**Reproduce**
regtest, bitcoin, linux
In my setup holds: `this.assetsCount == 0` , which causes the hidden amounts.

k9ert:
Should be fixed in [v1.12.0](https://github.com/cryptoadvance/specter-desktop/releases/tag/v1.12.0). Please reopen if necessary.

PR #1836, opened 2022-08-11
Bugfix: price in address table was not visible

relativisticelectron:
See https://github.com/cryptoadvance/specter-desktop/issues/1835


moneymanolis:
Just to clarify, the prices for LBTC weren't show and this is what you wanted to fix? If so, that's working. I just changed the variable name to a more precise one. Hope that is OK.Thanks as always

moneymanolis (approved):

relativisticelectron:
There was no $ value in the address table (for bitcoin), see https://github.com/cryptoadvance/specter-desktop/issues/1835#issue-1335811738 .

I determined it is this liquid part of the if statement to prevent any prices to be visible (even though I work with bitcoin regtest).

moneymanolis:
Ah OK, now I get it, it was for BTC, too. Was so focused on the Liquid part of things. Great catch. Merging now.

PR #1837, opened 2022-08-11
Add BIP-371 Taproot psbt fields (bip32 derivations for inputs & outputs)

kdmukai:
Requires: https://github.com/diybitcoinhardware/embit/pull/33

This update brings Specter Desktop psbts into compliance with BIP-371's `PSBT_IN_TAP_BIP32_DERIVATION` and `PSBT_OUT_TAP_BIP32_DERIVATION` fields. This replaces the workaround where the bip32 derivations were inserted into the non-Taproot psbt fields (`PSBT_IN_BIP32_DERIVATION`, `PSBT_OUT_BIP32_DERIVATION`).

* `Wallet.fill_psbt` now injects BIP-371 Taproot psbt fields into the psbt if they aren't provided by Bitcoin Core (won't be available from Core until v24.0?).
* Enables `taproot_derivations` by default in `Wallet.fill_psbt`.
* Per https://github.com/diybitcoinhardware/embit/pull/33, only root-level Taproot keyspends are supported.

Misc:
* Adds "tr" to correctly filter "Devices" list to include Taproot xpubs when creating a new single sig wallet.


moneymanolis:
Just a clarification for readers regarding the "misc change":
If you had a device which ONLY had a taproot xpub, it was not shown in the list of possible devices to setup a wallet with. Now it will be shown as well.

stepansnigirev:
Fixed in embit, use v0.6.1. Bug on the library side :(

moneymanolis:
I've bumped embit and added tests in https://github.com/cryptoadvance/specter-desktop/pull/1837/commits/d70381bfdc4cd2446ad7bf16e1a1a9dd350111a6, but looking at a the psbt in the test, I can't really understand / see taproot specific fields. "bip32_derivs" field is gone, but that's about it. What am I missing here?

moneymanolis (changes requested):
The `to_dict` method as described [here](https://github.com/cryptoadvance/specter-desktop/pull/1837#issuecomment-1289484424) has to be adopted to write the pytest for a taproot wallet.

moneymanolis (approved):

src/cryptoadvance/specter/wallet.py
```diff
@@ -1866,25 +1872,37 @@ def fill_psbt(
         b64psbt,
         non_witness: bool = True,
         xpubs: bool = True,
-        taproot_derivations: bool = False,
+        taproot_derivations: bool = True,
```
moneymanolis, line 1875: Does this change of the default make sense? 
It would only add the taproot PSBT fields if it is a taproot wallet. 
But, if a user adds taproot keys by accident / uses them very rarely he would carry those fields with him in the QR codes all the time even if his signing device (devices can also be switched) can't even process the taproot fields.

src/cryptoadvance/specter/wallet.py
```diff
@@ -1740,14 +1744,16 @@ def createpsbt(
                 True,  # bip32-der
             )
 
-            b64psbt = r["psbt"]
+            # Always explicitly fill psbt with any missing fields
```
moneymanolis, line 1747: This can create unnecessary large QR codes. In general, taproot_derivations should only be set to True if the device can support it.

moneymanolis:
One remaining question: Why not remove the `taproot_derivations` parameter altogether and always fill the the psbt fields with taproot fields if a taproot wallet is used? Even if a user is switching the device to use a non-taproot compatible device for the taproot wallet he is still better of by having the fields instead of bip32 derivs fields.

Issue #1838, opened 2022-08-15
"Sever error" when trying to create new wallet


bluruin17:
When I try to create a wallet that is derived from my hardware device keys I get the following error message:

> "Server error!
> Exception: 'NoneType' object has no attribute 'endswith'

And it gives me the following code:

```
Traceback (most recent call last):
  File "flask/app.py", line 1523, in full_dispatch_request
  File "flask/app.py", line 1509, in dispatch_request
  File "flask_login/utils.py", line 269, in decorated_view
  File "cryptoadvance/specter/server_endpoints/wallets.py", line 316, in new_wallet
AttributeError: 'NoneType' object has no attribute 'endswith'
```

Everything is running locally on Manjaro. I'm using Specter 1.10.5

relativisticelectron:
https://github.com/cryptoadvance/specter-desktop/blob/a58afbaeed3335fbdecb145c9db48ed24e379abc/src/cryptoadvance/specter/server_endpoints/wallets.py#L316

It seems that the fix https://github.com/cryptoadvance/specter-desktop/pull/1771/files of https://github.com/cryptoadvance/specter-desktop/issues/1756  was insufficient to cover also this bug.

relativisticelectron:
Opened RP: https://github.com/cryptoadvance/specter-desktop/pull/1839

k9ert:
Should be fixed in [v1.12.0](https://github.com/cryptoadvance/specter-desktop/releases/tag/v1.12.0). Please reopen if necessary.

PR #1839, opened 2022-08-15
Bugfix: explorer not set

relativisticelectron:
See: https://github.com/cryptoadvance/specter-desktop/issues/1838


k9ert:
Good point!

Issue #1627, opened 2022-03-08
UIUX: Improvement of balance display


moneymanolis:
Once #1580 is merged:

![image](https://user-images.githubusercontent.com/60378539/157185870-e995e0d9-10c3-4197-837c-4a09a5af081b.png)
It would be even nicer if the amounts would be aligned such that the decimal separator is at the same x position.   And for completeness the unit  (tBTC) after each amount would be good too.

_Originally posted by @relativisticelectron in https://github.com/cryptoadvance/specter-desktop/pull/1580#discussion_r821382127_

moneymanolis:
Quoting relevant comments from @cypherhoodlum and his PR (already a lot of CSS etc. groundwork in there) here as well:

https://github.com/cryptoadvance/specter-desktop/pull/1580#issuecomment-1075713089
https://github.com/cryptoadvance/specter-desktop/pull/1580#issuecomment-1075749817
https://github.com/moneymanolis/specter-desktop/pull/1

k9ert:
The last two boxes could be ticked by @cypherhoodlum PR here: https://github.com/moneymanolis/specter-desktop/pull/1

moneymanolis:
Yes, the main CSS approach is already in there.
But: x position is not fixed by it (only looks like it since the amounts have the same length)
And also, I am not completely sure whether the approach is the best regarding readability. Since this is a very important element, I think it is worth giving it a bit more thought. 
Tagging @kdmukai here as well, the master of seed signer balance displaying.

cypherhoodlum:
It seems this issue requires the #1580 to be merged into the master branch first:
<img src="https://user-images.githubusercontent.com/82510823/160277064-f4fd25d2-4566-43e3-9f4c-dfe04f1cd812.png" width="600" />
Or is there something I'm missing @k9ert @moneymanolis?

moneymanolis:
@cypherhoodlum #1580 is close to be merged. Finally, the CI worked again.

cypherhoodlum:
Good to hear, seemed like some nasty problems. I will finish styling the balance display once #1580 is fully merged.

moneymanolis:
fyi @cypherhoodlum: #1580 has been merged

cypherhoodlum:
Sorry, it's been a while. I'm starting to think the fixed x position based on the decimal point is too ambitious given all that's happening in the balances display. Here's how my proposal looks right now:

[...]

PR #1841, opened 2022-08-16
UI: Better Balance display

relativisticelectron:
See https://github.com/cryptoadvance/specter-desktop/issues/1627

A unified display for wallets and summed total balances:

- number alignment at the decimal point
- code has flexible truncating, currently with `maximum_digits_to_strip=7` and `minimum_digits_to_strip=6`
- Better satoshi counting enabled by default with `enable_digit_spaces_for_counting=True`

**Before:**
image

**Now:**
image
image
image

**Liquid:**
Screenshot_20220817_080754
image

Todo:
- [x]  $ amount
- [x] number alignment for liquid assets
- [x] make Satoshis counting better
- [x] refactor code for better readability


k9ert:
LGTM. I've added a small example page in the devhelper for the macro in order to get a better understanding of how the macro is working. Requested @moneymanolis for review as i don't want to merge UI specific stuff anymore.

relativisticelectron:
> LGTM. I've added a small example page in the devhelper for the macro in order to get a better understanding of how the macro is working. Requested @moneymanolis for review as i don't want to merge UI specific stuff anymore.

Thanks for the better documentation. I also added/copied a better argument description directly at the macro documentation.  

TBH: I do not get the concept of the dev extension. I have never used it and it is very counterintuitive to me. For me there are 2 places to put dev help:

1. The doc strings, or comments directly in the code (anyone looking at the code will see it)
2. The doc folder for something more general/ more extensive

k9ert:
> TBH: I do not get the concept of the dev extension. It is very counterintuitive to me. 

I can't claim that i have a fixed best practise definition of what the dev-extension is or should be. It started when i realized in some project where i used the [SB Admin template](https://startbootstrap.com/template/sb-admin), that it would be helpful to have the demo-pages available in my application right away. I could compare very easily my production-code with the demo-template-code within the same application. As i had a notion of a user and users could be "admin", i made it available to admin-users.

Also i think that (frontend-) development should be better supported and tested. As running code is more important than comprehensive documentation (see [agile manifesto](https://agilemanifesto.org/)) i thought that if we try to split up the frontend in smalller pieces/components/macros, it would be good to have a place where you can not only read about it in theory, but see it in action AND in isolation.

Anyway, the python-console is a great idea and i would suggest to put it into the dev-extension first. In that case it would only work in dev but than later on, we could potentially create a dedicated extension which one could use in prod. However i'm not yet sure whether it's a good idea to ship that by default (in prod).

relativisticelectron:
> image

@moneymanolis : I somehow fail in css to put the light-bulb right of the "Immature". I think even underneath is acceptable, but if you know a way, that would be great.

moneymanolis:
This PR is on my list for tomorrow, will check it out.

src/cryptoadvance/specter/util/common.py
```diff
@@ -29,6 +29,53 @@ def snake_case2camelcase(word):
...
+        break
+
+    if count_digits_that_can_be_stripped >= minimum_digits_to_strip:
+        for i in reversed(range(len(formatted_amount))):
+            if (
+                formatted_amount[i] == "0"
+                and len(formatted_amount) - i <= maximum_digits_to_strip
+            ):
+                # replace with https://unicode-table.com/en/2007/
+                formatted_amount = replace_substring(formatted_amount, i, 1, " ")
```
moneymanolis, line 63: Why not use `\u2007` in
grafik ?
Thus:` replace_substring(formatted_amount, i, 1, "\u2007")`
As you can see, GitHub's front end is already not even showing the figure space. Same for below with `\u2008`.

moneymanolis: Also: Did you have specific reasons to choose those unicode space types? I just found out that there are 20 different types.

relativisticelectron: Yes, I chose the spaces such, that they are equivalent in width to a number https://unicode-table.com/en/2007/  https://en.wikipedia.org/wiki/Figure_space and equivalent to a Punctuation https://unicode-table.com/en/2008/. 

> Why not use `\u2007` in

I'll try that.

moneymanolis: Cool. As for Cypress, just ignore it for now, it's just flakiness.

src/cryptoadvance/specter/util/common.py
```diff
@@ -29,6 +29,53 @@ def snake_case2camelcase(word):
     return "".join(x.capitalize() or "_" for x in word.split("_"))
 
 
+def replace_substring(text, start_position, replace_length, new_str):
+    return text[:start_position] + new_str + text[start_position + replace_length :]
+
+
+def satamount_formatted(value):
```
moneymanolis, line 36: Wouldn't it make sense to rename this to sth. more generic, like `convert_to_sats`or simply `satamount` and add a check whether the value is a float or a string and then return `"{:,.0f}".format(round(float(value) * 1e8))` (if value is a string) or  `"{:,.0f}".format(round(value * 1e8))` (if value is a float). We could then also reuse this function in other filters like `btcunitamount`.

moneymanolis:
I am just throwing in the Seed Signer display to discuss whether we want to add any colouring and whether we want to strip starting zeroes (as of now there is only the possibility of stripping trailing zeroes). I like the approach in this PR so far with  using two different spacings to easily identify the sats amount (the spacings are basically what Seed Signer is achieving with the different colours). For example, one can easily identify 140k sats in the unconfirmed row below. Would it add anything to color the leading 0s in this case? Any other use cases for colouring for us when sticking to the spacings approach?
I am also tagging @kdmukai. 

grafik

grafik

relativisticelectron:
Coloring, I think, is a good idea.  Removing the left 0's, is nice to look at, but mathematically wrong.

moneymanolis:
> Seedsigner has spacing + coloring

If there is spacing, I can't see it :-)

My suggestion would be to perhaps color in grey if the first two digits are 0s. So those in italic here would be grey: _0, 00_ 154 000 BTC

Could be done with sth. like this:
https://stackoverflow.com/questions/31803300/coloring-the-text-depending-on-numeric-value-using-css

relativisticelectron:
> > Seedsigner has spacing + coloring
> 
> If there is spacing, I can't see it :-)

I visualized here the spacings between blocks (left) and inside a block (right)
image

src/cryptoadvance/specter/util/common.py
relativisticelectron: > return `"{:,.0f}".format(round(float(value) * 1e8))` (if value is a string) or `"{:,.0f}".format(round(value * 1e8))` (if value is a float).

Isn't this already covered with the current implementation?

> Wouldn't it make sense to rename this to sth. more generic, like `convert_to_sats`or simply `satamount`

OK.

moneymanolis: I wouldn't take chances with floats, they are scary, perhaps you are right, I would rather make it more explicit. 
If you want to stress that the function returns a string, rather use type hints. I can change it if you want and push it. I am also writing tests as I go through the review.

relativisticelectron: > I can change it if you want and push it.

great.  I think later we can add also coloring.

kdmukai: Shouldn't use floats at all. The `Decimal` class and its `quantize` method are the only precise way to handle digits below the decimal point.

kdmukai: > Wouldn't it make sense to rename this to sth. more generic, like `convert_to_sats`or simply `satamount` and add a check whether the value is a float or a string and then return `"{:,.0f}".format(round(float(value) * 1e8))` (if value is a string) or `"{:,.0f}".format(round(value * 1e8))` (if value is a float). We could then also reuse this function in other filters like `btcunitamount`.

Either case sounds like a mess. I'm not familiar with how the `value` data is passed around by the time it gets here, but values should internally be represented as integer sats. If some other call further back is changing ints to floats or strings, then instead of having two interpretations (int -> float -> str; int -> str -> str), the prior formatting functions should be altered or avoided so that this is the only conversion to str.

kdmukai:
SeedSigner adds a 2-pixel spacer between each group, but it's only there as the subtlest possible hint to the eye to help differentiate the shades of gray (each shade of gray is not as visually distinct from its neighbor as the screenshot would suggest, which defeats the point of the gradation). On the tiny 1.3" screen it's more a psychological/perceptual trick than anything else.

Definite ACK on aligning the amounts.

But passionate NACK on the "sat comma-as-spaces"-style grouping (my full reasoning [here](https://bitcoinmagazine.com/culture/bitcoin-vs-sats-denomination-why-both)). One thing I didn't note in that article is that my eye always lands on the middle group of numbers due to the visual separation. So I end up having to scan backwards to find the decimal point and then can scan forward again.

I would want an option to decide how those 8 digits below the decimal place are displayed, which would actually open up even more possibilities (sat comma, sat comma-as-spaces, no treatment, btc|sats hybrid, truncate zeros (barf!)).

Does this formatting apply everywhere that an amount is displayed or is it just for this one display? It would seem too inconsistent to me to make such an opinionated change on the main display but not have that applied across the entire UI.

moneymanolis:
@stepansnigirev made the point that we should also think about copy and pasting amounts. For example, in our own send dialogue we'd run into a front end amount warning if we'd use the filter there and copy and pasted:

grafik

Also because of this, how about using two colors, e.g. like so: 

grafik

moneymanolis:
My suggestion would be: We can merge this PR pretty soon and don't change too much but set `enable_digit_spaces_for_counting` to False in the filter. Then we can decide whether it is worth it to do the extra work (and how difficult it is) to use colouring. I don't think it is a great idea to switch between spaces and colours too often (well, depending on fast we release but still). The copy & pasting problem is actually a red flag for me to use the spacing.

src/cryptoadvance/specter/util/common.py
moneymanolis: The original source of the amounts is Bitcoin Core which uses floats.

relativisticelectron:
> set `enable_digit_spaces_for_counting` to False

agreed. Copy&Pasting must be preserved.

> Also because of this, how about using two colors, e.g. like so (just realising that this is the Seed Signer approach above again):
> 
> grafik

I will implement the coloring and then we can evaluate if it should be used.  If not, then I will still leave it as a (by default disabled) feature in the function, to (perhaps) be used later.




> But passionate NACK on the "sat comma-as-spaces"-style grouping (my full reasoning [here](https://bitcoinmagazine.com/culture/bitcoin-vs-sats-denomination-why-both)). One thing I didn't note in that article is that my eye always lands on the middle group of numbers due to the visual separation. So I end up having to scan backwards to find the decimal point and then can scan forward again.

I like your "why-not-both" approach. This could perhaps be explored in a future PR.

relativisticelectron:
image

- the copy&paste problem is now solved (both for stripping 0's and for in betwen the digits)
- digit spacing and colors defined in css

relativisticelectron:
> Nice collaborative iteration!

yes, thanks for the input.



> One other minor note: "Summed" is a slightly odd word choice here. Technically correct, but is maybe more common for British English? "Combined Wallet Balance" is a more natural fit.

done

moneymanolis:
We had a discussion in the dev chat to start avoiding `round(float()) ` and instead use this https://docs.python.org/3/library/decimal.html). I'd rather do this in one go with all the filters, common functions etc. But, if someone one wants to directly start with replacing `round(float()) `  in this PR, go for it.

relativisticelectron:
> I'd rather do this in one go with all the filters, common functions etc.

agreed. Then problems can be solved in a consistent way.

kdmukai:
Submitted a PR to @relativisticelectron's branch. I haven't suggested changes this way before, but it should work that if you merge my PR into your branch, those changes will make it here.

kdmukai:
`satamount_formatted` suggests that `value` could be a string. Is that really possible? Or is it always a float? Ideally we'd start adding type hints to the input args.

If it can be str or float, looks like we'd have to do:
```
from typing import Union
def satamount_formatted(value: Union[float, str]) -> str
```

src/cryptoadvance/specter/templates/wallet/history/components/total_wallet_balances.jinja
```diff
@@ -1,113 +1,166 @@
...
-            <button type="button" class="btn" style="margin: auto;" onclick="showPageOverlay('liquid-assets-registry')">{{ _("Show assets list") }}</button>
-        </small>
-        <br>
-        {% endif %}
     </div>
-    {% if wallet.rescan_progress or specter.utxorescanwallet == wallet.alias %}
+    {% if rescan_progress or specter.utxorescanwallet == alias %}
         <span id="wallet_rescan_data" style="display: contents;">
-            <h2>Rescanning blockchain: <span id="{{ wallet.alias }}_balances_wallet_rescan_percents">{{ "%.2f"|format(specter.info["utxorescan"] if specter.utxorescanwallet == wallet.alias else wallet.rescan_progress * 100) }}</span>%</h2>
+            <h2>Rescanning blockchain: <span id="{{ alias }}_balances_wallet_rescan_percents">{{ "%.2f"|format(specter.info["utxorescan"] if specter.utxorescanwallet == alias else rescan_progress * 100) }}</span>%</h2>
```
moneymanolis, line 163: I don't get this (also in the code beforehand): Shouldn't it be the other way around? Show the lengthier rescan progress if a full rescan has been launched and if not show the progress of the quicker UTXO rescan? Otherwise, it might be confusing to users who think they are already done when looking at the UTXO scan.
So, in pseudo-code: 
if rescan_progress then show rescan_progress * 100 else show specter.info["utxorescan"]

moneymanolis: Or why not just have to separate h2-tags for the complete rescan and the UTXO scan only?

relativisticelectron: good question.

> Or why not just have to separate h2-tags for the complete rescan and the UTXO scan only?

Yeah.... would makes things much clearer.

src/cryptoadvance/specter/util/common.py
```diff
@@ -33,16 +34,16 @@ def replace_substring(text, start_position, replace_length, new_str):
     return text[:start_position] + new_str + text[start_position + replace_length :]
 
 
-def satamount_formatted(value):
+def format_btc_amount_as_sats(value: Union[float, str]) -> str:
```
moneymanolis, line 37: This is even less generic now, we can leave it like this and rename when doing the whole Decimal thing.

moneymanolis: The type hints are great!

src/cryptoadvance/specter/templates/wallet/history/components/total_wallet_balances.jinja
moneymanolis: OK, I'll change that then.

moneymanolis:
The latest version of the PR no longer ensures that we are aligned along the decimal point. For example, when we have at least six 0s (we currently strip the 0s then) the display looks this (see immature, I have the coloring turned on manually):
grafik
I think the reason for this is:

```
    formatted_amount = re.sub(
        r"(\u2007|\u2008)", r'<span class="unselectable">\1</span>', formatted_amount
    )
```

We are wrapping the unicode spaces around HTML here, but HTML ignores all spacing. CSS should come to the rescue. I think we just need to add sth. to the class `unselectable` which creates the same space. 

See also: 
https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Whitespace#html_largely_ignores_whitespace

moneymanolis:
I don't understand why we are not keeping the great colouring (with little spaces as the default). It is STRICLY better than not having it. Readability is improved so much with it. @k9ert what do you think?

moritzwietersheim:
Would also go without coloring, but really like the coloring, makes it very clear. 
Great PR @relativisticelectron - i am such a fan of well formatted UI (and spreadsheets 🤓) ... and @kdmukai has obviously done a ton of obsessing on this over at Seedsigner. So cool to see!

I recently used Ledger Live & Trezor Suite - and it's really sad to use, basically not even SATs displayed...and so on...

relativisticelectron:
> @relativisticelectron The latest version of the PR no longer ensures that we are aligned along the decimal point. For example, when we have at least six 0s (we currently strip the 0s then) the display looks this (see immature, I have the coloring turned on manually): grafik I think the reason for this is:
> 
> ```
>     formatted_amount = re.sub(
>         r"(\u2007|\u2008)", r'<span class="unselectable">\1</span>', formatted_amount
>     )
> ```
> 
> We are wrapping the unicode spaces around HTML here, but HTML ignores all spacing. CSS should come to the rescue. I think we just need to add sth. to the class `unselectable` which creates the same space.
> 
> See also: https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model/Whitespace#html_largely_ignores_whitespace

I cannot reproduce the   shift you see. For me it looks like:
image
and without formatting:
image

tested in Firefox and Chromium in Linux

kdmukai:
While we're talking about alignment, has anyone tried seeing how it looks if the labels (e.g. "Total amount:") are right-justified so the colons all align?

relativisticelectron:
> While we're talking about alignment, has anyone tried seeing how it looks if the labels (e.g. "Total amount:") are right-justified so the colons all align?

image

I don't think this is an improvement.

relativisticelectron:
> I don't understand why we are not keeping the great colouring (with little spaces as) the default. It is STRICLY better than not having it. Readability is improved so much with it.

Fine by me. We should in a followup PR also take this formatting for all other amounts too.

moneymanolis:
@relativisticelectron I think I already found an easy way to fix the alignment again, will push later.

relativisticelectron:
> @relativisticelectron I think I already found an easy way to fix the alignment again, will push later.

something like: https://www.w3docs.com/snippets/html/how-to-render-an-html-string-preserving-spaces-and-line-breaks.html

moneymanolis:
> We should in a followup PR also take this formatting for all other amounts too.

Should be pretty easy to do that once we've settled on the filters etc. Could be done in this PR IMO as well. Let's see.

relativisticelectron:
> Should be pretty easy to do that once we've settled on the filters etc. Could be done in this PR IMO as well. Let's see.

Some of the btc amount formatting is done in javascript. https://github.com/cryptoadvance/specter-desktop/pull/1862 groups at least this javascript into functions.

relativisticelectron:
> @relativisticelectron I think I already found an easy way to fix the alignment again, will push later. I just used &nbsp.

Hmm...  I just hope this will later not become a problem, when we switch fonts and https://unicode-table.com/en/00A0/ and digits do not have the same width.

kdmukai:
> > While we're talking about alignment, has anyone tried seeing how it looks if the labels (e.g. "Total amount:") are right-justified so the colons all align?
> 
> image
> 
> I don't think this is an improvement.

I like it better, but I've had enough strong opinions for one PR!

moneymanolis:
I double-checked again:

grafik

moneymanolis:
Found this:
grafik

Source: https://en.wikipedia.org/wiki/Whitespace_character#Unicode

So, I'd just use` &numsp; ` and `&puncsp;`

relativisticelectron:
> "spaces at the beginning and end of an element are removed"

Since spaces are a problem on Mac, I pushed [6be6f09](https://github.com/cryptoadvance/specter-desktop/pull/1841/commits/6be6f0977aa6c20889b8b84c52a8dfd6ff92e62a) to now just make the 0's transparent. It works just as well, and the python function actually got easier to understand.

moneymanolis:
https://github.com/cryptoadvance/specter-desktop/pull/1841/commits/6be6f0977aa6c20889b8b84c52a8dfd6ff92e62a works like a charm. Great solution! 
I am still not sure whether we should inject HTML via Python, though, but it works pretty well, so I am fine with that.

I will write some tests now for the current status of `format_btc_amount()`. The actual rendering could be tested indirectly by letting Cypress click around in the dev helper:

grafik

But, I'll have to check whether this is going smoothly with the authentication. The alternative would be to insert a little check in one of the existing tests where we have balances.

relativisticelectron:
> I am still not sure whether we should inject HTML via Python, though, but it works pretty well, so I am fine with that.

I think ideal would be to move this function to javascript. Then it could be used later for dynamically loaded formatting too.

moneymanolis:
> Could we agree on font-size: 1.0em? I think font-size: 1.1em is just too big.

My thinking was: The balance display is now very nice to look at and with the colouring it appears a bit smaller. But, as you wish!

moneymanolis:
TODOs that I see:

- [ ] Check how the displays for Liquid look like
- [ ] At least a little bit of front end testing / checking of the rendering

moneymanolis:
> I think ideal would be to move all this amount and unit formatting to javascript. Then it could be used later for formatting dynamically loaded amounts too. However this can be done in a followup PR...

Or we use both, let's see, I also need to check your PR (https://github.com/cryptoadvance/specter-desktop/pull/1862) first to get a better understanding of how much dynamically loaded amounts we are actually talking about.

relativisticelectron:
> TODOs that I see:
> 
>     *  Check how the displays for Liquid look like

image

moneymanolis:
@relativisticelectron I'll work on the Cypress testing of the PR today.

moneymanolis:
@relativisticelectron can I change the transparent class to: 
```
.transparent-text{
    opacity: 0;
}
```
You had used` #00000000` which is using black and then applying the opacity of 0 I think. 
This would make it easier to read and - more importantly - Cypress is picking an opacity of 0 up as "not visible" (see here: https://github.com/cypress-io/cypress/pull/8244)

relativisticelectron:
Sure. Thanks for working on the tests.

relativisticelectron:
Bringing the coloring also to the sats Unit:
image

tests/test_util_common.py
```diff
@@ -76,10 +76,22 @@ def test_format_btc_amount():
 
 def test_format_btc_amount_as_sats():
     btc_amount = 0.00560000
-    assert format_btc_amount_as_sats(btc_amount) == "560,000"
```
moneymanolis: I would have left those tests and used `format_btc_amount_as_sats(btc_amount, enable_digit_formatting=False)` or if we change the default (see other comment by me in the PR), change accordingly, in any case makes sense to test **both**.

moneymanolis:
> Bringing the coloring also to the sats unit: image

Cool, that you also implemented this for the sats amount, but I would not enable the `digit_formatting` here by default. With the new great layout for BTC amounts, users in all likelihood will only change to sats as unit if they have smaller amounts of I guess 1 mio. sats or less. If you really want to keep it as default, I would at least change the colours to make them distinctively different from those used for the BTC amounts.

moneymanolis:
Regarding the Cypress tests for this PR: To make them reliable we'd need to deactivate the automatic mining in the Cypress tests (which doesn't seem to be necessary anyway). Otherwise, we have some sort of race condition, if the mining is fast the wallets involved have larger balances and vice versa. So far, I didn't manage to achieve that despite it looking so simple: Just using the` --no-mining ` or other changes to the mining loop in `cli_noded.py` always break the test setup for me.

tests/test_util_common.py
relativisticelectron: > in any case makes sense to test **both**.

yes.

relativisticelectron:
> Cool, that you also implemented this for the sats amount, but I would not enable the `digit_formatting` here by default.

ok. Can you disable it?

moneymanolis:
> > Cool, that you also implemented this for the sats amount, but I would not enable the `digit_formatting` here by default.
> 
> ok. Can you disable it?

OK, will do in the next commit.

tests/test_util_common.py
moneymanolis: OK, I'll commit this as well in the next commit.

moneymanolis:
@k9ert can you please check https://github.com/cryptoadvance/specter-desktop/pull/1841/commits/fc845892a2067b3dd053a95429a01d4766eb0ec9

k9ert:
> @k9ert can you please check https://github.com/cryptoadvance/specter-desktop/commit/fc845892a2067b3dd053a95429a01d4766eb0ec9

LGTM

moneymanolis:
@relativisticelectron - just fyi: I had a discussion with @k9ert on how to resolve the failing tests. Using the `--no-mining` flag creates more headaches then anticipated, we still need it to run the Cypress tests for this PR reliably. We will merge as soon as this issue is resolved. I really like this PR now. Great (team-) work!

PR #1842, opened 2022-08-18
Dev tools: Adding full python access via javascript for developers

relativisticelectron:
As a developer I often have the problem, that I do not know the values of some specter objects (e.g. wallets) during runtime.  `print` and restart server is currently my way to go until now.  

This PR allows full access read+write to everything in `app`.
- Disabled for all, except `DevelopmentConfig` and `current_user == "admin"`
- Warning when first running any command:
image
- if you think this is a too dangerous option to have, we can limit it to `regtest` and `testnet` only. The code is (as comments) already in the endpoint, see https://github.com/cryptoadvance/specter-desktop/pull/1842/commits/1e97bc03aeb919b4a6cb28e6037b80e28a6f0c00.

**Examples**

image

Simple `vars` and `dir` lookup via a trailing `.`  (since there is no autocomplete):
image


k9ert:
Oh, wow! You already implemented this! Would you mind to move that over to the dev-extension? I think we should move as much stuff as possible in extensions to not clutter "specter-core". This has the disadvantage that this console would not be available in prod that easily but maybe we can first do it there anyway and slowly consider then further steps in prod?

relativisticelectron:
> Would you mind to move that over to the dev-extension? 

Sure.  I have never worked with extensions. I have to figure out how the extensions work.

relativisticelectron:
> @k9ert : Can I add to/extend the base.jinja with the extension such that the change is present in all endpoints? I need to declare a javascript function that is available in all endpoints of specter. I couldn't figure out how yet...

I found a way to inject service templates into base.jinja.  Not super elegent, but it works.

k9ert:
> I found a way to inject service templates into base.jinja. Not super elegent, but it works.

Awesome! Very good! This is the way! [768eec8](https://github.com/cryptoadvance/specter-desktop/pull/1842/commits/768eec839a43065d317024c77dfd24557881cfb8)

>  I have to figure out how the extensions work. 

I guess you just did :-D.

k9ert:
Some more hints:
* `DEVELOPER_JAVASCRIPT_PYTHON_CONSOLE = True` The devhelp extension does not have yet a config-file. You can create one by simply  placing a `config.py` in its folder.  SOmething like:
```
"""
Here Configuration of your Extension takes place
"""


class BaseConfig:
    ''' This is a extension-based Config which is used as Base '''
    DEVELOPER_JAVASCRIPT_PYTHON_CONSOLE = True
```
* `server.py` : I'm not sure why you placed that in `calc_module_name` but migrate it in `callback_after_serverpy_init_app`. You should get an app instance by `from flask import current_app as app` ... at least i hope so.
* `controller.py` code goes to `devhelp/controller.py`
* `common.py` sounds reasonable generic to keep it there.

relativisticelectron:
I moved everything.  Did I do the `config.py` correctly?

k9ert:
> I moved everything. Did I do the config.py correctly?

Better than my proposal as in this case the console is deactivated in prod, even if someone manages somehow to put the devstatus of the devhelper extension to a `devstatus_prod`. So this is now ready to review, right?

relativisticelectron:
>   So this is now ready to review, right?

Yes. Please.

k9ert:
This PR created some criticism by the [Bitcoin Review Podcast episode 6](https://pca.st/wq2aaexf) (minute 13:34):
* They assume this is used to do automation: This is not true
* They assume it's built into Specter Core: This is not true
* They assume this can be used in production: This is not true either
* They assume that security issues can potentially occur in software which you use as a dependency: That is true

So let's see in more details, why this is a reasonable PR:

# Automation
So first of all the intention of this PR can be read in the first sentence:

> As a developer I often have the problem, that I do not know the values of some specter objects (e.g. wallets) 
> during runtime. print and restart server is currently my way to go until now.

The way this is implemented doesn't make automation easy, at least not in any way i can imagine. Excel can be "integrated" with our csv-export.

# Specter Core?
We have an extension-system. If you look into the files which got modified/created with this PR, most of it (and all functionality) happens in `/src/cryptoadvance/specterext/devhelp` which is a special extension, shipped in the with the core-product but still separated from core.
Extensions are not only usefull to let the user load additional functionality, but they are also an important architectural pattern to keep things tidy and separated, don't create spagetti-code and finally avoid security-issues if you do it the right way. What's the right way?

# Questionable extensions on production

Some extensions are simply not made for production as they contain potential security-holes (like this one) but provide usefull functionality for developing or testing the application. This is the case for the devhelp-extension. Other extensions might simply be in alpha status and not yet mature enough to run in production.

In order to prevent that people run alpha software in prod, there is a hardcoded attribute in each extension, defining the devstatus. That's some sort of maturity-level of that plugin. For the devhelp-Extension, that's alpha (and not planned to ever change):
image

Also if you start Specter, you can specify which configuration to use. By default, the production-configuration is used and that configuration is not accepting that devstatus. You'll read a line like this in the logs:
```
SPECTERD: INFO in service_manager: ----> starting service loading
SPECTERD: INFO in service_manager: Service DevhelpService not activated due to devstatus ( prod > alpha )
```

# Other remarks

As a software-producer, here are the top 3 things to do if you're a beginner in this game:
* Be careful with dependencies. Limit them and don't use questionable dependencies.
* Pin the dependencies (`requirements.txt`) to specific versions. If you don't do that, you might drag in new versions in an uncontrollable way which might introduce security issues (or incompatibilities or bugs)
* In python use `pip-compile --generate-hashes` in order to make it possible for installations to check the hashes of your dependencies
* Spent time and effort maintaining your dependencies upgrading them to keep your system up-to-date and secure. There are services out there which create automatic PR and notify you about security issues in your dependencies. Use them!

We're doing all of that, but nevertheless: Yes, there are zero-day-attacks out there! Yes, that might affect Software you're using! 
Take care, it's a dangerous world out there! Limit your exposure to potential malicious code.

relativisticelectron:
I think the title of the PR (and then visible in the Release notes) could have been chosen more precisely, like: "Dev tools: Adding full python access via javascript for developers in developer-extension"

PR #1850, opened 2022-08-23
Feature: Improve Dev-Console

k9ert:
I think i was a bit too fast merging the original PR. This improves on the DevConsole by:
* Most recent results now on top and not appended on bottom
* `current_user` available and also the `namespace`
* methods and properties in foldable containers

image


relativisticelectron:
Executing 
image
for the first time is not compatible with `NEVER_CALLED_PYTHON_COMMAND`

image

src/cryptoadvance/specterext/devhelp/service.py
```diff
@@ -26,6 +27,8 @@ class DevhelpService(Service):
             "app": app,
             "flask": flask,
             "flask_login": flask_login,
+            "current_user": current_user,
```
relativisticelectron, line 30: This doesn't work:
image

It is accessible via:
image

src/cryptoadvance/specterext/devhelp/service.py
```diff
@@ -26,6 +27,8 @@ class DevhelpService(Service):
...
+            "namespace": console.namespace,
```
relativisticelectron, line 31: Can we remove this again?
image

relativisticelectron (commented):
see comments.  For me the javascript console in any browser (developer mode) is completely sufficient (and available in any endpoint and syntax highlighting).  Is there a benefit to an in-website console?

relativisticelectron:
Could you add the `util`  (from `import util`) to the namespace? The tools there are not accessible otherwise.

src/cryptoadvance/specterext/devhelp/service.py
k9ert: Fxed!

src/cryptoadvance/specterext/devhelp/service.py
k9ert: done

k9ert:
> Could you add the util (from import util) to the namespace? The tools there are not accessible otherwise.

Done.

> for the first time is not compatible with NEVER_CALLED_PYTHON_COMMAND

Fixed

> For me the javascript console in any browser (developer mode) is completely sufficient (and available in any endpoint and has syntax highlighting). Is there a benefit to an in-website console?

IMHO the main 2 arguments are accessibility and convenience: 
* In a large project with many features, often enough you don't know that a feature is existing because it's badly documented (or not at all). Making it visible explicitely might be a huge deal, even if you don't use it but know that it's existing
* Even if you know that it's existing, what's the syntax again? Also, as a non frontend engineer, you might be not that familiar with the console or you might want to keep results on a different page across requests. I like the term "mechanism, not policy".

Makes sense?

relativisticelectron:
> > For me the javascript console in any browser (developer mode) is completely sufficient (and available in any endpoint and has syntax highlighting). Is there a benefit to an in-website console?
> 
> IMHO the main 2 arguments are accessibility and convenience:
> 
>     * In a large project with many features, often enough you don't know that a feature is existing because it's badly documented (or not at all). Making it visible explicitely might be a huge deal, even if you don't use it but know that it's existing
> 
>     * Even if you know that it's existing, what's the syntax again? Also, as a non frontend engineer, you might be not that familiar with the console or you might want to keep results on a different page across requests. I like the term "mechanism, not policy".
> 
> 
> Makes sense?

Yes, makes sense :-)

PR #1852, opened 2022-08-25
Mobile: Optimisation of tx-table

moneymanolis:
This drops the time column for mobile. Would look like that:

grafik


moneymanolis:
@relativisticelectron what do you think? I'd prefer this. Could you test in your local "mobile" setup?

relativisticelectron:
> @relativisticelectron what do you think? I'd prefer this. Could you test in your local "mobile" setup?

yes, that works!

PR #1854, opened 2022-08-25
UI: Do not hide elements automatically by default

relativisticelectron:
See https://github.com/cryptoadvance/specter-desktop/pull/1170#issuecomment-1212735785 and following comments


moneymanolis (approved):

moneymanolis (changes requested):
Just tested the setting of the timeout for an individual user and got this error below which prohibits to save the new value in `users.json`. We'd need to fix that in this PR, too, IMO. (otherwise the user will never have the chance to have autohiding enabled).

grafik

relativisticelectron:
I cannot reproduce the error.  I can save any value in the UI.

moneymanolis:
As admin, yes, but add a user and try with this user.

relativisticelectron:
> As admin, yes, but add a user and try with this user.

Ok, I can reproduce this error.

moneymanolis:
I'll have a look at it. I'd push it directly here if it is a little fix.

moneymanolis:
@relativisticelectron can you double check that the fix in https://github.com/cryptoadvance/specter-desktop/pull/1854/commits/f455b651d982380c0bd90d9e3c3e600ca081ac68 works. @k9ert can you approve?

moneymanolis (approved):

relativisticelectron:
> @relativisticelectron can you double check that the fix in [f455b65](https://github.com/cryptoadvance/specter-desktop/commit/f455b651d982380c0bd90d9e3c3e600ca081ac68) works. Also added some testlines for the defaults. @k9ert can you approve?

Yes is works now.

k9ert (approved):

Issue #1858, opened 2022-08-26
Bug during loading: KeyError: 'blocks'


relativisticelectron:
**Describe the bug**
During loading shortly was this error message visible:
```
[2022-08-26 12:12:36,251] INFO in checker: Checker health Checking every 600 seconds now
[2022-08-26 12:12:36,266] ERROR in controller: Uncaught exception: 'blocks'
[2022-08-26 12:12:36,267] INFO in wallet_manager: Loading <wallet-name> to Bitcoin Core
[2022-08-26 12:12:36,270] ERROR in controller: Traceback (most recent call last):
  File "flask/app.py", line 1523, in full_dispatch_request
  File "flask/app.py", line 1509, in dispatch_request
  File "flask_login/utils.py", line 272, in decorated_view
  File "cryptoadvance/specter/server_endpoints/wallets.py", line 57, in wallets_overview
  File "cryptoadvance/specter/specter.py", line 226, in check_blockheight
  File "cryptoadvance/specter/node.py", line 400, in check_blockheight
KeyError: 'blocks'
```

Afterwards specter worked fine.

**Desktop (please complete the following information):**
 - Specter Version 1.12.0,  mainnet

PR #1859, opened 2022-08-26
Bugfix: during loading: KeyError: 'blocks'

relativisticelectron:
See https://github.com/cryptoadvance/specter-desktop/issues/1858

I cannot reproduce the bug https://github.com/cryptoadvance/specter-desktop/issues/1858 , however this bugfix PR seems reasonably straight forward to me.


relativisticelectron:
I think I can reproduce the error (elements node running, and then switching to the non-running btc node) and the current fix, only solves it partially.

Issue #1863, opened 2022-08-27
Improve message when adding Ledger device


wombat6:
When adding a new Ledger device, the current message says something like "Connect your Ledger". Then Specter doesn't find the device over USB. The problem I had was I didn't realize I had to activate the Bitcoin app on my Ledger before Specter would see it. My suggestion is to change the message to something like "Connect your Ledger. Be sure to activate the Bitcoin app on your Ledger so Specter can see the device."

moneymanolis:
I am on it. I think we also need to add a device check when setting up a new device (at the` new_device_keys.` endpoint). Seems like this is not done for HWI / USB devices.

PR #1864, opened 2022-08-29
UIUX: Individual HWI progess for Ledger devices

moneymanolis:
### Background 
Apparently, Ledger devices can only be connected if the user has also activated the Bitcoin app on his ledger.
See #1863
### Contents
- Add specific HWI progress message for Ledger (hwi.jinja)
- Fixes to use device_types (other jinja files) - the last thing was actually kind of a bug: It was for example possible to select a Ledger as device, connect a different USB HWW like a Trezor and HWI still worked. 
### Visuals
Screenshot 2022-08-29 at 11 50 59
After a couple of unsuccessful attempts:

Screenshot 2022-08-29 at 11 51 10

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -1,3 +1,24 @@
+# Release Notes
+
+## v1.12.0 August 26, 2022
+- Feature: add faucet and exfund extensions #1820 (Stepan Snigirev)
+- Feature: Dev tools - Adding full python access via javascript for developers #1842 (relativisticelectron)
+- UIUX: Complete overhaul of the tooltips used in Specter Desktop #1813 (Manolis Mandrapilias)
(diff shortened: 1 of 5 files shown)


wombat6:
Confirmed that the message shows up on the "Detecting..." dialog after you click "Get via USB".

One minor suggestion: it would be good if the message appeared on the "Upload Keys" page itself, where it says "Connect your hardware device to the computer via USB." That way the user knows before trying to connect. I'm glad to submit a PR if you agree with that.

wombat6:
One more suggestion. In my original issue, I used the word "activate". Ledger uses the word "open": "Open the Bitcoin app." To avoid confusion, I suggest using the same word.

moneymanolis:
OK, will change it to "open".

PR #1868, opened 2022-09-01
Added "Open Bitcoin app" verbiage to Ledger Upload Keys page

wombat6:
Targets issue #1863, in addition to the PR that @moneymanolis
submitted.

The idea is to spare the user the annoyance of having to
cancel the upload because s/he didn't know to open the Bitcoin
app before pressing "Upload via USB".

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -1,3 +1,24 @@
+# Release Notes
+
+## v1.12.0 August 26, 2022
+- Feature: add faucet and exfund extensions #1820 (Stepan Snigirev)
+- Feature: Dev tools - Adding full python access via javascript for developers #1842 (relativisticelectron)
+- UIUX: Complete overhaul of the tooltips used in Specter Desktop #1813 (Manolis Mandrapilias)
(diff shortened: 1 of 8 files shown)


moneymanolis:
Would that work for you, too? A bit less attention grabbing for those users who know about having to activate the Bitcoin app beforehand:
grafik
If so, I'd push directly here. 

Just to clarify: If the user activates the Bitcoin app after pressing the button, which is basically starting the enumerating process in hwi, enumerate does not work?

wombat6:
Oh, you are right -- if I open the Bitcoin app while Specter is trying to import the keys, Specter will find them.

Let me know what you want to do with this PR. Fine with closing it without merging, and also fine with making the changes you suggest and keeping it. Your call. If we keep it, I'd still recommend using the word "open" rather than "activate" since Ledger Live describes the process as opening the Bitcoin app.

moneymanolis:
> Oh, you are right -- if I open the Bitcoin app while Specter is trying to import the keys, Specter will find them.

That's good! Doesn't hurt to have the extra info beforehand, IMO. I'll push here and will rename to open.

Issue #1867, opened 2022-09-01
Cobo Vault "Upload Keys" page displays escaped HTML


wombat6:
The "Upload Keys" page for Cobo Vault displays escaped HTML tags instead of displaying them as actual HTML.

**To Reproduce**
1. Go to "Add New Device"2. 
2. Select Cobo Vault.
3. Look at the text that begins "To get the master public keys QR codes to scan".

**Expected behavior**
Display HTML, not escaped HTML

**Screenshots**
![Screen Shot 2022-08-31 at 10 45 53 PM](https://user-images.githubusercontent.com/112285082/187846135-99fb0a62-60ff-4b96-b87c-df6f1d3b3883.png)

**Desktop (please complete the following information):**
 - Specter Version: 1.12.0

wombat6:
I'm glad to take this one.

wombat6:
ColdCard markup has the same issue. Just including that within this issue since it's basically the same problem and can be corrected with the same PR.

wombat6:
Anyone care to merge this one?

PR #1871, opened 2022-09-02
Corrected escaped HTML on Cobo Vault Upload Keys page

wombat6:
See issue #1867. This removes the HTML escaping by fixing the incorrect use
of the Jinja _() alias.


wombat6:
Here's a screenshot of the corrected markup:

Screen Shot 2022-09-01 at 10 36 57 PM

Note that I also replaced the `->` symbol with &rarr; which looks a lot better.

wombat6:
Corrected ColdCard page:

Screen Shot 2022-09-01 at 10 47 44 PM

moneymanolis:
Thanks for fixing this @wombat6!

Issue #1865, opened 2022-08-29
slider for fees not working in dynamic mode


topuli:
**Describe the bug**
It is not possible to change the fees amount usind the slider in dynamic mode.
Opening the advanced section immediately after clicking send, the slider ist in the middle position. After a few seconds it moves to the left (very low). From here on i am not able to change the fees using the slider

**To Reproduce**
1. Go to send
2. Click on advanced
3. Watch the slider

**Expected behavior**
Ability to change the fees amount using the slider

**Screenshots**
![image](https://user-images.githubusercontent.com/48984696/187247942-5461b5d1-53cb-401d-8328-8473fc9fa944.png)

**Desktop (please complete the following information):**
 - Where is your node running: local (internal)
 - Where is your specter running: binary
 - If you're running a binary, which OS: Linux
 - Browser Firefox
 - Specter Version 1.10.5

relativisticelectron:
I can confirm this bug.  In the beginning the slider works as expected, but then the slider gets stuck at minimum and can't move any more.

relativisticelectron:
I guess the cause is a currently low fee environment and then https://github.com/cryptoadvance/specter-desktop/blob/9dfb6c3a775b18ccc0734795e7b878f0a42331b3/src/cryptoadvance/specter/templates/includes/fee-selection.html#L241  sets the minimum fee rate....

relativisticelectron:
Was solved in  https://github.com/cryptoadvance/specter-desktop/pull/1878

PR #1873, opened 2022-09-04
Bug fix: stuck fee slider in minimum fee environment

relativisticelectron:
See: https://github.com/cryptoadvance/specter-desktop/issues/1865

The fix is to round up the fastest rate, resulting in 2 sat/vbyte in a maximum feerate in a minimum feerate environment.  
Now it looks like:
Screenshot_20220904_162939

### DIFF
--- a/src/cryptoadvance/specter/templates/includes/fee-selection.html
+++ b/src/cryptoadvance/specter/templates/includes/fee-selection.html
@@ -238,7 +238,8 @@
         // Gets executed after fee initialisation
         initWithFees() {
             // this.feesSlider.min is set above via min_fee
-            this.feesSlider.max = Math.floor(this.fees["fastestFee"] * 1.4)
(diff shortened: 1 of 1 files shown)


moneymanolis:
While testing, I saw that an additional category (`economyFee`) was added to the mempool space API (https://github.com/mempool/mempool/issues/1726) which would make the fee recommendation in Specter a bit more granular (Specter is fetching the recommended fee categories similar to `curl -sSL "https://mempool.space/api/v1/fees/recommended"`). I will make a separate PR with that and also ensure that the slider is still functional in a low fee environment where the result from mempool space basically is:
`{"fastestFee":1,"halfHourFee":1,"hourFee":1,"economyFee":1,"minimumFee":1}`

@relativisticelectron leave the PR open for now, in case I don't manage to finish it off quickly.

relativisticelectron:
Was replaced by https://github.com/cryptoadvance/specter-desktop/pull/1878

PR #1766, opened 2022-06-18
Feature: Notification system

relativisticelectron:
This is a notification system for sync and async messages
Issue: https://github.com/cryptoadvance/specter-desktop/issues/1695

**Features:**
- A `NotificationManager` (1 global one) forwards notifications to UINotifications (such as `LoggingNotifications`, `FlashNotifications`, `JSConsoleNotifications`, `JSNotifications`, `WebAPINotifications` , which are instantiated for each user)
- Notifications can be created from _python_ and from _javascript_ and handled both through `NotificationManager` in an identical way
- Individual notifications can be sent to one or multiple UINotifications
- Events such as onClose are possible for  `JSNotifications`, `WebAPINotifications`
- `WebAPINotifications` (= [Notifications API](https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API/Using_the_Notifications_API) ) is only available in localhost, tor-browser or https 
- If one UINotifications is not available, the notification is forwarded to a fallback UINotifications

Peek 2022-06-18 14-48
- Example call from javascript:
```
CreateNotification('this is the title', {target_uis:['js_message_box', 'WebAPI'], body:'body line 1\nline 2', image:'/static/img/ghost_3d.png', timeout:3000})
```

**Further things that are enabled by this**
- Any "flash" messages could automatically also be logged, by changing 
```
    def flash(self, *args, **kwargs):
       .....
        kwargs["target_uis"] = {"flash"}
       .....
``` 
(this could even be in some config file) to 
```
    def flash(self, *args, **kwargs):
       .....
        kwargs["target_uis"] = {"flash", "logging"}
       .....
``` 
- one can create python logging messages from javascript by calling
```
CreateNotification('this is a message', {target_uis:['logging'], body:'this is the body of the message'});
```
- further callbacks, like onclick, onhover,.... are possible for `JSNotifications`, `WebAPINotifications`  are straight forward to implement. One could even extend this to add Buttons in the Notifications.
- more UINotifications (even if it is just a different design of `JSNotifications`, or you want to send a rpc command to another process, or start a program like [notify-send](https://www.thegeekstuff.com/2010/12/ubuntu-notify-send/)) can be added very easily.

**Limitations**
- `WebAPINotifications` (= [Notifications API](https://developer.mozilla.org/en-US/docs/Web/API/Notifications_API/Using_the_Notifications_API) ) does not seem to support images in _Linux_, and _Android_ for Firefox or Chromium.

**Future extension ideas**

- persistent storage of messages 
[...]

PR #1883, opened 2022-09-12
Chore: redirect flash call

k9ert:
In order to shrink PR #1766 , this PR is refactoring all calls to flash to a central point which can later (in that PR or otherwise) be treated differently. Also (and visually the most changes) sorting imports.


src/cryptoadvance/specter/server.py
```diff
@@ -62,6 +62,14 @@ def calc_module_name(config):
         return "cryptoadvance.specter.config." + config
 
 
+def flash(*args, **kwargs):
```
relativisticelectron, line 65: I am unsure why this double-definition of flash should ever work and not create a recursive loop....

relativisticelectron: Why move the flash function in server and not in util?

src/cryptoadvance/specter/server.py
```diff
@@ -10,7 +10,7 @@
 from cryptoadvance.specter.services import callbacks
 from cryptoadvance.specter.util.reflection import get_template_static_folder
 from dotenv import load_dotenv
-from flask import Flask, jsonify, redirect, request, session, url_for
+from flask import Flask, jsonify, redirect, request, session, url_for, flash
```
relativisticelectron, line 13: I would do `from flask import flash as flask_flash`

relativisticelectron (commented):
I didn't test the code yet, comments are from looking at the changes.

src/cryptoadvance/specter/server.py
k9ert: Good point. Renamed it. I think everything which is called "util" or "helper" or something like that is suppose to not be dependent on the state of objects.
Doing something like `current_app.specter.some_manager.some_method()` in a util package is imho a very bad idea. There, you want to have deterministic functions or objects which don't do much magic apart maybe from using external libraries which themself also are deterministic.

So `server.py` was probably also a bad idea and i moved it to `server_endpoints/__init__.py`

src/cryptoadvance/specter/server.py
k9ert: agree

src/cryptoadvance/specter/managers/wallet_manager.py
```diff
@@ -20,6 +20,8 @@
     Wallet,
     purposes,
 )
+from flask import current_app
```
relativisticelectron, line 23: `current_app` is not used.

src/cryptoadvance/specter/util/wallet_importer.py
```diff
@@ -1,18 +1,19 @@
...
 from embit.descriptor import Key as DescriptorKey
 from embit.descriptor.arguments import AllowedDerivation
 from embit.liquid.descriptor import LDescriptor
-from cryptoadvance.specter.key import Key
-from flask import flash
 from flask_babel import lazy_gettext as _
 
+from ..key import Key
+from ..managers.wallet_manager import WalletManager
+from ..server import flash
```
relativisticelectron, line 13: This should be removed, right?

relativisticelectron (commented):
`flash` also appears here:

- https://github.com/cryptoadvance/specter-desktop/blob/f2fb8f0cd7c086ac0d01791639dd2167e616ca34/src/cryptoadvance/specter/services/bitcoinreserve/controller.py#L2
- https://github.com/cryptoadvance/specter-desktop/blob/f2fb8f0cd7c086ac0d01791639dd2167e616ca34/tests/xtestdata_testextensions/ext_root_adhoc_1/breznguat/controller.py#L2
- https://github.com/cryptoadvance/specter-desktop/blob/f2fb8f0cd7c086ac0d01791639dd2167e616ca34/tests/xtestdata_testextensions/ext_root_fully_qualified_1/src/boatacccorp/specterext/tretboot/tretboot/controller.py#L2
- https://github.com/cryptoadvance/specter-desktop/blob/f2fb8f0cd7c086ac0d01791639dd2167e616ca34/tests/xtestdata_testextensions/ext_root_fully_qualified_2/src/boatacccorp/specterext/ruderboot/controller.py#L2


EDIT: solved.

src/cryptoadvance/specter/server.py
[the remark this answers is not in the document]
moneymanolis: I'd have the same question :-)

src/cryptoadvance/specter/server_endpoints/__init__.py
```diff
@@ -0,0 +1,9 @@
+from flask import flash as flask_flash
+
+
+def flash(*args, **kwargs):
+    """An indirection in order to potentially handle that differently
```
moneymanolis, line 5:
```suggestion
    """An indirection in order to potentially handle a flash differently.
```

src/cryptoadvance/specter/server_endpoints/__init__.py
```diff
@@ -0,0 +1,9 @@
...
+    This function could potentially be placed in util but as it might
```
moneymanolis, line 6:
```suggestion
    This function could be placed in util but as it might
```

src/cryptoadvance/specter/util/wallet_importer.py
moneymanolis: +1

moneymanolis, line 13:
```suggestion

```

moneymanolis (approved):
See little comments / suggestions before merging.

src/cryptoadvance/specter/server.py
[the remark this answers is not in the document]
relativisticelectron: Ok. solved.

relativisticelectron (approved):

PR #1887, opened 2022-09-15
Feature: Extensions can add new devices

k9ert:
This PR introduces the possibility to define new Devices within extensions.

As an example and first usecase, the electrum device has been moved to a dedicated extension called electrum.

Also the `ext gen` process has been adapted in order to create device-example right away.

As further ideas, a device-class could have attributes like e.g. `new_device_keys_redirect` which are named after the device-endpoints and the endpoints redirect to extension-specific endpoints, if a device has set them. This would make customisation much easier. Haven't implemented that yet, waiting for the first requirements coming up.


src/cryptoadvance/specter/util/reflection.py
```diff
@@ -78,7 +78,7 @@ def get_classlist_of_type_clazz_from_modulelist(clazz, modulelist):
             module = import_module(fq_module_name)
         except ModuleNotFoundError as e:
             # ToDo: make it somehow clear where specific extensions are coming from: external or within same repo
-            logger.warn(
+            raise SpecterError(
```
relativisticelectron, line 81: Please revert this change. This additional raising of an exception is triggered every time for me 

```
ModuleNotFoundError: No module named 'cryptoadvance.specterext.liquidissuer'

During handling of the above exception, another exception occurred:

```

relativisticelectron:
A few remarks about the services:
- In `/services/choose` there should be:
   1. An activate/deactivate checkbox for already installed services. Services should be loaded/unloaded on the fly.
   2. A (download &) **Install & Activate** button.
- `ext/bitcoinreserve/` leads to `jinja2.exceptions.TemplateNotFound: bitcoinreserve/index.jinja`

src/cryptoadvance/specter/util/reflection.py
k9ert: Why is that the case for you? This doesn't happen on my computer. In the `requirements.in`, you can read:
```
# Extensions
cryptoadvance-liquidissuer==0.2.4
specterext-exfund==0.1.7
specterext-faucet==0.1.2
```
So this should not happen. `continue` here is imho hiding configuration issues exactly like the one you currently have?!.

relativisticelectron: I agree that your change reveals previously existing problems. 

However I just did `pip install --require-hashes -r requirements.txt`  and still get this error when I run 
```
python3 setup.py install && python3 -m cryptoadvance.specter server --config DevelopmentConfig     --port 1234   --ssl --debug
```

Is there anything else needed for these services?

docs/extensions/intro.md
```diff
@@ -0,0 +1,170 @@
+# Introduction
+
+Extensions is a way to extend Specter Desktop in a way which does not need PRs in the core-project. They can be used
+to 
+* create elaborated Frontends on top of Specter Desktop or even replace the whole Frontend
+* Add Additional Menu Items currently in Wallets and Settings
+* Create additional Devices with the functionality you would like to have
```
wombat6, line 7: The Markdown for a list item is a hyphen, not an asterisk.

docs/extensions/intro.md
```diff
@@ -0,0 +1,170 @@
...
+            python3 -m twine upload --repository testpypi dist/*
+
+        You can get all this information again via:
+        python3 -m cryptoadvance.specter ext gen --help
+$ 
+```
+The created file structure looks like this and you will feel right at home if you have some knowledge about how Flask works:
+
+
+![](./images/extensions_file_layout.png)
```
wombat6, line 91: Missing image?

docs/extensions/intro.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Virtualenv Management
+
+The simplified example of creating an extension above is ommitting the use of a virtual environment. However, it's really __highly__ recommended that virtual environments are used for extension development. Without the Output above you would do something like:
```
wombat6, line 95: Since we're recommending virtual envs, wouldn't it make sense to move this before the non-VE approach?

docs/extensions/intro.md
```diff
@@ -0,0 +1,170 @@
...
+
+Often enough, extension development goes hand in hand with the development of Specter Desktop iteself. Especially in the very beginning you need to constantly develop on the extension-concept on Specter Desktop while also using/consuming that functionality in an extension.
+
+In such a case, the best option is to "reuse" the virtualenv of Specter Desktop also for your extension. Other than what is described above, you would do something like the below. If you have checked out the Specter Desktop Code in parallel with the extension-code, you would ommit the creation of the virtualenv above and rather activate the existing one in your Specter Desktop Development Directory:
+
+```
+. ../specter-desktop/.env/bin/activate
+pip3 install -e .
+python3 -m cryptoadvance.specter server --config DevelopmentConfig --debug
+```
```
wombat6, line 117: Maybe it doesn't make a difference, but I think of it more as creating a virtualenv for the extension and then installing the Desktop core into my extension virtualenv, instead of the reverse. That way my extension development (including virtualenv) is totally self-contained -- if I delete the extension directory, I'm back to clean.

wombat6 (changes requested):

docs/extensions/intro.md
k9ert: You're making me angry! Anyway, after calming down i finally understood some of the strange effects which puzzled my markdown life and also what wishful thinking can do to the world!
Just kidding, thanks, very good point ;-)

docs/extensions/intro.md
k9ert: yes, missing dot.

docs/extensions/intro.md
k9ert: Yes, agreed. I've included the use of the virtual env directly in the TL;DR section.

docs/extensions/intro.md
k9ert: I agree, it's just consequent. It's just that i didn't do that until now, but i'll change my practice here. I already suffered at least once because of unclear dependencies.
However, i've not yet tested that practice yet. So I've added it as the better approach but not yet tested.

moneymanolis (approved):

Issue #1886, opened 2022-09-14
Missing PSBT_GLOBAL_XPUB fields in generated PSBTs


bigspider:
**Describe the bug**

Generating an xpub for a multisignature wallet, the psbt should include the global xpubs for [BIP-0174](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki) compliance. The absence could create problems when trying to use the PSBT with software that expects the field to be filled.

**To Reproduce**
Create a transaction for a multisignature wallet.

**Expected behavior**
All the global xpubs are present 

**Example**
I generated the following PSBT for a 3-of-3 multisig:
```
cHNidP8BAFICAAAAAdw1zgN1ummWFzAPKzo5tssHR5CTFvzjwx2KqLGXQHyxAAAAAAD9////AccMAQAAAAAAFgAU/52lZ+YvMOqGVPodX71HvvjjvhMhlRUAAAEAfQIAAAABv9tbc/eJXLb2aC9/TGL9I/rp6gIpNvHO+z+6jl8p+HgAAAAAAP7///8CiA0BAAAAAAAiACDOfJQ4UBk76jpzVUk6G5CE9e9rdDoYBJiULo61486YrL2c4AQAAAAAFgAUwc6ro8YBrajLUWn5YjvAqpv0FIcOziMAAQEriA0BAAAAAAAiACDOfJQ4UBk76jpzVUk6G5CE9e9rdDoYBJiULo61486YrAEFaVMhAlAXEIvs6o3txTALsiOGs6swNnrCYvnOXlgybrg+OiL1IQLdG+hu1ye9khPmIRP5oVrMr77ANG8zcCbT2xDYnhscVyEDaxbowfl5+kzA8FtqMAr//5QUWbbyDed95VsBYO+OTKxTriIGAlAXEIvs6o3txTALsiOGs6swNnrCYvnOXlgybrg+OiL1HHYiOm4wAACAAQAAgAAAAIACAACAAAAAAAAAAAAiBgLdG+hu1ye9khPmIRP5oVrMr77ANG8zcCbT2xDYnhscVxziQkO0MAAAgAEAAIAAAACAAgAAgAAAAAAAAAAAIgYDaxbowfl5+kzA8FtqMAr//5QUWbbyDed95VsBYO+OTKwc9azC/TAAAIABAACAAAAAgAIAAIAAAAAAAAAAAAAA
```
[...]

PR #1888, opened 2022-09-15
Add non-witness-utxos and xpubs when using hwi with ledger

bigspider:
Currently, Specter Desktop doesn't add the global xpubs to the psbt when signing transactions with ledger devices. Since [HWI uses them](https://github.com/bitcoin-core/HWI/blob/e731395bde13362950e9f13e01689c475545e4dc/hwilib/devices/ledger.py#L288-L295) to figure out the right parameters for the hardware wallet, HWI falls back to the legacy protocol.

The next release of the Ledger bitcoin app (version 2.1.0) will remove compatiblity with the legacy protocol (before version 2.0.0 of the app), so it will not be able to sign with HWI psbts without the global xpubs.

This instructs Specter to add the xpubs.
Moreover, it also adds the non-witness-utxos, since a warning would be shown if the non-witness-utxo is missing on segwitv0 transactions.

Closes: #1886

### DIFF
--- a/src/cryptoadvance/specter/devices/ledger.py
+++ b/src/cryptoadvance/specter/devices/ledger.py
@@ -8,4 +8,4 @@ class Ledger(HWIDevice):
     icon = "ledger_icon.svg"
 
     def create_psbts(self, base64_psbt, wallet):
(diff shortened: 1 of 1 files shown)


k9ert:
Tested and worked with my Ledger! Thank you very much appreciated!

PR #1892, opened 2022-09-20
Privacy Fix: Fix possible Chrome data leakage by disabeling spellcheck

relativisticelectron:
The article  https://www.otto-js.com/news/article/chrome-and-edge-enhanced-spellcheck-features-expose-pii-even-your-passwords 
describes that data could be sent to a a 3.party spell-checker. `spellcheck="false"` for inputs is supposed to mitigate the risk.

Since all data in specter is sensitive and none of it requires a spellcheck, This PR adds `spellcheck="false"` to all input fields.

**Possible alternatives:**

- Add a custom <specterinput> web component that has `spellcheck="false"` and replace all inputs/textarea by that.
- Add a pre-commit python script that add `spellcheck="false"` to all input fields in all directories. Not super elegant, but it would solve the maintenance issue.
- https://stackoverflow.com/questions/47227173/disable-spellcheck-and-autocomplete-globally-in-a-html-document seems to be ideal, but it doesn't work because most inputs are in a shadowDOM

**Issues with this solution:**

-  `spellcheck="false"` needs to be added to all input fields in the future, which is bound to be forgotten
- This solution doesn't cover services in other repositories

**Is this a problem?**
Yes, on master I get in Chromium with (basic) spellcheck in Chromium enabled
image
If the users had in the past enabled the "enhanced" spellcheck, the input data will be transmitted to a 3. party.

**TODO:**
- [ ]  test if any issues arise

@moneymanolis : I am unsure how to solve this issue best. What do you think?


moneymanolis:
@relativisticelectron I'll reproduce and check this tomorrow. At first glance, I am wondering whether we can't set the attribute in the body, this is how lastpass has apparently solved it:
grafik

relativisticelectron:
Good idea and it works!

moneymanolis (approved):

moneymanolis:
I couldn't reproduce the leakage in Specter Desktop (and neither on random sites btw.) but doesn't hurt to have this disabled since we don't need it.

PR #1893, opened 2022-09-21
Feature: add_device_mnemonic menu adaption

k9ert:
* needs lots of refactorings
* Should probably need a refactoring of HotWallet into a dedicated extension


k9ert:
Doesn't seem to be needed right now. Closing.

PR #1862, opened 2022-08-27
UI bugfix: Fix missing values in tx details

relativisticelectron:
there was a html-bug.

Before:
image

Now:
image

PR #1900, opened 2022-09-24
UI Bug: Fix float issue

relativisticelectron:
Unclear if https://github.com/cryptoadvance/specter-desktop/pull/1862 introduced this, however the display of the output values made float issues apparent:

image

This PR fixes the formatting, while it does not solve the underlying issue that `rawtx.vout[i].value`  is a float and has (sometimes) an error at the last digit.  A discussion about internal representation via integer/decimal/float was also here: https://github.com/cryptoadvance/specter-desktop/pull/1841#issuecomment-1234231422


relativisticelectron:
I solved by https://github.com/cryptoadvance/specter-desktop/pull/1901  already

relativisticelectron:
It seems https://github.com/cryptoadvance/specter-desktop/pull/1901 is more complex, so it would be good to fix this issue asap with this PR.

k9ert:
the cypress-test seem to fail with this:
```javascript
        // The fee should be subtracted from recipient 2, so the amount should be less than 2
        cy.get('div.tx_info > :nth-child(1) > :nth-child(1)').then(($div) => {  // nth-child is indexed from 1 https://css-tricks.com/almanac/selectors/n/nth-child/
            amount = parseFloat($div.text())
            expect(amount).to.be.lt(2)
            expect(amount).to.be.gt(1)
        })
```
See here: https://cirrus-ci.com/task/5413513951182848
Can you have a look at the tests? Is parseFloat here still the right function?

src/cryptoadvance/specter/templates/includes/helpers.jinja
```diff
@@ -34,11 +34,15 @@
...
             if (convertToSat && (["Lsat", "sat", "tLsat", "tsat"].indexOf(formattedUnitLabel) > -1)) {
-                formattedValue = parseInt(value * 1e8)
+                // if it is sat, then format at integer
+                formattedValue = parseInt(parseFloat(valueInBTC) * 1e8);
+            }
+            else{
+                // else, then cut off at 8 digits
+                formattedValue = parseFloat(valueInBTC).toFixed(8);
             }
             formattedValue = `${numberWithCommas(formattedValue)}`;
```
moneymanolis, line 47: `numberWithCommas` only makes sense to apply to the sats amount, no? But, we are getting rid of the function anyway, so just leave it.

relativisticelectron: I thought it also makes the commas in here 
image

moneymanolis: Ah, true!

moneymanolis:
I know the tests are green but I want to merge these changes from https://github.com/cryptoadvance/specter-desktop/pull/1901 below to make the tests less flaky:

grafik

I'll merge afterwards.

moneymanolis (approved):

Issue #1861, opened 2022-08-27
UI Improvement needed:  Transaction history added/substracted amounts unclear


relativisticelectron:
Receiving, sending and self transfers all show up as white amounts:
![image](https://user-images.githubusercontent.com/60378539/187044298-3f6a728d-dd0f-47b6-8093-2ad471809e49.png)

It is completely unclear what is added or subtracted from the wallet.  I propose a more typical (like electrum or bitcoin core) style formatting of the numbers:
![image](https://user-images.githubusercontent.com/60378539/187044376-75065692-6ef5-4190-bc5c-2f64e5fc78f5.png)

1. Self-transfers should show the fees lost.  
2. added amounts should be in default color (optional is a +)
3. sent/subtracted amounts should be red to be consistent with any sensible book-keeping formatting

I think this is extremely important for readability. @k9ert , @moneymanolis  What do you think?

moneymanolis:
Shouldn't sends also include the "fees lost"? @relativisticelectron

k9ert:
Ok, i've started to work on this. For now i will ommit the coloring but there is enough left:
* The amounts are sometimes not properly calculated. At least if you have txs which have inputs not coming from the wallet
* The amounts should not only be negative for outlow, they also should have a different meaning in the utxo-view: There, you're only interested in the `utxo_amount` of the utxos rather than the `flow_amount`. Those two terms will get new keys in tx-jsons and also will be used in the header of the table.

k9ert:
For the reference: https://github.com/cryptoadvance/specter-desktop/releases/tag/v1.14.5

PR #1901, opened 2022-09-24
UIUX/Chore: Amount & unit formatting, refactoring of inline-event handlers and introduction of JS unit tests

relativisticelectron:
This PR refactors all (BTC) amount & unit display (formatting started in https://github.com/cryptoadvance/specter-desktop/pull/1841) into a central set of javascript functions and applies it everywhere. This will also support the consistent formatting for https://github.com/cryptoadvance/specter-desktop/issues/1861. 

- This PR greatly simplifies formatting amounts, units, and prices with the functions:  `formatBtcAmountAndUnit`, `formatBtcAmount`, `formatPrice`
- Also applies the same `--cmap-bg-address-is-mine` color (now a tiny bit darker)  to the tx details inputs/outputs, as in the recipient box.  Previous yellow color was ambiguous.

**With this PR:**

image
image
image
image

**Needs lots of testing:**
- [x] Liquid Lsat/LBTC
- [x] Liquid assets.... I only have a tractor..... , anyone wants to send me some other test-Liquid asset?
- [x] confidential Liquid inputs/outputs
- [x] hide sensitive details
- [x] extend balance cypress test, remove pytest
- [x] remove `numberWithCommas`
- [ ] Cypress unit tests?  

**Out of scope**

- Same datastructure for assets, amounts, and assetlabels in `/wallet/<wallet_alias>/addressinfo/`  and https://github.com/relativisticelectron/specter-desktop/blob/677c971bf70393f064ac048bd8970bbea1131c56/src/cryptoadvance/specter/templates/includes/tx-row.html#L254


src/cryptoadvance/specter/templates/includes/helpers.jinja
```diff
@@ -29,41 +192,85 @@
...
+                        symbol='{{ specter.alt_symbol }}', 
+                        price='{{ specter.alt_rate }}',
+                        wrapInSpan=true,
+                        ){        
+        {% if not specter.price_check %}
+            return ''
+        {% endif %}
+
+
+        if ('{{ specter.hide_sensitive_info }}' == 'True'){
```
moneymanolis, line 251: This line is causing in Firefox (not in Chromium, though ...):

grafik

See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Stmt_after_return

Probably caused by the structure of the semi-colons. Ca

We should really start to use a JS linter ...

relativisticelectron: Should be solved in [80987a4](https://github.com/cryptoadvance/specter-desktop/pull/1901/commits/80987a4c5f75342186b50b78d6719f02c0fe1087)

moneymanolis: Looks cleaner, I'd say!

src/cryptoadvance/specter/templates/includes/helpers.jinja
```diff
@@ -1,26 +1,189 @@
...
+                e.g. 3 123 456, and color the blocks. Defaults to false.
+        Returns:
+            str: The formatted btc amount as html code.
+        */
+    function internalFormatBtcAmountAsSats(
+        value,
+        enableDigitFormatting=false,
+    ) {
+        const locale = 'en-US';
+        if (isNaN(parseFloat(value))){
```
moneymanolis, line 20: Not sure I understand the purpose of this if clause, what do you want to achieve here?

relativisticelectron: If it is NaN, it should stop trying to process the number.

src/cryptoadvance/specter/templates/includes/helpers.jinja
```diff
@@ -1,26 +1,189 @@
...
+            return parseFloat(value)        
+        }
+        var s = (parseFloat(value) * 1e8).toLocaleString(locale, { 
+            minimumFractionDigits: 0,
```
moneymanolis, line 24: What's the purpose of the options? `minimumFractionDigits` and `maximumFractionDigits`?

src/cryptoadvance/specter/templates/includes/helpers.jinja
```diff
@@ -1,26 +1,189 @@
...
+        const locale = 'en-US';
+        if (isNaN(parseFloat(value))){
+            return parseFloat(value)        
+        }
+        var s = (parseFloat(value) * 1e8).toLocaleString(locale, { 
+            minimumFractionDigits: 0, 
+            maximumFractionDigits: 0 
+        });         
+
+        let thousandsSeparator = Number(1000).toLocaleString(locale).charAt(1);
```
moneymanolis, line 28:
```suggestion
        let thousandsSeparator = ',';
```

src/cryptoadvance/specter/templates/includes/helpers.jinja
```diff
@@ -1,26 +1,189 @@
...
+        let decimalSeparator = Number(1.1).toLocaleString(locale).charAt(1);
```
moneymanolis, line 29:
```suggestion
        let decimalSeparator = '.';
```

src/cryptoadvance/specter/templates/includes/helpers.jinja
relativisticelectron: no, I left if on purpose general, so in the future switching to international is easier.

src/cryptoadvance/specter/templates/includes/helpers.jinja
relativisticelectron: no, I left if on purpose general, so in the future switching to international is easier.

src/cryptoadvance/specter/templates/includes/helpers.jinja
relativisticelectron: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString   How many digits are there  after the decimal separator.

src/cryptoadvance/specter/templates/includes/helpers.jinja
```diff
@@ -1,26 +1,189 @@
...
+        
+        // combine the thousandsSeparator with the left number to an array
+        var array = [];
+        for (var i in s){
+            var letter = s[i];
+            if (letter == thousandsSeparator){
+                array[array.length-1] += letter;
```
moneymanolis, line 36:
```suggestion
                array[array.length-1] += letter; // Using i would result in offset by 1 mistake
```

relativisticelectron: Could you elaborate this comment? I am unsure what it is trying to say.

src/cryptoadvance/specter/templates/includes/helpers.jinja
moneymanolis: OK, I see, but wouldn't it be better to return an actual error then (instead of NaN basically)?

src/cryptoadvance/specter/templates/includes/helpers.jinja
moneymanolis: OK, I'll leave a comment then for our future selves.

src/cryptoadvance/specter/templates/includes/helpers.jinja
moneymanolis: I am generally trying with these helper functions to make them as simple as possible to improve maintainability and collaboration with other contributors, and here I initially thought I could just replace `[array.length-1] `with `i`, but since `i` is here stemming from the string `s`, `i` as index is off by one (actually only for the second thousand seperator). If it confuses even more, we can drop the comment.

src/cryptoadvance/specter/templates/includes/helpers.jinja
moneymanolis: Got it. Probably the defaults are 0 anyways? But, isn't bad to have it explicit here.

src/cryptoadvance/specter/templates/includes/helpers.jinja
relativisticelectron: Good point. I added comments in [1ac96e6](https://github.com/cryptoadvance/specter-desktop/pull/1901/commits/1ac96e6473ac762c65c65ddb9d98e304715002ea)

src/cryptoadvance/specter/templates/includes/helpers.jinja
relativisticelectron: I added comments in [1ac96e6](https://github.com/cryptoadvance/specter-desktop/pull/1901/commits/1ac96e6473ac762c65c65ddb9d98e304715002ea)

src/cryptoadvance/specter/templates/includes/helpers.jinja
relativisticelectron: I added comments in [1ac96e6](https://github.com/cryptoadvance/specter-desktop/pull/1901/commits/1ac96e6473ac762c65c65ddb9d98e304715002ea)

src/cryptoadvance/specter/templates/includes/helpers.jinja
```diff
@@ -1,26 +1,198 @@
...
+        unit = unit.toUpperCase();
+        return ([null, "", "LBTC", "BTC", "TBTC", "TLBTC"].indexOf(unit) > -1)
+    }
+
+
+
+
+    // Formats the network independent units to network dependent units,
+    // e.g.: btc -> tBTC,  sat --> tsat ,  btc --> tLBTC, sat --> tLsat 
+    function formatUnitLabel(targetUnit='{{ "BTC" if specter.is_liquid else specter.unit }}'){
```
moneymanolis, line 167: `targetUnit={{ "BTC" if specter.is_liquid else specter.unit }}` is defined multiple times here, would improve readability IMO to only set `targetUnit` in the outer functions: `formatBtcAmountAndUnit` and `formatBtcAmount`. Also what is the point of` "BTC" if specter.is_liquid`? Just check the `specter.unit` with Liquid is the same as with First Layer Bitcoin.

relativisticelectron: Thanks for addressing this.  My current approach is:

- If it is not liquid, then this function will convert the number to 'sat' or keep 'BTC' depending on `specter.unit`
- if it is liquid, then targetUnit="BTC"  will prevent any conversion to sat, regardless of `specter.unit`. I currently do not know how to generate such a unit label ``<asset-label data-asset="${Object.keys(this.addressData.assets)[0]}"></asset-label>`` in `formatUnitLabel`.

But I agree the current structure is not self-explanatory. I will put some thought in to handle the liquid case better.

src/cryptoadvance/specter/templates/includes/helpers.jinja
```diff
@@ -1,26 +1,198 @@
...
+            }
+            break
+        }
+
+        if (countDigitsThatCanBeStripped >= minimumDigitsToStrip){
+            // loop through the float number, e.g. 0.03 000 000, from the right and replace 0's or the '.' until you hit anything != 0
+            for (var j in array){
+                var i = array.length - j - 1;
+                if ((array[i] == "0") && (array.length - i <= maximumDigitsToStrip)){
+                    array[i] = `<span class="unselectable transparent-text ${hideStripped ? 'hidden' : ''}">${array[i]}</span>`;
```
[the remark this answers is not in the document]
relativisticelectron: Correct. Sometimes (like the wallet Balance), it is necessary to keep the stripped 0 horizontal spaces, to be able to right-align them.  If it is in-line and the unit is directly placed behind it, then there should be no horizontal space.

src/cryptoadvance/specter/templates/includes/helpers.jinja
relativisticelectron: > Perhaps it would be best to separate Liquid in completely own functions.

Yeah...... I think that would be cleanest...     I think it is best you pause the review until I separated liquid out.

src/cryptoadvance/specter/templates/includes/helpers.jinja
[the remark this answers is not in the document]
moneymanolis: OK, then all good! Really nice work.

src/cryptoadvance/specter/templates/includes/helpers.jinja
moneymanolis: Separation is always good. I will work in parallel to figure out how we best test all this. I was thinking to use Cypress to do some JS unit tests there similar to what you did in the Notifications PR:
grafik

relativisticelectron: > I will work in parallel to figure out how we best test all this.

thanks!  I have extended the cypress test, but a cypress unit test would be better.

I separated the Liquid part in own functions. I think everything works and is ready for review.

src/cryptoadvance/specter/templates/includes/helpers.jinja
relativisticelectron: Yes, you are right.  Done in  ab44af63  . And I now notice that I actually don't use this case...

cypress/integration/spec_balances_amounts.js
```diff
@@ -63,33 +63,171 @@ describe('Test the rendering of balances and amounts', () => {
...
+        // get the ghot receive address to the clipboard
+        // cy.selectWallet('Ghost wallet')
+        // cy.selectWallet('Ghost wallet')       // Once again because only once doesn't work for some stupid unknown reason
+        // cy.get('#btn_receive').click()
+        // cy.get('#address').click()  // get receive address from ghost wallet 
+        cy.selectWallet('Funding wallet')
+        cy.selectWallet('Funding wallet')       // Once again because only once doesn't work for some stupid unknown reason
+        cy.get('#btn_send').click()
+        // check that the amount_available for sending is displayed correct
+        // cy.get('#wallet-amount_available').should('have.text', '20.00000000 tBTC')
```
skwp, line 117: are these tests relevant? kill commented code or make em pass?

src/cryptoadvance/specter/server_endpoints/wallets_api.py
```diff
@@ -494,6 +495,17 @@ def addresses_list(wallet_alias):
 
     addresses_list = wallet.addresses_info(address_type == "change")
 
+    # add the asset_label
```
skwp, line 498: consider extracting complex code into self descriptive function

`add_asset_label`

src/cryptoadvance/specter/templates/includes/address-data.html
```diff
@@ -108,6 +110,16 @@ <h1>{{ _("Address details") }}</h1><br>
                     let walletName = jsonResponse.walletName;
                     let address = jsonResponse.address;
                     let walletLink = `{{ url_for('wallets_endpoint.wallet', wallet_alias='WALLET_ALIAS') }}`.replace("WALLET_ALIAS", this.wallet);
+
+                    var formattedAmountAndUnit = formatBtcAmountAndUnit(this.amount);
+                    {% if specter.is_liquid %}
+                        var labels = [];
```
skwp, line 116: code in html template - is it possible to extract to helper?

src/cryptoadvance/specter/templates/includes/tx-row.html
```diff
@@ -244,40 +244,16 @@
...
-                            if (unique_assets.length == 1){
-                                this.amountText.innerHTML += ` <asset-label data-asset="${this.tx.asset[0]}" data-label="${this.tx.assetlabel[0]}"></asset-label>`;
-                            }else{
-                                this.amountText.innerHTML = `${unique_assets.length} assets`;
-                            }
-                        }else{
-                            this.amountText.innerHTML += ` <asset-label data-asset="${this.tx.asset}" data-label="${this.tx.assetlabel}"></asset-label>`;
-                        }
-                    }
-                }
```
skwp, line 280: very nice cleanup

src/cryptoadvance/specter/templates/includes/address-data.html
relativisticelectron: Unfortunately the structure of `this.assets` and `this.tx` in https://github.com/relativisticelectron/specter-desktop/blob/677c971bf70393f064ac048bd8970bbea1131c56/src/cryptoadvance/specter/templates/includes/tx-row.html#L254 is different. I chose therefore a generic approach: Do not assume a specific format and simply pass arrays (or single values).

cypress/integration/spec_balances_amounts.js
relativisticelectron: Done in  42ae0a2e

moneymanolis:
Resulting from my efforts to include JS unit tests for this PR (you can't do this if there are jinja tags within the file) and also based on architectural input from @k9ert: Ideally, there should not be any jinja tags used in `helpers.jinja` (of course then we would also rename it). I went through the used jinja tags: Why can't we replace those tags with JS parameters that are set with jinja variables at the respective jinja files, say, at `total_wallet_balances.jinja`?

You don't have to do this (I am working on this with), but I am just mentioning this context here: Ideally, we'd (1) use smaller JS helper modules and (2) also in order to test them in Cypress, they have to be exportable modules. As of now, for example, `helpers.js` is imported in `base.jinja `with `<script type="text/javascript">` which disables exports of functions etc. from this module. If we just change the type to `module` in `base.jinja` some JS functions used in `base.jinja `are not found since this import type is always defaulting to a deferred import. For more background, see: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script

moneymanolis:
@relativisticelectron I made a PR to your branch [here](https://github.com/relativisticelectron/specter-desktop/pull/4). The idea would be to include the formatting JS functions in this PR in the helper-module `formatting.js` from my PR and then test them in `spec_formatting.js` in Cypress (also from my PR). I am happy to do the unit tests in Cypress for you once the formatting functions are consolidated in `formatting.js` to assist you since we are already putting a lot of requirements on you in this PR review. Please note, there can't be any jinja tags in `formatting.js`otherwise the tests won't run. Let me know if you need help to "get rid" of jinja tags.

relativisticelectron:
@moneymanolis I moved everything to `formatting.js`. The functions are available in tx-table, but not in the price_bar.jinja ...  Currently I am stuck why `window.some_function` doesn't always work

k9ert:
Anything missing for this? @moneymanolis 
I think we want to make a bugfix release on monday and we can then hopefully merge this on tuesday.
However there are still cypress issues. Looks like some includes didn't work or so?
```
  <testsuite name="Test sending transactions" timestamp="2022-10-12T19:09:38" tests="7" time="1.4540" failures="6">
    <testcase name="Test sending transactions Send a standard transaction" time="0.0000" classname="Send a standard transaction">
      <failure message="The following error originated from your application code, not from Cypress. It was caused by an unhandled promise rejection.
  &gt; other_recipient is not defined
When Cypress detects uncaught errors originating from your application it will automatically fail the current test.
This behavior is configurable, and you can choose to turn this off by listening to the `uncaught:exception` event.
https://on.cypress.io/uncaught-exception-from-application" type="ReferenceError"><![CDATA[ReferenceError: The following error originated from your application code, not from Cypress. It was caused by an unhandled promise rejection.
  > other_recipient is not defined
When Cypress detects uncaught errors originating from your application it will automatically fail the current test.
This behavior is configurable, and you can choose to turn this off by listening to the `uncaught:exception` event.
https://on.cypress.io/uncaught-exception-from-application
    at setMaxAmount (http://localhost:25444/wallets/wallet/test_hot_wallet_1/send/new:5917:21)
    at RecipientBox.<anonymous> (http://localhost:25444/wallets/wallet/test_hot_wallet_1/send/new:5811:5)
    at RecipientBox.dispatchSendMax (http://localhost:25444/wallets/wallet/test_hot_wallet_1/send/new:2650:9)
    at HTMLSpanElement.<anonymous> (http://localhost:25444/wallets/wallet/test_hot_wallet_1/send/new:2611:10)]]></failure>
    </testcase>
    <testcase name="Test sending transactions Open up transaction details" time="0.0000" classname="Open up transaction details">
      <failure message="Timed out retrying after 4000ms: Expected to find content: &apos;Value: 20.00000000 tBTC&apos; within the element: &lt;p.tx_info&gt; but never did." type="AssertionError"><![CDATA[AssertionError: Timed out retrying after 4000ms: Expected to find content: 'Value: 20.00000000 tBTC' within the element: <p.tx_info> but never did.
    at Context.eval (http://localhost:25444/__cypress/tests?p=cypress/integration/spec_wallet_send.js:145:52)]]></failure>
    </testcase>
    <testcase name="Test sending transactions Adding and deleting recipients" time="0.0000" classname="Adding and deleting recipients">
      <failure message="The following error originated from your application code, not from Cypress.
  &gt; addRecipient is not defined
```

relativisticelectron:
> However there are still cypress issues. Looks like some includes didn't work or so?

It appears that in-html javascript references like `onclick="addRecipient('', '', 'btc', '')">` are not compatible with the module `<script type="module">` .  This is easily fixable, by defining this `onclick` event in javascript, like
```
document.getElementById('create_psbt_btn').addEventListener("click", function() {
			validateForm(true);
		})
```

moneymanolis:
@relativisticelectron thanks for all the work.

I made a little example with the capitalize function to implement a suggestion by @joernroeder to use a central Specter object that holds (via "sub-objects") the JS helper functions in[ this commit](https://github.com/cryptoadvance/specter-desktop/pull/1901/commits/abab4330d385dc8e89448d0897b0489d8dfc2aea)

Advantages:

- Easy import for unit tests, just` import { Specter } from '../../src/cryptoadvance/specter/static/helpers.js'`
- Better readability: `titleButton.innerHTML = Specter.format.capitalize(display_text)`
- With `attachToSpecter()` we can just loop over all imports to add them `helpers.js` - speaking of which, we can simplify here by importing all files, will do this as well.

What do you guys think? @k9ert @joernroeader @relativisticelectron 

I can then go ahead and implement this. 
After that, I can write the unit tests.

relativisticelectron:
> What do you guys think? @k9ert @joernroeder @relativisticelectron

Sounds good.
 
> I can then go ahead and implement this. After that, I can write the unit tests.

Thanks.

> EDIT: I am unsure if `<script type="module">` is OK for webcomponents, but I guess it works.... Almost all `<script >` statements needs to be changed to `<script type="module">` , because almost everywhere the js code from modules is needed. All the changes to "module" are tedious... I'm on it.

The issue I am referring to here is if a function call like
```
<script>
functionOfModule()
</script>
```
will fail because `<script>` is executed before `<script type="module">`. And since now almost all `<script>` reference (directly or indirectly) call functions in a  `<script type="module">`, we have to change all the events as described in https://github.com/cryptoadvance/specter-desktop/pull/1901#issuecomment-1279124390

moneymanolis:
> we have to change all the events (e.g. onclick) as described in https://github.com/cryptoadvance/specter-desktop/pull/1901#issuecomment-1279124390

Inline event handlers are an anti-pattern in JS anyways I would say. See here for example: https://www.htmlgoodies.com/javascript/working-with-inline-event-handlers/

To my knowledge, they were largely introduced by feature hacker contributors. 

So, this is an important refactoring. Additional work of course, but I'd say it is worth it.

moneymanolis:
@relativisticelectron I'd suggest you pause until I am done with refactoring. When we use:
`Specter.format.helperFunctionName` it would be nice to not have the 'format' twice. E.g.
`Specter.format.BtcAmountAndUnit ()` and not `Specter.format.formatBtcAmountAndUnit()`
What do you think?

moneymanolis:
In https://github.com/cryptoadvance/specter-desktop/pull/1901/commits/16e638934e11abe0b6a6e5c2cf06f1f51531670a I've finalised using a central Specter object, there is now:
```
Specter.format
Specter.common
Specter.mobile
```

All functions from `format.js `are refactored.
From `common.js` these are open:
```
 copyText
 showNotification
```
They are also used with inline event handlers, and I didn't want to create one huge merge conflict by touching this in parallel, rather waiting here, until @relativisticelectron is finished.

And from `mobile.js`:
 `toggleMobileNav`
Same as above.

I've also started with the unit tests in `spec_format.js`. @relativisticelectron I think the best will be if you start writing little tests there, you will then easily see when you need to finetune your functions.

For example, you will see that if you want to test `unitLabel` that you need the following parametrisation in order to test them:
`function unitLabel(targetUnit=specter_unit, enableHTML=true, is_liquid=false, is_testnet=false)`

I also don't think it is good idea to set those global variables in `base.jinja`:
```
		<script>
			/* setting global js variables for helpers.js */
			const is_liquid = {% if specter.is_liquid %} true  {%else%}  false {%endif%};
			const is_testnet = {% if specter.is_testnet %} true  {%else%}  false {%endif%};
			const price_check_enabled = {% if specter.price_check %} true  {%else%}  false {%endif%};
			const hide_sensitive_info_enabled = {% if specter.hide_sensitive_info %} true  {%else%}  false {%endif%};
			const specter_unit = '{{ specter.unit }}';
			const alt_rate = '{{ specter.alt_rate }}';
			const alt_symbol = '{{ specter.alt_symbol }}';
		</script>
```
I'd prefer it if the individual jinja templates set the relevant arguments for the JS functions.

We will definitely need - besides a lot of tests (unit and integration tests) a bit more eyes on this PR since we are touching quite a lot of stuff.

moneymanolis:
@relativisticelectron circling back to my latest comment, there are two issues IMO:
(1) Parametrisation of the functions:

`function unitLabel(targetUnit=specter_unit, enableHTML=true, is_liquid, is_testnet) `
instead of just
`function unitLabel(targetUnit=specter_unit, enableHTML=true)`

(2) How are arguments passed to those parameters. Via global variables etc.?

moneymanolis:
Based on input from @joernroeder, I think it would be best to also attach those variables below in the `base.jinja` to the global Specter object:

		<script>
			/* setting global js variables for helpers.js */
			const is_liquid = {% if specter.is_liquid %} true  {%else%}  false {%endif%};
			const is_testnet = {% if specter.is_testnet %} true  {%else%}  false {%endif%};
			const price_check_enabled = {% if specter.price_check %} true  {%else%}  false {%endif%};
			const hide_sensitive_info_enabled = {% if specter.hide_sensitive_info %} true  {%else%}  false {%endif%};
			const specter_unit = '{{ specter.unit }}';
			const alt_rate = '{{ specter.alt_rate }}';
			const alt_symbol = '{{ specter.alt_symbol }}';
		</script>

like so: 
```
// create or extend specter object
window.Specter = window.Specter || {}
window.Specter.is_liquid =  {% if specter.is_liquid %} true  {%else%}  false {%endif%};
```

I will work on that on Monday @relativisticelectron, don't worry about this part.

moneymanolis:
@relativisticelectron I've done the first test for unitLabel in https://github.com/cryptoadvance/specter-desktop/pull/1901/commits/fd047fe375eef862b12289d16395af79d988f4e3

and will continue with the remaining ones. 

I think I have figured out how to do the tests exactly in Cypress. I had some issue but then found this: https://docs.cypress.io/api/commands/window#Cypress-uses-2-different-windows - which helped a lot.

src/cryptoadvance/specter/static/helper-modules/format.js
```diff
@@ -0,0 +1,340 @@
...
+					unit='BTC', 
+					symbol=Specter.altSymbol, 
+					price=Specter.altRate,
+					wrapInSpan=true,
+					){        
+	if (!Specter.priceCheck){
+		return '';}
+	else {		
+		if (Specter.hideSensitiveInfo){
+			return ["#########"];}
```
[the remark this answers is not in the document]
relativisticelectron: Thanks for noticing. Fixed in  4aa58589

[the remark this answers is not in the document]
relativisticelectron: Thanks for noticing. That was a mistake. Fixed in  4aa58589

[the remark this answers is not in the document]
moneymanolis: No probs, that is what the tests are for ;-)

moneymanolis:
TODO
- [ ] Liquid unit test
- [ ] `copyText` and `toggleMobileNav` refactor
- [ ] Check that HWI flow works, global JS functions are used there heavily.
- [ ] Testing, testing, testing ...

relativisticelectron:
@moneymanolis I have done all of the "onsomething" replacements.   When you have time, I would be glad about any help getting the cypress tests working again.

moneymanolis:
Great, thanks a bunch and of course I will help you with that. But, will only get to it after the next release. A lot to do regarding the Spectrum addition ...

moneymanolis:
Closing this for now. The formatting of amounts need to picked up again based on the new UIUX (Specter v2). @otkstdio @relativisticelectron you two should get together and find a way to incorporate most of this in the new design.

PR #1904, opened 2022-09-26
Bugfix: Hover effect in balance display

moneymanolis:
There was a background effect when hovering over the balance display:

grafik

Solved by removing unnecessary hover style.

### DIFF
--- a/src/cryptoadvance/specter/static/styles.css
+++ b/src/cryptoadvance/specter/static/styles.css
@@ -749,10 +749,6 @@ th{
     font-weight: inherit;
     border-bottom: 3px solid var(--cmap-bg-lightest);
 }
-tbody tr:not(.nohover):hover{
-    background: rgba(255,255,255,0.03);
-    color: #fff;
-}
 .xpub{
     max-width: 300px;
 }


relativisticelectron:
This change lets the hover effect in  `devices/device/..../` vanish.  However I think it would be way better to explicitly add this hover effect back in, than leaving it implicit (current master).

src/cryptoadvance/specter/templates/base.jinja
```diff
@@ -38,6 +38,9 @@
 				border-right: 0;
 				{% endif %}
 			}
+			tr:hover {
+			background: rgba(255,255,255,0.03);
```
relativisticelectron, line 42: can you indent the inner part?

relativisticelectron (approved):
LGTM

src/cryptoadvance/specter/templates/base.jinja
```diff
@@ -38,6 +38,9 @@
 				border-right: 0;
 				{% endif %}
 			}
+			tr:hover {
+				background: rgba(255,255,255,0.03);
```
relativisticelectron, line 42: This has introduced the hovering effect back to the balance display.

relativisticelectron (changes requested):
The `tr` in the balance display and in the tx table need differentiation.   You could leave `tr` without the hover and add a hover-css-class.

src/cryptoadvance/specter/templates/base.jinja
moneymanolis: Yeah, too general in` base.jinja`, indeed. Will add a hover-css-class for the respective rows.

moneymanolis:
@relativisticelectron anything else spotted by you? Otherwise I'd merge.

relativisticelectron (approved):
works well.

moneymanolis:
Thanks for the review, I appreciate the attention to details!

PR #1907, opened 2022-10-03
Bugfix: Service Manager loads wrong service Test-config

relativisticelectron:
Since `CypressTestConfig` and `TestConfig` end both with `TestConfig`, the service manager had unreliable loading of the config.   This wrong loading was visible [here](https://cirrus-ci.com/task/4520334396227584?logs=test#L245) 
image

Maybe the intention was to the the `TestConfig` also implicitly for cypress. However creating an explicit `CypressTestConfig` is better.


k9ert:
Good catch!
A couple of lines later, the Loading mechanism checks via inheritance, whether another config is better suited if no "exact match" (now exact match, but before "endswith-match") has been found. Can you modify that as well with exact match?
image

k9ert (approved):

Issue #1834, opened 2022-08-03
UI: Unclear explanation to use the txlist search 


relativisticelectron:
The instructions in the tooltip for searching for the date and amount are not clear to me (see below)
![errors](https://user-images.githubusercontent.com/60378539/182673949-f1a26105-dccc-43d1-9f23-31a2792438e6.gif)

**Issues:**

- The date for searching requires 0's
- The date format for searching is not the same as shown in the table
- the instruction for searching suggests one could search for amounts "lower" than 1 BTC, by entering "< 1 BTC".

moneymanolis:
All good points! Definitely suboptimal. So much to improve ... Can you work on it?

PR #1912, opened 2022-10-05
UI Bugfix:  Remove black empty bar in tx-table after searching  with no results

relativisticelectron:
In https://github.com/cryptoadvance/specter-desktop/issues/1834 there was a blue bar appearing in tx-table 
image

This PR fixes that.

EDIT:  The visible part is a black empty bar.  But if I understand correctly, it is caused by the small blue-background in the beginning of the empty row.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -5,7 +5,10 @@ addopts = --bitcoind-version v22.0.0 --elementsd-version v0.21.0.2
 markers =
     slow: mark test as slow.
     elm: mark test as elementsd dependent
-#log_cli = 1
+# If you need live logging to debug, uncomment the next line
+# log_cli = True 
+# Then set the desired logging level on the command line, for example:
+# pytest --log-cli-level INFO
 
 filterwarnings =
     ignore::DeprecationWarning:bitbox02[.*]
--- a/utils/test-cypress.sh
+++ b/utils/test-cypress.sh
@@ -20,13 +20,24 @@ fi
 
 function check_consistency {
   if ! npm version 2> /dev/null 1>&2 ; then
-    echo "npm is not on the PATH. Please install node and bring on the PATH"
+    echo "npm is not on the PATH. Please install node and bring it on the PATH."
     exit 1
   fi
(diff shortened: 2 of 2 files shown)


src/cryptoadvance/specter/templates/includes/tx-table.html
```diff
@@ -1005,6 +1005,7 @@ <h1>{{ _("Export transactions to CSV") }}</h1>
                     this.pageCount = 1;
                 } else {
                     this.tableHeader.classList.remove('hidden');
+                    this.feedbackRow.classList.remove('feedback-row');
```
moneymanolis, line 1008: Not sure how this relates to the blue bar, also not necessary for the feedbackRow, we are adding `hidden` here already.

moneymanolis:
Not sure, I am following, if you want to get rid of the focus blue highlight, you'd need that:
```
        .search:focus {
            outline: none;
        }
```
See also here: https://stackoverflow.com/questions/7474147/input-field-blue-focus-highlight-impossible-to-get-rid-of

Let me know whether you wanted that and I can push that.

src/cryptoadvance/specter/templates/includes/tx-table.html
relativisticelectron: Well, I noticed that the page originally starts from `class="hidden"` and then later `class='feedback-row'` is added and not removed again. It does cause the blue bar even though "hidden" is added.

relativisticelectron:
I think the blue surrounding bar is OK, and consistent with other input fields.

moneymanolis:
Wait, I think I understand now, you not referring to the blue focus outline but to this:

grafik

moneymanolis:
During testing I came across this, will create an issue out of it. Since there are so many issues with the search bar, I will create a milestone for this, so we can gather all related issues with the milestone tag.

The feedback in the tx table search is misleading / incorrect when there are no hits. The feedback shown is intended for a fresh wallet or a wallet without rescan.

grafik

PR #1914, opened 2022-10-07
Chore: Removing the Singleton

k9ert:
The Singleton idea never had the value i thought it would have. Let's get rid of it.
@kdmukai i know i was my idea, but it was a bad idea nevertheless :sweat_smile:

PR #1918, opened 2022-10-10
Chore: Make data encrytion for services optional

relativisticelectron:
This PR   builds on https://github.com/cryptoadvance/specter-desktop/pull/1914 and implements the optional data encryption for services necessary for https://github.com/cryptoadvance/specter-desktop/pull/1766#issuecomment-1259183230 

A more useful diff until https://github.com/cryptoadvance/specter-desktop/pull/1914  is merged   is the [diff](https://github.com/k9ert/specter-desktop/compare/remove_singleton...relativisticelectron:specter-desktop:20221010_optional_encrytion_for_services?expand=1)


src/cryptoadvance/specter/services/service.py
```diff
@@ -47,30 +48,34 @@ def __init__(self, active, specter):
         self.active = active
         self.specter = specter
 
+    @classmethod
+    def _storage_manager(cls):
+        return (
+            app.specter.service_encrypted_storage_manager
+            if cls.encrypt_data
+            else app.specter.service_unencrypted_storage_manager
```
relativisticelectron, line 56: I don't like that I have to access here the `app`. A `storage_manager` should be given at initialization, however, since these are not instances, but class methods, I do not know how to make it nice.

k9ert: What's wrong with that? The app-object is also used in the controller and in a lot of other use-cases. The only two places where you can't access the app is in the `__init__()` method of the service-class and the `callback_after_serverpy_init_app` of the service. 

It's very complicated to pass it in the `__init__()` method as this is in the middle of the initialisation process. It would be high likely that you create hen/egg issues. Therefore usually extension-frameworks like ours usually ask the developer of an extension to do the setup in a callback which is later in the lifecycle.

The latest one in our case is `callback_after_serverpy_init_app`. We can't access the app here either, because the flask_object is not yet fully instantiated.That's definitely odd here as this method is the place where an extension is suppose to setup all it needs to start working properly.

So I'd propose to pass both `storageManager`s in named parameters here: https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/server.py#L261

Would that work for you?

relativisticelectron: If I set do
```
    def callback_after_serverpy_init_app(self, scheduler: APScheduler, storage_manager:ServiceEncryptedStorageManager):
        """
        Args:
            scheduler (APScheduler): _description_
            storage_manager (ServiceEncryptedStorageManager):  can be ServiceEncryptedStorageManager or a derived class
        """
        self.storage_manager = storage_manager

```

in the Service class,  it won't be available in https://github.com/cryptoadvance/specter-desktop/blob/6a8ceb391b949cd672259ee541586b2ef84aafb2/src/cryptoadvance/specter/services/service.py#L57 , right?

k9ert: As explained, it's not a good idea to do setup-stuff in the `__init__` method. 
Well the `ServiceEncryptedStorageManager.get_instance()` method would no longer be available after merging #1914. 
But doing stuff in class-methods is exactly what we should imho avoid here.
So where, apart from the service-methods do you want to use the `storageManager` ?

k9ert: As a summary: Those class-based methods in the Service-class need to be replaced with normal instance-methods.

relativisticelectron: > As a summary: Those class-based methods in the Service-class need to be replaced with normal instance-methods.

Yes, the `classmethod`s should be removed. However this is out of scope of this PR.  (Naively attempting it https://github.com/cryptoadvance/specter-desktop/pull/1918/commits/7b1859d3f64fa7df4d4005616e9dde6a30fadbdf  breaks the Swan service completely).

relativisticelectron:
[Sometimes](https://github.com/cryptoadvance/specter-desktop/runs/9035764296) pytest in CI also gives strange errors.... and a [commit later](https://github.com/cryptoadvance/specter-desktop/pull/1918/commits/1d23446a0aa18c17beaec1cfc4e25c301e6b5364) the errors are gone...

k9ert:
> By the way: How to test the SwanService?

I've tested the Swan Authorisation flow and it works!

k9ert (approved):

PR #1919, opened 2022-10-11
Which address and wallet types supported?

moritzwietersheim:
Which address and wallet types does Specter Desktop support exactly?

I understand that this is only nested Segwit and native Segwit. No Legacy addresses.  But P2SH is supported? ( Context: https://shiftcrypto.ch/blog/what-are-bitcoin-address-types/ ) 

@moneymanolis - could we please clarify this with 2 sentences in the FAQ. Thank you!


moneymanolis:
Moved to an issue.

Issue #1935, opened 2022-10-21
Bug: Repeated `python3 setup.py install` leads to errors in services


relativisticelectron:
When running
```
pip install -e .  && python3 setup.py install 
```
for the first time the python folder  `python3.10/site-packages/` contains
![image](https://user-images.githubusercontent.com/60378539/197263281-4ec9fe18-5d10-4930-abb8-903dead8f48b.png)
and `python3.10/site-packages/cryptoadvance/specterext/` contains
![image](https://user-images.githubusercontent.com/60378539/197263542-dcf5a87b-f281-4e37-b10f-6294597bafb3.png)

However running 
```
pip install -e .  && python3 setup.py install 
```
again , now 
-  `python3.10/site-packages/cryptoadvance/specterext/` contains
![image](https://user-images.githubusercontent.com/60378539/197263723-96377ae3-e5bc-4280-b7d5-bffb29e7e083.png)
which then raises the exception at specter start;
```
cryptoadvance.specter.specter_error.SpecterError: 
                Module cryptoadvance.specterext.liquidissuer.service  could not be found. This could have these reasons:
```
-  `python3.10/site-packages/` remains contains the same folders.

k9ert:
This might be due to at least one of the extensions (or specter itself) having a `__init__.py` file in `./src/cryptoadvance`, or `./src/cryptoadvance/specterext/`. That should not be the case. The reason is an explicit design-decision i did back in #524 : `cryptoadvance` and later `cryptoadvance/specterext` should be namespace-packages exactly in order to make it possible to "merge" specifc subtrees together in one virtualenv.

Can you check that in your case? I guess i recently discovered something similiar and removed it but can't remember where.

k9ert:
Not sure why i couldn't reproduce in the first place but now i can :-|

relativisticelectron:
> I think that `python3 setup.py install` was never ok in the first place. IMHO it's installing specter-desktop as a regular package. If then, the second time, `pip install -e .` is realizing that the package is installed regulary and removes it (unfortunately not properly). We have to revisit why we're installing it in the first place.

Well, I am re-installing it after I changed something (and do not want to setup a completely new virtualenv):
https://github.com/cryptoadvance/specter-desktop/blob/a739a062ca44005a1613f072f8f99def3c71c6c0/docs/development.md?plain=1#L98

relativisticelectron:
[...]

PR #1939, opened 2022-10-24
Extension callbacks refactoring and middleware implementation

k9ert:

So this PR does those things:
- Implementing a new `return_style` called "middleware" and the old-style will be called `dict`:
  - `dict` will return a dict where the extension_id will be used as key and the returnvalue as a value (to be renamed to "collect")
image

On top of those changes, a new graphql-extension has been created:
```
{
  bookmarks {
    title
    desc    
    
  }
}
```
```
{
  users {
    username
    
  }
}
```


relativisticelectron:
When you refactor the locations of the services, can you also solve https://github.com/cryptoadvance/specter-desktop/issues/1935 ?

k9ert:
This has been closed and instead we have now #2105 and #2106

PR #1942, opened 2022-10-25
Fix: No encryption for devhel extension required

relativisticelectron:
In https://github.com/cryptoadvance/specter-desktop/pull/1918 the encryption of the service data was made optional.  This PR removes the requirement `user_secret_decrypted_required` for the devhelp extension. That way the settings are also accessible without authentication.


k9ert:
Would you mind to adjust the extension-generation process as part of this PR? It's not difficult and you would get insoghts there:
- Add an option to the click annotations like `--encrypted-userdata` here: https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/cli/cli_ext.py#L39 (including adding it as a param to the method signature)
- Add a section explaining what it means and that the extension should decide between encrypted/unencrypted here: https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/cli/cli_ext.py#L111-L112
- Add the variable to the `ExtGen` instantiation and there ...
- The service and the controller would needs some adjustments. They are rendered here: https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/cli/cli_ext.py#L150-L160 however, the code for the templates is living here:
  - https://github.com/cryptoadvance/specterext-dummy/blob/master/src/dummyorg/specterext/dummy/service.py#L12
  - https://github.com/cryptoadvance/specterext-dummy/blob/master/src/dummyorg/specterext/dummy/controller.py#L53
  
Mind the different templating syntax. Those are "metatemplates" as they are creating templates which themself are later templates.

You can test it via using `--tmpl-fs-source` and checkout the dummy-project somewhere. Also i've created a small section about https://docs.specter.solutions/desktop/extensions/intro/#virtualenv-management

k9ert:
One other thing here would be that the templates are becoming more dependent on the version of specter controlling the generation-process. It would be cool if we could get another cli-switch which would specify the version of specter, simply assuming that the dummy-repo has tags with that version (and maybe falling back to master if it doesn't).
This is already established in parts: https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/services/extension_gen.py#L41 but need to be parametrized and maybe the fallback-solution need to be implemented.

k9ert:
Copied the extension-generation wishlist here:
https://github.com/cryptoadvance/spectrum/issues/10

will test and merge soon.

k9ert (approved):

PR #1944, opened 2022-10-26
Chore: Tagging specterext-dummy

k9ert:
The Extension-generation process is dependent on the specterext-dummy template-repo:
https://github.com/cryptoadvance/specterext-dummy

So as soon as extgen code changes, there might be incompatibilities between both repos. So this PR creates a tag in that repo with the same version-number than the tag of the specter-desktop-release.

It needs two additional env-vars in the gitlab-CICD-var-section (https://gitlab.com/k9ert/specter-desktop/-/settings/ci_cd):
- SSH_SPECTEREXT_DEPLOY_KEY contains the secret part of an ssh-key which is configured as a deploy-key with write-access in the specterext-repo: https://github.com/cryptoadvance/specterext-dummy/settings/keys
- KNOWN_HOSTS is a copy of a `known_hosts` as it would be created accepting the fingerprint of the github-servers.


k9ert:
@relativisticelectron i've tagged you for reviewing but it's more like a "please notice". This is related to #1942

moneymanolis (approved):

Issue #1949, opened 2022-10-31
Incomplete deletion of wallet


moneymanolis:
The deletion of a wallet via the UI only deletes the json file in the Specter folder but not the directory on the Bitcoin Core side. The call at the endpoint is:

```
app.specter.wallet_manager.delete_wallet(wallet, app.specter.bitcoin_datadir, app.specter.chain)
```
where` app.specter.bitcoin_datadir` is returning `self.node.datadir` which looks like it is an empty string if the node in question was generated without autodect on so the datadir was probably set to "" through the form. 
It says in `node.py` about the the datadir attribute: "A directory where a bitcoin.conf can be found, **relevant for autodetect**" - this seems incomplete - since it is also crucial for finding and removing the Bitcoin Core data.

PR #1950, opened 2022-10-31
Bugfix: Incomplete wallet deletion

moneymanolis:
Fixes https://github.com/cryptoadvance/specter-desktop/issues/1949


src/cryptoadvance/specter/managers/wallet_manager.py
```diff
@@ -348,19 +348,24 @@ def create_wallet(self, name, sigs_required, key_type, keys, devices, **kwargs):
...
-                bitcoin_datadir = os.path.join(bitcoin_datadir, chain)
+                datadir = os.path.join(datadir, chain)
             candidates = [
-                os.path.join(bitcoin_datadir, wallet_rpc_path),
-                os.path.join(bitcoin_datadir, "wallets", wallet_rpc_path),
+                os.path.join(datadir, wallet_rpc_path),
+                os.path.join(datadir, "wallets", wallet_rpc_path),
             ]
             for path in candidates:
                 shutil.rmtree(path, ignore_errors=True)
```
k9ert, line 371: Please move the deletion logic inside the node. Can't we have a node-method called `delete_wallet` which will do whatever is necessary?!

moneymanolis: done in https://github.com/cryptoadvance/specter-desktop/pull/1950/commits/5b6280f60473750e578dc3d46e5319d605894984

moneymanolis:
Deletion of the Core wallet file on the node was moved to the node class. 
There was also a UI bug where the node datadir in autodetect was only set at the beginning via `get_default_datadir()`. Once the datadir was changed, the modified datadir was used and displayed in the UI. See the change in `node_settings.jinja`
There is also feedback in the UI now on whether the wallet files on the Core node could be deleted (only works if the node is on the machine as Specter). They represent very sensitive data after all:

#### Both deleted

grafik

#### Only the Specter wallet deleted

grafik

src/cryptoadvance/specter/node.py
```diff
@@ -530,6 +531,40 @@ def check_blockheight(self):
...
         return is_liquid(self.chain)
 
+    def delete_wallet_file(self, wallet) -> bool:
+        """Deleting the wallet file located on the node. This only works if the node is on the same machine as Specter.
+        Returns True if the wallet file could be deleted, otherwise returns False."""
+        wallet_rpc_path = os.path.join(
+            wallet.rpc_folder, wallet.alias
+        )  # e.g. specter/jade_wallet
+        logger.debug(f"The wallet_rpc_path is: {wallet_rpc_path}")
+        self.rpc.unloadwallet(wallet_rpc_path)
```
k9ert, line 541: The method doesn't sound like it's the responsibility of it to unload the wallet. So i assume it's doing that here just to be sure? What happens if it's already unloaded? So this is safe to call twice? Maybe make a comment here!

src/cryptoadvance/specter/node.py
```diff
@@ -530,6 +531,40 @@ def check_blockheight(self):
...
+            path = os.path.join(datadir, f"{self.chain}/wallets", wallet_rpc_path)
+        else:
+            path = os.path.join(datadir, wallet_rpc_path)
+        try:
+            shutil.rmtree(path, ignore_errors=False)
+            logger.debug(f"Removing wallet file at: {path}")
+            wallet_file_removed = True
+        except FileNotFoundError:
+            logger.debug(f"Could not find any wallet file at: {path}")
+            pass
```
k9ert, line 565: We don't need a `pass` here.

src/cryptoadvance/specter/node.py
```diff
@@ -620,13 +655,8 @@ def node_type(self):
         return "BTC"
 
     @property
-    def datadir(self):
```
k9ert: Removing the property here: Are you sure you checked for all references of this? Check e.g. `specter.py:289` and maybe in more occasions?!

src/cryptoadvance/specter/wallet.py
```diff
@@ -1562,6 +1562,10 @@ def get_address_name(self, address, addr_idx):
         # TODO: remove
         return self.getlabel(address)
 
+    @property
+    def rpc_folder(self):
```
k9ert, line 1566: please add docstring

k9ert (approved):
Apart from the changes mentioned, approved!

src/cryptoadvance/specter/wallet.py
moneymanolis: With moving the unloading of the wallet on the Bitcoin Core side back to the wallet manager, we don't need that property on the wallet side at all anymore. Will remove it.

src/cryptoadvance/specter/node.py
moneymanolis: Outdated, I introduced the property and removed it again. Not sure why GitHub showed it at all.

src/cryptoadvance/specter/node.py
```diff
@@ -534,6 +535,35 @@ def check_blockheight(self):
...
+            logger.debug(
+                f"The node datadir before get_default_datadir is: {self.datadir}"
+            )
+            datadir = get_default_datadir(self.node_type)
+            logger.debug(f"The node datadir after get_default_datadir is: {datadir}")
+        else:
+            datadir = self.datadir
+        wallet_file_removed = False
+        path = ""
+        wallet_rpc_path = os.path.join(wallet.manager.rpc_path, wallet.alias)
```
k9ert, line 554:
```suggestion
        wallet_rpc_path = os.path.join(wallet.manager.rpc_path, wallet.alias)
        try:
            wallet.rpc.unloadwallet(wallet_rpc_path)
            raise SpecterError("Trying to delete a Walletfile but wallet was not unloaded corectly!")
        except RpcError:
            pass
```
Make sure that the wallet is REALLY unloaded. If you delete the file while the wallet in still loaded, you're breaking Bitcoin Core which in turn results in  `BrokenCoreConnectionException`.

moneymanolis: Shouldn't the the SpecterError be raised in the RpcError?

moneymanolis: Apart from that, good point!

moneymanolis: Thanks for the input, resolved in https://github.com/cryptoadvance/specter-desktop/pull/1950/commits/8e3adaf369c5b8da9f6bb238d1d61c93ce28fc70

PR #1955, opened 2022-11-02
Chore: Refactor Swan Extension to cryptoadvance.specterext

k9ert:
For legacy reasons, the swan extension is located in `cryptoadvance.specter.services` but it should be located in `cryptoadvance.specterext`. This PR moves it there and creates a test along the way.


src/cryptoadvance/specter/util/reflection.py
```diff
@@ -64,8 +64,8 @@ def get_package_dir_for_subclasses_of(clazz):
     elif clazz.__name__ == "Service":
         return str(
             Path(
-                import_module("cryptoadvance.specter.services").__file__
-            ).parent.resolve()
+                import_module("cryptoadvance.specterext.devhelp").__file__
```
moneymanolis, line 67: why devhelp?

tests/test_services.py
```diff
@@ -369,3 +370,45 @@ def test_both_Storages_in_parallel(empty_data_folder, user1, user2):
...
+
+    addr_obj_mock = MagicMock()
+    addr_obj_mock.used = False
+    addr_obj_mock.is_reserved = False
+
+    wallet_mock.get_address_obj.return_value = addr_obj_mock
+
+    addresses = s.reserve_addresses(wallet_mock, "someLabel", 2, False)
+    print(addresses)
+    assert addresses == ["a", "b"]
```
moneymanolis, line 414: Could we add an assert like `assert wallet_mock.address_index == X `here in the end to verify that the address index was adjusted correctly or is this not possible with the mock?

tests/test_services.py
```diff
@@ -369,3 +370,45 @@ def test_both_Storages_in_parallel(empty_data_folder, user1, user2):
...
+    caplog.set_level(logging.DEBUG)
+    specter_mock = MagicMock()
+    specter_mock.data_folder = empty_data_folder
+    s = TestService(True, specter_mock)
+
+    wallet_mock = MagicMock()
+    # We assume that we haven't yet reserved any addresses:
+    wallet_mock.get_associated_addresses.return_value = []
+    # We assume that those are the unused addresses returned by the wallet
+    wallet_mock.get_address.side_effect = ["a", "b", "c", "d", "e", "f"]
```
moneymanolis, line 402: Perhaps `side_effect` works as well but wouldn't `return_value` be more correct?

moneymanolis (approved):

src/cryptoadvance/specter/util/reflection.py
k9ert: I'll add a comment:
```
        # here, we'd like to know the Path of the namespace module cryptoadvance.specterext
        # but that fails because you cannot import namespace-modules.
        # So we take a module that we know exists and take its parent.
```

tests/test_services.py
k9ert: No, i don't think so. You would return an array each time but with side_effect, it's an iterable.
https://stackoverflow.com/questions/24897145/python-mock-multiple-return-values

src/cryptoadvance/specter/util/reflection.py
moneymanolis: OK, not trivial :-)

PR #1958, opened 2022-11-04
Bugfix: Various bugfixes

k9ert:
* If Extensions are unknown, we should fail a bit more gracefully
*


k9ert:
Messed up branching

PR #1959, opened 2022-11-04
Bugfixes: Various bugfixes

k9ert:

PR #1960, opened 2022-11-04
Bugfix: Various bugfixes2

k9ert:
This is build upon #1959 but adds more bugfixes on top. It's separated in order to make smaller PRs. Also both PRs (#1959 and this one) mostly contain bugfixes which are also contained in #1952 but in order to separate bugfixing from Spectrum Addition, we're putting them in here.
* [nodes template magic to improve extensibility](https://github.com/cryptoadvance/specter-desktop/pull/1960/commits/cd9fe9a27afc7e781991d5de707ed62ecc081789). You can now add various templates to a node-class which will get respected e.g. if you click on edit or show an image representing that node or simply showning the status in the upper left corner.
* We have sometimes issues with Threads not being aware of the app-context. For that we created `FlaskThread`. But sometimes now, we have the opposite problem: A thread might not need the App-context but FlaskThread is used. So let's make it not count: [smaller bugfixes](https://github.com/cryptoadvance/specter-desktop/pull/1960/commits/fa92cf6cda208095a0f1d2c1879d2aa2c1ef524e). One for all. We might still get `RuntimeError` with no app-context available, but in those cases we will now know much better what's wrong.


src/cryptoadvance/specter/util/flask.py
```diff
@@ -8,14 +8,28 @@
 
 class FlaskThread(Thread):
     """A FlaskThread passes the applicationcontext to the new thread in order to make stuff working seamlessly in new threadsS
-    copied from https://stackoverflow.com/questions/39476889/use-flask-current-app-logger-inside-threading"""
+    copied from https://stackoverflow.com/questions/39476889/use-flask-current-app-logger-inside-threading
```
moneymanolis, line 11: Please check the concept again, is this useful like this?

k9ert:
I guess i screwed up with that force-push. Setting to draft.

k9ert:
The gist of all of that has been incorporated in other PRs already. Closing.

Issue #1803, opened 2022-07-14
Transactions in "unsigned" can be signed, just unbroadcast.


BitcoinMechanic:
The text here should probably read "Transactions Awaiting Signing/Not Yet Broadcast To The Network"

![image](https://user-images.githubusercontent.com/45926711/179061258-0924d345-d93a-4c47-8344-cd713808b8e7.png)

k9ert:
So a reasonable (not too long) text would be "Transactions Awaiting Signing/Broadcasting" ?

BitcoinMechanic:
Yes

BitcoinMechanic:
Or a whole new tab - "Signed (unbroadcast)

PaarthAgarwal:
I feel this issue has been solved by #1972 and can be closed now.

PR #1972, opened 2022-11-16
Improve copy for incomplete transactions tab

otkstdio:
As mentioned in this issue: https://github.com/cryptoadvance/specter-desktop/issues/1803

> The text here should probably read "Transactions Awaiting Signing/Not Yet Broadcast To The Network"

The fix in this PR is a small copy change to make it clear that the "Unsigned" tab actually includes both unsigned and unbroadcasted transactions. I've renamed the tab to "Incomplete" and added further clarity to the states that are included in the tab when inside it as per @k9ert's suggestion:

> So a reasonable (not too long) text would be "Transactions Awaiting Signing/Broadcasting" ?


k9ert:
I'd like to see screenshots for all ui-related changes. It's a great way of backtracking changes and makes it much more explicit what have changed and why. Can you add that?

otkstdio:
@k9ert Sure! The issue was that in the send/unigned tab it included both unsigned and unbroadcasted transactions, but it didn't show that unbroadcasted transactions would/should exist here. This is a copy change to make the more clear. The discussion for this issue can be found here: https://github.com/cryptoadvance/specter-desktop/issues/1803

Here are the before and after screenshots.

Before: 
Screen Shot 2022-11-21 at 11 09 40 AM

After:
Screen Shot 2022-11-21 at 11 12 08 AM

relativisticelectron:
I think "Incomplete" is even less fitting than "Unsigned".  What about "**Saved**" (borrowed this terminology from electrum, where one can "save" a transactional locally, meaning it is not yet broadcasted but may be signed).

moneymanolis:
I like "Saved" - incomplete is really not so great. A signed tx that you only need to broadcast is not sth. I'd intuitively consider "Incomplete". @otkstdio can you change if you are okay with it?

otkstdio:
@relativisticelectron @moneymanolis Great suggestion! I like "Saved" more to. I prefer things that lean more positive in their language so this is a nice improvement.

otkstdio:
@relativisticelectron @moneymanolis Updated 👍

moneymanolis (approved):

Issue #1925, opened 2022-10-15
Bug: Sorting in tx table doesn’t work


moneymanolis:
For example clicking on the arrow next to time in the screenshot doesn't do anything, same for other columns:

<img width="844" alt="grafik" src="https://user-images.githubusercontent.com/70536101/195996431-1b16c8c0-a9f7-455d-bab4-1b5a2c5cdfc4.png">

Issue #1805, opened 2022-07-14
UI: Sorting in the tx-table


relativisticelectron:
The UI is confusing, suggesting there is column sorting available:
![sorting](https://user-images.githubusercontent.com/60378539/179070812-b268a976-a2eb-4ff6-8635-6b8d659975bd.gif)

k9ert:
Agree. That's indeed a stupid thing. I'd even recommend to remove these useless errors as a good_first_issue rather than having that confusion.

PaarthAgarwal:
I think this issue has been solved by #1973 and can be closed.

k9ert:
Thank you @PaarthAgarwal . Reviewing issues like that is very much appreciated!

PR #1973, opened 2022-11-16
Cleanup:(tx-table) Remove arrows and clickable headers on transactions table

otkstdio:
While the transactions sortability is broken i've removed the clickability and arrows to avoid confusing the user. This is mentioned in both https://github.com/cryptoadvance/specter-desktop/issues/1805 and https://github.com/cryptoadvance/specter-desktop/issues/1925. Following @k9ert suggestion to just remove the arrows for now while the core functionality doesn't work.

>  I'd even recommend to remove these useless errors as a good_first_issue rather than having that confusion.


k9ert:
Good! Can you do screenshots before/after?

otkstdio:
@k9ert This just removes the code that adds the clickability of the table headers, while the sorting doesn't actually work, as per the suggestion from this issue: https://github.com/cryptoadvance/specter-desktop/issues/1805

You can no longer click the headers and toggle the arrows for the sorting of the transaction table seen here:
195996431-1b16c8c0-a9f7-455d-bab4-1b5a2c5cdfc4

moneymanolis:
Just fyi, @otkstdio I just removed the arrow for the time column.

moneymanolis (approved):

Issue #1968, opened 2022-11-11
Bug: Node details are lost


relativisticelectron:
**Describe the bug**
After switching from [4aa5858](https://github.com/cryptoadvance/specter-desktop/pull/1901/commits/4aa585893e3649afd52fb06fe25a8e294a6f1b46) (of https://github.com/cryptoadvance/specter-desktop/pull/1901 )  to  current master   a739a062 the node details are not loaded (or are lost): 
![image](https://user-images.githubusercontent.com/60378539/201314369-8b259370-a465-451c-b881-bcacad9bdfd3.png)

- When entering the node details, it can find the node, but doesn't load the wallets
- After restarting specter again the node details are lost again.

**Cause**
My suspicion is that one of these merges on master in the meantime caused this:

- de8f5af619d06fdb8b508dae1c5e54d36f5b3bd3
- 55e2a5fb761b8a48c53f699724580a22fcbb00a2
- 65bb7d2309845270e123718b9ab17bf1d94077d5
- ba5873d23be308c67c9446c928880bd9c4b1067b
- 1d2fea976ac06823cbbfc43b46f6b26ed342bab7
- bef9cf69f88db10f8aec4ded8c8fc73355ef8e7f
- a739a062ca44005a1613f072f8f99def3c71c6c0

**Desktop (please complete the following information):**
 -  local(external) , regtest
 - Linux

moneymanolis:
Can confirm, there a gazillion node jsons created:
![grafik](https://user-images.githubusercontent.com/70536101/201761116-453236bf-d4d9-40cb-8645-c77576bcc487.png)

PR #1974, opened 2022-11-16
Chore: Node manager refactoring

moneymanolis:
- This PR changes the keys in the `node_manager.nodes `dictionary to the node aliases instead of the node names to avoid bugs like we had in https://github.com/cryptoadvance/specter-desktop/issues/1968. This PR fixes https://github.com/cryptoadvance/specter-desktop/issues/1968
- Bugfix where the default node was created all the time not only when it was not available anymore
- Node manager test coverage improved

### DIFF
--- a/src/cryptoadvance/specter/managers/node_manager.py
+++ b/src/cryptoadvance/specter/managers/node_manager.py
@@ -29,7 +29,9 @@ class NodeManager:
         internal_bitcoind_version="",
         data_folder="",
     ):
-        self.nodes = {}
+        self.nodes = (
+            {}
+        )  # Sth. like: {'nigiri_regtest': <Node name=Nigiri regtest fullpath=...>, 'default': <Node name=Bitcoin Core fullpath=...>}
         self.data_folder = data_folder
         self._active_node = active_node
         self.proxy_url = proxy_url
@@ -51,15 +53,13 @@ class NodeManager:
             # creating folders if they don't exist
             if not os.path.isdir(data_folder):
                 os.mkdir(data_folder)
-        nodes_files = load_jsons(self.data_folder, key="name")
+        nodes_files = load_jsons(self.data_folder, key="alias")
         for node_alias in nodes_files:
             try:
-                self.nodes[
-                    nodes_files[node_alias]["name"]
-                ] = PersistentObject.from_json(
+                self.nodes[node_alias] = PersistentObject.from_json(
                     nodes_files[node_alias],
                     self,
-                    default_alias=node_alias,
+                    default_alias=nodes_files[node_alias]["alias"],
                     default_fullpath=calc_fullpath(self.data_folder, node_alias),
                 )
             except SpecterInternalException as e:
@@ -67,6 +67,9 @@ class NodeManager:
 
         if not self.nodes:
             if os.environ.get("ELM_RPC_USER"):
+                logger.debug(
+                    "Creating an external Elements node with the initial configuration."
+                )
                 self.add_external_node(
                     node_type="ELM",
                     name="Blockstream Liquid",
@@ -79,7 +82,9 @@ class NodeManager:
                     protocol="http",
                     default_alias=self.DEFAULT_ALIAS,
                 )
-            logger.info("Creating initial node-configuration")
+            logger.debug(
+                "Creating an external BTC node with the initial configuration."
+            )
             self.add_external_node(
                 node_type="BTC",
                 name="Bitcoin Core",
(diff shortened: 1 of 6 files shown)


moneymanolis:
@relativisticelectron could you review?

src/cryptoadvance/specter/managers/node_manager.py
```diff
@@ -29,7 +29,9 @@ def __init__(
         internal_bitcoind_version="",
         data_folder="",
     ):
-        self.nodes = {}
+        self.nodes = (
+            {}
+        )  # Sth. like: {'nigiri_regtest': <Node name=Nigiri regtest fullpath=...>, 'default': <Node name=Bitcoin Core fullpath=...>}
```
k9ert, line 34: Why the single brackets? I don't understand that. The example does not help with this?!

k9ert (commented):
One question for the internal dictionary representation but LGTM as far as i can see.

src/cryptoadvance/specter/managers/node_manager.py
moneymanolis: They are from black because of the comment added.

k9ert (approved):

relativisticelectron:
> @relativisticelectron could you review?

When switching from my working configuration (on PR https://github.com/cryptoadvance/specter-desktop/pull/1901 ) to this branch, the node details are lost.

moneymanolis:
@relativisticelectron could please check against master, #1901 is a whole different beast.

relativisticelectron:
> @relativisticelectron could please check against master, #1901 is a whole different beast.

- 9a3f92d6 is the last working commit on master
- de8f5af6 breaks the node savings on master
- Jumping from 9a3f92d6 to  this PR  fdf248ef  breaks the node savings

moneymanolis:
@relativisticelectron could you elaborate a bit more on "the node details (default.json) are not found and new node json are created"

What is your setup exactly? How many node jsons do you have etc. Which node jsons are created?
There is also a migration happening across those commit I think. See here: https://github.com/cryptoadvance/specter-desktop/blob/de8f5af619d06fdb8b508dae1c5e54d36f5b3bd3/src/cryptoadvance/specter/util/migrations/migration_0002.py
where this key-value pair is added: `"python_class": "cryptoadvance.specter.node.Node"`

I'd need to replicate your issue basically :-)

src/cryptoadvance/specter/managers/node_manager.py
moneymanolis: Resolved in https://github.com/cryptoadvance/specter-desktop/pull/1974/commits/b19cb6d5bcc5dd37bf921d6b92a586135683f83b

moneymanolis:
> > @relativisticelectron could please check against master, #1901 is a whole different beast.
> 
>     * [9a3f92d](https://github.com/cryptoadvance/specter-desktop/commit/9a3f92d6187e933d2ff47ea5986356fa326cfd9a) is the last working commit on master
> 
>     * [de8f5af](https://github.com/cryptoadvance/specter-desktop/commit/de8f5af619d06fdb8b508dae1c5e54d36f5b3bd3) breaks the node savings on master
> 
>     * Switching from [9a3f92d](https://github.com/cryptoadvance/specter-desktop/commit/9a3f92d6187e933d2ff47ea5986356fa326cfd9a) (working) to  this PR  [fdf248e](https://github.com/cryptoadvance/specter-desktop/commit/fdf248ef315e54db0d44ec280ef3461caaadf62f)  loses the node details

I think the reason is by doing this you skip the migration and that it is why you don't have `"python_class": "cryptoadvance.specter.node.Node"` in your node jsons. 

@k9ert just double checking can I assume in this PR that the node-migration has happened? Or do we have to handle the case that it hasn't?

k9ert:
I can't imagine how the migration gets prevented?! Persistence changing releases are some kind of special beast. At least it makes it easier, to put it into code. [migration_0002.py](https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/util/migrations/migration_0002.py) is already in master but we haven't had a release until then. So assume:
* that if the release happens, it'll affect all old nodes and the migration will run on all of them with no obvious way of rolling back.
* Yes, we'll have a more or less comprehensive description to roll back but we need to prepare to roll forward.
* The current state of the migration was buggy (my fault, didn't tested it good enough) and so this is a bugfix. As we haven't released yet, you need to assume that the migration will run on top of this PR if you want to test things. This is especially annoying situation for developers, especially if you test with a lot of different nodes.

> @k9ert just double checking can I assume in this PR that the node-migration has happened? 
> Or do we have to handle the case that it hasn't?

So short: answer: You need to assume that the migration happens after this PR and we shouldn't handle any other case. But i'm not yet sure which issue arises here as you haven't touched the migration-code and it shouldn't matter as much:
* Either the migration ran and the changes are there or
* a developer rolled back to work on an other PR which needs the old structure but in this case he also is suppose to clean up the migrations_data_structure

Last line: 
https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/util/migrations/migration_0002.py#L32-L47

k9ert:
@moneymanolis please have a look at my last three commits, here. Replacing the name with alias in the migration-script solved the ever growing node-list for me.

relativisticelectron:
> a developer rolled back to work on an other PR which needs the old structure but in this case he also is suppose to clean up the migrations_data_structure @relativisticelectron

Ok, thanks. Cleaning up the migration_data.json  solved it. It seems I had migrations ( 2 and 3) from previous experiments in migration_data.json and that prevented the essential migration from happening here.

Issue #1922, opened 2022-10-15
UIUX: Labeling is clunky


moneymanolis:
The labelling works but is not very user friendly and probably not consistent across browsers.

<img width="281" alt="grafik" src="https://user-images.githubusercontent.com/70536101/195983209-5b7b2a16-2c1d-4a29-a7bc-7a19dd1379ed.png">

moneymanolis:
Cursor works in Chrome / Chromium

PR #1978, opened 2022-11-17
feat:(label) Improved address label user experience

otkstdio:
New label UX for easier labelling and cleaner UX. Targeting this issue https://github.com/cryptoadvance/specter-desktop/issues/1922

https://user-images.githubusercontent.com/85369328/202569641-17f4f320-99bc-447a-8d5e-d1ed15c731c3.mov


k9ert:
Awesome! Looks really good! LGTM!

moneymanolis:
We are trying to support Firefox and Chrome/Chromium, so my tests are with both:

**Chrome:**
Works really well. Just the positioning of the service logos could be improved:

Under addresses:
grafik

Under transactions - history:
grafik

**Firefox:**
Doesn't really work, no cursor, can't confirm, only cancel, can't input anything:

moneymanolis:
I wanted to add proper backend validation / error handling at wallets api endpoint to ensure consistency between front end changes and the backend. Following the code I can see that we use rpc calls (setlabel and listlabels) in `addresslist.py` which confused me since I thought we didn't use rpc calls in this respect anymore. @stepansnigirev @k9ert - could you help clarify?

src/cryptoadvance/specter/templates/includes/address-label.html
```diff
@@ -1,73 +1,104 @@
...
         .label {
             word-break: break-all;
             background: none;
             color: #fff;
             font-size: 1em;
             max-width: 80%;
             outline: none;
+            margin-top: 1px;
+            margin-left: 2px;
+            white-space: nowrap;
```
moneymanolis, line 23: This causes that the addresses here are not fully displayed:

grafik

We could keep the `nowrap` and just shorten the addresses like we do in the addresses overview:

grafik

moneymanolis:
Short video on the changes in https://github.com/cryptoadvance/specter-desktop/pull/1978/commits/d5a0c1657fe5a2f8cc8256f861e4ba458e3926f9 (not seen in the video is that if the label is empty it can still be saved, makes kind of sense since we have logos and perhaps someone just wants to have the logo):
https://user-images.githubusercontent.com/70536101/203639739-36984c83-6a97-460b-b4ce-1ca120217d50.mov

moneymanolis:
@otkstdio please have a look at my changes. If okay, could you do the two remaining tasks:

- [ ] Fix address display in transactions - history
- [ ] Fix service logo positioning in transactions - history
(both issues were "introduced" with this PR)

otkstdio:
@moneymanolis All fixed 👍

relativisticelectron:
I like the "click to edit" UX!

For long labels and narrow screen, the other columns are not visible any more.
image

moneymanolis:
@otkstdio can you have a look at electron's points?

otkstdio:
Hey @relativisticelectron!

I think we need a solution to the label being too long in another PR (one I'm working on). This is certainly a problem but it needs to be fixed in the context of the table not being able to handle all sized inputs, not just the label. I think we'll need the ability to scroll horizontally in tables with extended data points. Also, what's the use case of having a label of an address the same as the address? I'd hope users wouldn't actually do this as labels are effectively human readable descriptors to reflect the not-so-human-readable address, not the address itself. 

When it comes to clicking on other elements, I suspect this would show the user intent of cancelling the editing process and thus not updating the label. Would you agree?

relativisticelectron:
On master long labels are wrapped to multiple lines, such that all columns stay visible. I think this is the desirable behavior.




> When it comes to clicking on other elements, I suspect this would show the user intent of cancelling the editing process and thus not updating the label. Would you agree?

Yes.

moneymanolis:
@otkstdio I'd also say let's keep the wrapping of the addresses for now and change that once you came up with a more general solution that you were mentioning above:
> This is certainly a problem but it needs to be fixed in the context of the table not being able to handle all sized inputs, not just the label.

moneymanolis:
Some more issues:

The border here looks a bit random, should either be unified or dropped:

grafik

This PR deletes basically the copy functionality of saving the address of a label to clipboard. This at first look like no issue but if you open up tx info for example, there is now no way anymore to copy an address there:

moneymanolis:
Regarding the last task ("Add a copy icon for addresses in the address overview"): What do you think about sth. ike this @otkstdio (I'd leave the final styling to you of course):
https://user-images.githubusercontent.com/70536101/205062661-24cee5ab-6f8c-4d08-ad74-24d67f27491e.mov

relativisticelectron:
> I'd suggest to change the default label of the address by sth. like "External address", see screenshot above? @relativisticelectron what do you think?

If an address doesn't belong to the wallet, it seems more logical to me to show the address, and not give it any label (by default). That way the user can compare the address with the reference without any more clicks.  I.e. the current behavior on master.

moneymanolis:
> clicking into an empty field of the table should end the edit mode

@relativisticelectron how about this?

https://user-images.githubusercontent.com/70536101/206923805-c1243d6d-6b7b-4142-84c6-ef89fedeb6fe.mov

moneymanolis:
One of the major problems is that with Firefox you can use spaces in the labels. I couldn't find a fix for that. @relativisticelectron tagging you just in case you have sth. in mind to fix this, otherwise we have to leave it like that and Firefox user have to use sth. like hyphens ...

andreasgriffin:
> > clicking into an empty field of the table should end the edit mode
> 
> @relativisticelectron how about like that? 

great!

relativisticelectron:
> > clicking into an empty field of the table should end the edit mode
> 
> @relativisticelectron how about like that? 

great!

moneymanolis:
@relativisticelectron could you test before I merge this, would be great, thanks!

relativisticelectron:
Firefox: I am clicking, but cannot get rid of the selection to edit the text.
Peek 2022-12-13 18-45

moneymanolis:
> Firefox: I am clicking, but cannot get rid of the selection to edit the text.

Can't reproduce on Firefox on Mac ... :-(

relativisticelectron:
image
I am clicking on the selected text to get a cursor to edit it, and only get this message for each click:
image

moneymanolis:
As for 
> long labels are not wrapping
I know. @otkstdio will take care of that in his first or second round of CSS revamp. 

- The console message is fine / intended. 
- Thanks for the hint on the click propagation, will do that!

moneymanolis:
@relativisticelectron could you please test again? Note: propagation hasn't been addressed yet.

relativisticelectron:
> @relativisticelectron could you please test again?

Sorry, but it still doesn't work.

moneymanolis:
@relativisticelectron but this is intended / normal in the selection all that we use. Please test by removing the whole selection and then type sth. and by using the arrows and add sth. for example.
In particular, does the spacebar work as intended?

relativisticelectron:
> @relativisticelectron but this is intended / normal in the selection all that we use. 

I expect a editable field to work not only with the keyboard but also with the mouse.  The wallet-name edit function works as I would expect an editable field to work:
Peek 2022-12-15 09-28


> Please test by removing the whole selection and then type sth. and by using the arrows and add sth. for example. In particular, does the spacebar work as intended?

Yes, spacebar works as intended.

moneymanolis:
The relevant lines are (in the `edit.onclick()` in `address-label.html`)
```
    let selection = window.getSelection();
    selection.selectAllChildren(this.label);
```

The idea behind this is, that you usually want to replace the default label in the addresses (sth. like Address #5) completely and the above lines make this easier. Let me think about perhaps checking whether this is the first click on the label, enabling the editing (select all) and then the next clicks are not selecting anymore.

relativisticelectron:
Peek 2022-12-15 21-05
It seems clicking on the side of the update button fails to update it.  (Center clicking works)

moneymanolis:
> It seems clicking on the side of the update button fails to update it. (Center clicking works) The console shows Clicking somewhere else on the screen. Canceling editing.

I see, `cancelEditing` was called first (via the global event listener which ensures the "clicking somewhere else cancels the editing"). 

From GPT:
"Event listeners are executed in the order in which they are registered. If you register the global event listener first and then the event listener for the update button, the global event listener will be executed first and then the event listener for the update button will be executed."

I am using stopPropagation now on all click event handler in the component so the global one is not even receiving the event and it also makes checking where the click originated from in the global listener obsolete.

moneymanolis (approved):

moneymanolis:
@relativisticelectron can I merge?

k9ert (approved):

moneymanolis:
I am merging this now and will directly open an issue with @relativisticelectron input to deal better with long labels.

Issue #1963, opened 2022-11-08
error: redirect_uri_mismatch


aphex3k:
> # oops! something went wrong
> ``` 
> error: redirect_uri_mismatch
> error_description: redirect_uri did not match any of the client's registered redirect_uris
> state: kjkmdskdmsmmsmdslmdlsm
> ```

OAuth with Swan is not working and throws a `redirect_uri_mismatch` error instead.

**Steps To Reproduce**

1. Log in to specter desktop, e.g. 'http://umbrel.local:25441'
2. Click on '+ Choose plugins'
3. Click on 'Swan'
4. Click on 'Existing Swan users'
5. Click on 'Link your account'
6. New tab opens presenting the error shown in the screenshot
`https://api.swanbitcoin.com/oidc/auth?client_id=specter&redirect_uri=http://umbrel.local:25441/svc/swan/oauth2/callback&response_type=code&response_mode=query&code_challenge=<redacted>&code_challenge_method=S256&state=kjkmdskdmsmmsmdslmdlsm&scope=offline_access%20v1%20write:vendor_wallet%20read:vendor_wallet%20write:automatic_withdrawal%20read:automatic_withdrawal&prompt=consent`

**Expected behavior**
This has worked at least once at some point in a version of specter desktop prior to updating to 1.13.1. I don't know exactly in which version this regression was introduced though. Auto-withdrawals from Swan to Specter had been working for this account for quite some time. Eventually the auto-withdrawal stopped working which is why I attempted to re-connect the plugin.

**Screenshots**
![Fg09NavUcAENSGK](https://user-images.githubusercontent.com/433270/200470109-90747c51-72f6-4054-b373-90a6bf8a54b1.png)

 - Where is your node running: **umbrel**
 - Where is your specter running: **umbrel**
 - If you're running a binary, which OS: I don't think it matters, but I'm also running specterd hwibridge on **macOS** since I access specter remotely
 - Browser: any, I tried Chrome, Brave, Safari and Edge
 - Specter Version: **1.13.1**

moneymanolis:
Just tested the integration on master and all worked fine.

aphex3k:
Is this an issue?
![spc](https://user-images.githubusercontent.com/433270/200642402-c8a00ecf-588d-4459-98a6-2c28b38d2656.jpg)
![svc](https://user-images.githubusercontent.com/433270/200642404-f4cd48c5-6ba3-4da6-8beb-eea074a9d1d4.jpg)

[...]

PR #1981, opened 2022-11-19
Bugfix: Updating the auto-withdrawal in Swan integration could lead to an error

moneymanolis:
If you had 10 addresses reserved already and updated the withdrawal it tried to reserve 10 more addresses and went into the else part of `if len(addresses) < num_addresses `in `reserve_addresses` in `service.py` - since 10 is equal to 10.
It then returned a list of addr objs from `wallet.get_associated_addresses() `instead of a list of addr strs.

With this PR we return an empty list if there are already enough addresses reserved and don't even make another call to the Swan API if this is the case.

This PR also adds more tests to test the` reserve_addresses()` method from the Service class.
@k9ert about to whether we keep your tests with the mocks which I renamed to `test_reserve_addresses_with_mocks`

Fixes https://github.com/cryptoadvance/specter-desktop/issues/1963


moneymanolis:
TODO

- [ ] Fix the tests

moneymanolis:
Perhaps handy to fix this little UI bug here in this PR as well, the front end is not showing the chosen threshold but always weekly:

grafik

k9ert:
Can you rename `tests/test_services.py` to something like `tests/test_ext_swan_services.py`?
I'm missing the link between the test file and what's actually tested. Maybe there is more stuff in there which is rather indeed testing the services-core functionality but at least your test is testing ext-swan code?!

k9ert:
> TODO
> 
>     * [ ]  Fix the tests

I guess you rather need to fix the implementation here, as `Swan is expecting an alomost flat list of addresses`, no?

moneymanolis:
> The code is testing service-functionality but the tests are imho (?!) not testing the changes in the SwanService-class. Might be difficult to test as you need to mock super().

Let me see what I can add, the main change is in the reserve_addresses() method of the service class which is used by the reserve_addresses() method of the Swan "service" class.

moneymanolis:
> UI bug: the front end is not showing the chosen threshold but always weekly:
> 
> grafik
> 
> Addressed in [9ac7bad](https://github.com/cryptoadvance/specter-desktop/commit/9ac7bad4dbe1991022ed94f0e1889cd35b93e284)

Now:

https://user-images.githubusercontent.com/70536101/203388538-a638ea6a-af9e-42b8-99d2-807065923ec1.mov

k9ert (approved):

PR #1987, opened 2022-11-26
UIUX: Better node selection

moneymanolis:
This PR is branched off of https://github.com/cryptoadvance/specter-desktop/pull/1952, mainly since it uses the more modular node templates.

Main goals: 
- Reduce unnecessary info, e.g. block height etc. - this is what the node info templates are for.
- Make it easier for the user to see which nodes are running (green dot) and which not is selected (orange dot). Especially with Spectrum it will be quite common that two nodes are running / reachable but only one will be selected and used. 
- Put the emphasis in the pop up on node **selection** and better separate this from the configure buttons. They are moved further to the right and made a bit less striking colourwise. 

In order to easier achieve this, the popup was completely refactored to use flexboxes instead of block contents.

The indicators are not selectable they are just imgs to show whether a node is running / reachable (green) and / or is the currently selected node (orange). I just chose the svgs like this because their forms went well with the roundish style of the logo etc. I tried a bunch of stuff: E.g. instead of the green dot, I had a text saying "running". But, in the end, I though this was the cleanest. The dots have html titles, so if you hover over them in case you forgot what they represent, it shows the descriptions. The hint for the user to see what is interactive is a hover effect of the color of the node name. 

### Visuals
Screenshot:
grafik

Video:
https://user-images.githubusercontent.com/70536101/204106269-e3610652-9a9b-4e68-b083-a87bbcc58c77.mov


moneymanolis:
This is how it looks like for example to replace the green dot with a "Running" text, IMO less clean:

grafik

moneymanolis:
Perhaps we could just change the svg indicators to make them look less like radio buttons - similar to how it looks like in Trezor Suite:
grafik
grafik

moneymanolis:
Change the title perhaps to `Nodes overview` instead of `Node selection` since you can also configure the nodes in the popup?

relativisticelectron:
> Change the title perhaps to `Nodes overview` instead of `Node selection` since you can also configure the nodes in the popup?

- "Node overview" sounds good.
- I like the green/red indicator, one could even think about if an ["led" light](https://external-content.duckduckgo.com/iu/?u=https%3A%2F%2Fclipground.com%2Fimages%2Fled-clipart-icon-7.jpg&f=1&nofb=1&ipt=3d8c9d8d41e64051bf9b57280f0838de49cc791cc6aade909d1198fd737f0a3b&ipo=images) would be even better. (But it is nice as it is already)
- I do not like the yellow selected icon.  I looks like the green/yellow indicator, indicating to me that there is a problem with this node.  Perhaps a better way would be of changing the background color of this node row.

moneymanolis:
TODO
- [ ] Test Liquid nodes - are they correctly shown as running / not running?

moneymanolis:
This is how it would like now @relativisticelectron, what do you think?

https://user-images.githubusercontent.com/70536101/208509819-8b7aa4a1-cc4d-4068-926d-b6b8a0f136e6.mov

relativisticelectron:
Wow. It looks beautiful!

moneymanolis:
> Should the settings-gear be inside the yellow border? Not sure on that one.

I put it out on purpose to separate selecting a node from changing its settings / configuring it. Probably just not ideal to have both on one tiny popup but I didn't want to change too much since @otkstdio will change that probably in his css revamps anyways, so I'd just leave like this and he can take it from there.

moneymanolis:
Slight UI update in https://github.com/cryptoadvance/specter-desktop/pull/1987/commits/0518ada6e2753232a68f1f206fe0c5ab4bbac083:


https://user-images.githubusercontent.com/70536101/208658921-650132b2-3d15-4509-9f8f-3e5f5e9d1277.mov

src/cryptoadvance/specter/server_endpoints/nodes.py
```diff
@@ -372,5 +372,18 @@ def internal_node_logs(node_alias):
...
+    except SpecterError:
+        flash(
+            _(f"Can't switch to {node_alias}, the node seems to not exist anymore ..."),
+            "error",
+        )
+    except Exception as e:
+        flash(
+            _(f"There's been a glitch in the matrix: {e}"),
+            "error",
+        )
```
k9ert, line 389:
```
logger.exception(e)
```
please!

moneymanolis: Resolved in https://github.com/cryptoadvance/specter-desktop/pull/1987/commits/7e2e9a42aeb2766bdfd0a38338d814aec89358ec

k9ert (approved):

PR #1996, opened 2022-12-03
Doc: Corrected build instructions

relativisticelectron:
Fix for https://github.com/cryptoadvance/specter-desktop/issues/1935


docs/development.md
```diff
@@ -95,7 +95,6 @@ pip3 install virtualenv
 virtualenv --python=python3 .env 
 source .env/bin/activate
 pip3 install -r requirements.txt --require-hashes
-pip3 install -e .
 python3 setup.py install # also compiles the babel translation-files
```
k9ert, line 98: I would have thought that the solution here is to remove the `python3 setup.py install` but keep `pip3 install -e .`?! After all the compilation of the babel translation-files is not necessary for a dev-env, imho?!

relativisticelectron: Yes it turns out that keeping `pip3 install -e .`  without `python3 setup.py install` also works (I didn't realize that).   I would prefer however build instructions that include babel instructions.  What if someone tests translation issues?

moneymanolis: Plus one for `pip3 install -e . `- works much more reliably. @relativisticelectron if you find someone who tests translation issues I will personally guide this person through his dev setup ;-)

moneymanolis (changes requested):
Keep `pip3 install -e .`

relativisticelectron:
> Keep `pip3 install -e .`

Done.

moneymanolis (approved):

Issue #2000, opened 2022-12-05
Bug:  AttributeError: 'NoneType' object has no attribute 'group'


relativisticelectron:
**Describe the bug**
Clicking on the txid to see the tx details overlay I get an error (The tx details overlay still works though). Only with a certain tx I have this error.

**Trace**
```

[2022-12-05 21:37:29,232] INFO in wallets_api: Looking up a txid that didn't originate from the user's wallet. Can't return any tx data.
[2022-12-05 21:37:29,233] INFO in _internal: 127.0.0.1 - - [05/Dec/2022 21:37:29] "POST /wallets/wallet/2of2/decoderawtx/ HTTP/1.1" 200 -
[2022-12-05 21:37:29,273] INFO in _internal: 127.0.0.1 - - [05/Dec/2022 21:37:29] "POST /wallets/wallet/2of2/decoderawtx/ HTTP/1.1" 200 -
[2022-12-05 21:37:29,312] DEBUG in wallets_api: Derivation path of this descriptor  could not be parsed. Sth. wrong with the regex pattern which was (\/[0-9h]+)(\/[0-9h]+)(\/[0-9h]+)(\/[0-9h]+)?(\/[0-9h]+)??
[2022-12-05 21:37:29,312] ERROR in controller: Uncaught exception: 'NoneType' object has no attribute 'group'
[2022-12-05 21:37:29,313] ERROR in controller: Traceback (most recent call last):


  File "/home/<user>/programs/anaconda3/envs/specter310/lib/python3.10/site-packages/flask/app.py", line 1523, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/<user>/programs/anaconda3/envs/specter310/lib/python3.10/site-packages/flask/app.py", line 1509, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
  File "/home/<user>/programs/anaconda3/envs/specter310/lib/python3.10/site-packages/flask_login/utils.py", line 269, in decorated_view
    return func(*args, **kwargs)
  File "/home/<user>/programs/specter-desktop/src/cryptoadvance/specter/server_endpoints/wallets/wallets_api.py", line 546, in addressinfo
    handle_exception(e)
  File "/home/<user>/programs/specter-desktop/src/cryptoadvance/specter/specter_error.py", line 52, in handle_exception
    raise exception
[...]

PR #2001, opened 2022-12-05
Bugfix for addressinfo

relativisticelectron:
Fix for https://github.com/cryptoadvance/specter-desktop/issues/2000

rpc gettransaction will also return the receive tx, if the tx did not originate from the wallet, but from another loaded wallet (as far as I have experimented), so the check https://github.com/cryptoadvance/specter-desktop/blob/de8f5af619d06fdb8b508dae1c5e54d36f5b3bd3/src/cryptoadvance/specter/server_endpoints/wallets/wallets_api.py#L260  will return the tx.

This PR makes the `addressinfo` function more robust.

- [x] Is this silencing of this error OK, or was it actually intended to raise an exception for some other use-case?


src/cryptoadvance/specter/server_endpoints/wallets/wallets_api.py
```diff
@@ -527,8 +527,9 @@ def addressinfo(wallet_alias):
             match = re.search(derivation_path_pattern, descriptor)
             if not match:
                 logger.debug(
-                    f"Derivation path of this descriptor {descriptor} could not be parsed. Sth. wrong with the regex pattern which was {derivation_path_pattern}?"
+                    f"Derivation path of this descriptor {descriptor} for address {address} could not be parsed. Sth. wrong with the regex pattern which was {derivation_path_pattern}?"
                 )
+                return jsonify(success=False)
```
moneymanolis, line 532: The only thing which can't be returned in this case could be the derivation path, would be overkill to only return False because of this IMO. The regex isn't perfect, too, for a multisig it is not capturing the derivation path correctly for example.

A partial return with True would be better, sth. like:
```

            return {
                "success": True,
                "address": address,
                "descriptor": descriptor,
                "xpubs_descriptor": xpubs_descriptor,
                "derivation_path": None,
                "walletName": wallet.name,
                "isMine": address_info and not address_info.is_external,
                **address_info,  # address_info is an instance of Address(dict)
            }
```

The derivation path is also only needed for the QR code message signing in Specter AFAIK.

moneymanolis:
I can't reproduce https://github.com/cryptoadvance/specter-desktop/issues/2000
How did you achieve this?

src/cryptoadvance/specter/server_endpoints/wallets/wallets_api.py
```diff
@@ -527,8 +527,9 @@ def addressinfo(wallet_alias):
...
             logger.debug(f"This is the derivation path match: {match.group()}")
```
moneymanolis, line 533: I would check whether match is not None here:

```
if match:
    logger.debug(f"This is the derivation path match: {match.group()}")
    derivation_path = "m" + match.group()

```

moneymanolis (changes requested):
Change to check whether a derivation path can be inferred, if so, return success with everything, otherwise return as much as possible without the derivation path (should be everything else I guess).

src/cryptoadvance/specter/server_endpoints/wallets/wallets_api.py
relativisticelectron: In https://github.com/cryptoadvance/specter-desktop/issues/2000
- descriptor == ""
- xpubs_descriptor == ""
- address_info == None
- wallet.name is misleading, because the address actually does NOT belong to the wallet

So actually the return is not really a success and does not contain useful information.

src/cryptoadvance/specter/server_endpoints/wallets/wallets_api.py
relativisticelectron: I think both is fine here, since a bool(Match object) == True.
```
def search(pattern, string, flags=0):
    """Scan through string looking for a match to the pattern, returning
    a Match object, or None if no match was found."""
    return _compile(pattern, flags).search(string)
```

relativisticelectron:
> I can't reproduce #2000 How did you achieve this?

This is definitely an edge case.  Bitcoin Core `gettransaction` seems (for some reason) to return something even though the `getaddressinfo` shows  `"ismine": false`
image

In any case: specters `addressinfo` should handle an an address that doesn't belong to the wallet gracefully.

relativisticelectron:
> Change to check whether a derivation path can be inferred, if so, return success with everything, otherwise return as much as possible without the derivation path (should be everything else I guess).

I think `addressinfo` should return `success=False` if no  derivation path is contained in the `descriptor`. (I can only think of pre-BIP39 wallets where that could be the case, and specter doesn't have to support those, right?).  Otherwise the output of `addressinfo` is unreliable and requires if checks whenever derivation path is needed.

relativisticelectron:
@moneymanolis : The error handling is now much clearer.   I also added a test for this behavior.

src/cryptoadvance/specter/server_endpoints/wallets/wallets_api.py
```diff
@@ -523,20 +523,39 @@ def addressinfo(wallet_alias):
             xpubs_descriptor = wallet.get_descriptor(
                 address=address, keep_xpubs=True, to_string=True, with_checksum=True
             )
+            if (not descriptor) or (not xpubs_descriptor):
+                logger.debug(
+                    f"No descriptor or xpubs_descriptor was found for address {address} in wallet {wallet.name}"
+                )
+                return jsonify(success=False)
```
moneymanolis, line 530: I think with https://github.com/cryptoadvance/specter-desktop/pull/2014 we could just raise a SpecterError here since this shouldn't happen and then just check for the error in the client code.

moneymanolis:
> I think addressinfo should return success=False if no derivation path is contained in the descriptor. (I can only think of pre-BIP39 wallets where that could be the case, and specter doesn't have to support those, right?).

Just fyi: We have some super advanced users that use custom, non-standard derivation paths. There also some unusual derivation paths like the old multisig stuff from Coldcard `m/45'`

src/cryptoadvance/specter/server_endpoints/wallets/wallets_api.py
relativisticelectron: However, then this PR would not fix https://github.com/cryptoadvance/specter-desktop/issues/2000  any more:
> In #2000 :
> 
>     * descriptor == ""
> 
>     * xpubs_descriptor == ""
> 
>     * address_info == None
> 
>     * wallet.name is misleading, because the address actually does NOT belong to the wallet

moneymanolis:
The fetch call to addressinfo is used here:

`address-data.html`
 --> Uses almost everything from the response, the derivation_path is only used for a very specific feature, though (QR code based message signing, which is only available with SeedSigner and the DIY and only for SingleSig wallets)

`address-row.html`
  --> Uses the descriptors for displaying the address on device

`tx-data.html`
  --> Only uses the `isMine` info

My suggestion would be, assuming https://github.com/cryptoadvance/specter-desktop/pull/2014, don't define success `true` or `false` at the endpoint (basically get rid of `"success": bool`), especially since success is not clearly defined here, but rather let the individual client code decide whether it has everything it needs. If SpecterErrors are raised on the server side (e.g. because `wallet.get_descriptor `fails) they will - with #2014 - automatically result in error messages (these can also be picked up in the client code, see for example: https://github.com/cryptoadvance/specter-desktop/pull/2014/commits/f6e4d84be16ba9f7a7c6a18fb69737b32b87a937).

src/cryptoadvance/specter/server_endpoints/wallets/wallets_api.py
moneymanolis: Can you write a test perhaps for the concrete issue, still not getting it. Or have you already done this, if so, please point to it. This would also ensure that this PR in the end solves the original issue.

moneymanolis: See comment below regarding `success`.

relativisticelectron:
> My suggestion would be, assuming #2014, don't define success `true` or `false` at the endpoint (basically get rid of `"success": bool`), especially since success is not clearly defined here, but rather let the individual client code decide whether it has everything it needs. If SpecterErrors are raised on the server side (e.g. because `wallet.get_descriptor `fails) they will - with #2014 - automatically result in error messages (these can also be picked up in the client code, see for example: [f6e4d84](https://github.com/cryptoadvance/specter-desktop/commit/f6e4d84be16ba9f7a7c6a18fb69737b32b87a937)).

I agree `success` is not defined here. I understand it as: "Everything worked."

If I put myself in a API-user/JS client perspective, I would like:

- I expect API responses to have complete info.   API calls with results on "best effort" basis, should be marked as "partial result"/"incomplete json" or such.
- If the API is not able to give a full result, give me an error/errorcode/"success=false" (something I can catch in Javascript), (and optionally give me more info about the failiure, or other available info)

@k9ert : It seems https://github.com/cryptoadvance/specter-desktop/pull/2014 uses both fields `success` or `error` in the response.  Could we agree on 1 field to make catching errors consistent in all API calls?

moneymanolis:
OK, so how about just splitting up the API better to have complete info:

`tx-data.html` is easy:
While fixing the tests yesterday, I saw that there is a dedicated endpoint for `is_address_mine ` (returns a boolean), we you could just swap with this one and done.

`address-data.html`
For this we could make a separate endpoint for derivation paths. If we can't retrieve a derivation path we should disable the message signing feature (it will fail ugly down the line somewhere I guess, perhaps only at signing device stage ...).

`address-row.html`
Separate endpoint just for descriptors?

@k9ert @relativisticelectron what do you think?

moneymanolis:
Pulling this out of the comments since it is important:

@relativisticelectron Can you write a test perhaps for the concrete issue, still not getting it. Or have you already done this, if so, please point to it. This would also ensure that this PR in the end solves the original issue.

relativisticelectron:
> @relativisticelectron Can you write a test perhaps for the concrete issue, still not getting it. Or have you already done this, if so, please point to it. This would also ensure that this PR in the end solves the original issue.

The test with
```
# send post request with address, not belonging to wallet
```
simulates https://github.com/cryptoadvance/specter-desktop/issues/2000 . 
It checks
```
    assert res.status == "200 OK"
    assert res.data.decode() == '{"success":false}\n'
    assert (
        caplog.text.count(
            f"No descriptor or xpubs_descriptor was found for address {valid_address_not_beloging_to_wallet} in wallet ghost_machine"
        )
        == 1
    )
```

moneymanolis:
Getting it now. 

IMO my suggestion 
> tx-data.html is easy:
While fixing the tests yesterday, I saw that there is a dedicated endpoint for is_address_mine (returns a boolean), we you could just swap with this one and done.)

would already fix. It would then just check whether the address belongs to the wallet or not and not try to get a descriptor / derivation / what not.

moneymanolis:
Let's see what @k9ert thinks about my suggestion and take from there.

moneymanolis:
@relativisticelectron just making you aware of this milestone, if you refactor sth. from the list, just add a "DONE" at the end:
https://github.com/cryptoadvance/specter-desktop/milestone/4

relativisticelectron:
> @relativisticelectron just making you aware of this milestone, if you refactor sth. from the list, just add a "DONE" at the end: https://github.com/cryptoadvance/specter-desktop/milestone/4

Thanks.    I touched `tx-data.html` in https://github.com/cryptoadvance/specter-desktop/pull/2001/commits/16485aed48db498c270ad6b346963f884e1fdfb6 but I did not refactor all of  API Error Handling in  `tx-data.html`.

moneymanolis:
Adding this input from GPT, perhaps it would be best to avoid the simply returning of Booleans at the endpoints.

image

In the case of `is_address_mine` for example, wouldn't it be better to return sth. like this @relativisticelectron:
```

  address_lookup = {
    'is_mine': True,
  }
  return jsonify(address_lookup)
```

andreasgriffin:
> It seems #2014 uses both fields success or error in the response.

@k9ert : I think an error/success cleanup PR would be a good followup to  #2014 and https://github.com/cryptoadvance/specter-desktop/pull/2001 .   I would keep this PR however limited to the problem it was there to address. I think it is complete in that regard. This PR

- [ ] Fixed the bug in `addressinfo` 
- [ ] has a pytest to test `addressinfo` for the Bug in #2000 
- [ ] Simplified the API call by replacing `addressinfo` with `is_address_mine`
- [ ] `is_address_mine` returns a json with ` 'is_mine' ` instead of a boolean

relativisticelectron:
> It seems #2014 uses both fields success or error in the response.

@k9ert : I think an error/success cleanup PR would be a good followup to  #2014 and https://github.com/cryptoadvance/specter-desktop/pull/2001 .   I would keep this PR however limited to the problem it was there to address. I think it is complete in that regard. This PR

- [x] Fixed the bug in `addressinfo` 
- [x] has a pytest to test `addressinfo` for the Bug in #2000 
- [x] Simplified the API call by replacing `addressinfo` with `is_address_mine`
- [x] `is_address_mine` returns a json with ` 'is_mine' ` instead of a boolean

moneymanolis:
@relativisticelectron headers were missing in the tests and with the by now merged error handling `wallet.get_address_info(address=address)` throws a Request error as "intended".
Check https://github.com/cryptoadvance/specter-desktop/pull/2001/commits/823cb5ae3c6103aacdfba930e5fc8e4b3e985286 for details.

Let me know whether this is fine and then I'd merge if the tests are green.

relativisticelectron:
Thanks for adding the header. Yes it can be merged.
At some point we could remove the JS console log output (see https://github.com/cryptoadvance/specter-desktop/pull/2001/commits/f73a7f17312daa351c1978a18fdc5d51e8d19f79 ), but right now I think we can leave it in here.

moneymanolis:
> At some point we could remove the JS console log output

Yes, just having it for now for debugging. 

OK, will merge.

moneymanolis (approved):

PR #2011, opened 2022-12-08
Chore: Moving the Spectrum Extension to a core extension

k9ert:
The Spectrum extension has a lifecycle which is much closer to specter-desktop than to spectrum. Also, in the future, the spectrum extension might be the main way to use specter.

This PR is moving the Specterext Spectrum into this repo.

The corresponding PR to remove the extension from Spectrum is here: https://github.com/cryptoadvance/spectrum/pull/25


moneymanolis (approved):
Checked out both corresponding PR branches in Specter Desktop and Spectrum and all worked fine.

PR #2015, opened 2022-12-09
Bugfix: unloaded wallets at start

k9ert:
Currently, the wallets are not showing up on startup if you use Spectrum. So you have to wait 10-20 seconds or so, that they are loaded.

This PR fixes this.


moneymanolis (approved):

moneymanolis (approved):
Wallets are loaded now much faster when using Spectrum! Great!

PR #2016, opened 2022-12-10
Fix nix shell by generating hashes with pip-compile

hodlwave:
I noticed that my local dev environment using `nix-shell` was broken because the nix shell is still using python 3.8 whereas there is an ongoing migration of specter-desktop to python 3.10. Specifically, I ran into a requirement-hashes-not-found error because the `backports.zoneinfo` dependency is required for python < 3.10 ([source](https://github.com/cryptoadvance/specter-desktop/blob/2f27befb39c27da9867ae97029a09c03054f4dd3/requirements.in#L27)), but it's no longer included in [requirements.txt](https://github.com/cryptoadvance/specter-desktop/blob/master/requirements.txt). The long-term solution is to upgrade nix-shell to rely on python 3.10, but the short-term fix is to simply generate requirements.txt as part of the nix-shell shellHook.

### DIFF
--- a/utils/shell.nix
+++ b/utils/shell.nix
@@ -7,10 +7,12 @@ stdenv.mkDerivation {
     python38Full
     python38Packages.virtualenv
     python38Packages.pip
+    python38Packages.pip-tools
     libusb1
   ];
   shellHook = ''
     cd ..
+    pip-compile --generate-hashes requirements.in > requirements.txt
     pip3 install virtualenv
     virtualenv --python=python3 .env
     source .env/bin/activate


relativisticelectron:
In https://github.com/cryptoadvance/specter-desktop/commit/8b867d697a161cc09b376b87b91ad822800a310a#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552  the  `backports.zoneinfo` was removed. AFAIK it is required to run specter on python 3.8.   `backports.zoneinfo`  has to be put back into the `requirements.txt`

k9ert (approved):

k9ert:
Yes, we're on a stretch there. We need to progress and go with the flow. @relativisticelectron did a very good job with support for 3.10. Sure, we also would like to also support older versions but only to a certain extent. I don't think it's currently worth the time making efforts to systematically support 3.8 but that does not mean that we can support it by workarounds like this or wherever we can and the effort is not too high.

Thanks for the PR.

PR #2017, opened 2022-12-14
Chore: Preparation for Notification System

k9ert:
This is a prep-PR for #1766. In order to make that digestable, this PR extracted all changes on `src/cryptoadvance/specter` into this PR. So we can deal with that here and merge it earlier. So then afterwards we can focus on the rest (which doesn't need that rigorous review as it'll be an extension in Alpha.


k9ert:
No necessary.

PR #1991, opened 2022-11-30
Specter Desktop User Interface Improvement

otkstdio:
Hey all 👋,

I'm opening up a work in progress pull request here as I work toward a refined user interface for the Specter Desktop product. This PR will focus on both the onboarding experience and creating a concise design language across the product interface.

Beyond the UI improvements we're introducing an improved styling model using [Tailwind CSS](https://tailwindcss.com). I'll likely be adding some documentation to create some design standards among FOSS contributors so Specter's user interface can scale in a FOSS manner while maintaining a consistent design language. Tailwind will make this effort a lot easier as it proposes a design system across it's utility classes and also simplifies the location in which we manage styles (entirely with inline utility classes). This will improve on the current styling landscape in Specter's codebase where styles are currently spread across a root `styles.css` file, inline HTML `<style>` tags and inline `style=""` styling.

I'll attach some images that include the new designs and you can keep track of committed code to this branch below. 

About Specter

Receive

Select Device

OTK

### Task list
- [ ] Optimise screen (after device added)
grafik

- [x] Test Connection in Node Connection View
- [ ] Device connections
- [x] Wallet refreshing issue
- [ ] Titles on add device screen
- [ ] Spectrum Screens
- [x] Plugins screen
- [ ] Improve file uploads to show the file selected
- [ ] Wrong icon for wallets
- [ ] Electrum Device (http://127.0.0.1:25441/devices/new_device_keys/electrum/)
- [ ] Change “Multiple Signature” to “Multiple Signatures"
- [ ] Switch toast error when adding duplicate network node to a form error not a toast error.

PR #2037, opened 2023-01-05
autodetect from bitcoin.conf file with network set

Kexkey:
Hi Specter team!

When using a `bitcoin.conf` file with remote Bitcoin node configurations, Specter wouldn't auto-detect the right connection configs when not mainnet and more than only mainnet configs.  For some reason, it was ignoring the network when set in the config file.  For example, if bitcoin.conf file was:

```
regtest=1
rpcconnect=bitcoin
main.rpcport=8332
test.rpcport=18332
regtest.rpcport=18443
rpcuser=bitcoin
rpcpassword=CHANGEME
```

Specter kept using port 8332, which is the default port in its config.  It was reading the configurations for all networks in the config file even when a network was specified, like in the example above.  Also, when rotating between the different network configs trying to connect to the node, it was exiting on the first failure instead of continue trying with the other network configs.

Maybe I misunderstood the rationale behind ignoring the network config and keeping trying for every network config found in the config file, sorry if that's the case.  Ignoring the network config makes it impossible to run several nodes on the same machine on different ports and use the auto-detect feature.  For example, if I run mainnet, testnet and regtest nodes on the same `bitcoin` host, current algorithm would only connect to the first one responding.  I couldn't switch node by simply replacing `regtest=1` by `testnet=1` in the bitcoin.conf file, for instance.

This PR only changes rpc.py.  What it does:

- catches `BrokenCoreConnectionException` in `autodetect_rpc_confs` on connection failure so that it continues trying to connect when there are several networks in the confs;
- uses the network config in bitcoin.conf (`regtest=1` for example) to set the network and uses the network-specific config instead of trying everything found in the config file;
- takes the network-specific configs over the default ones (https://github.com/bitcoin/bitcoin/blob/master/doc/bitcoin-conf.md#network-specific-options);
- tries to connect with credentials from bitcoin.conf file first, and if it fails uses the cookie file if found (https://github.com/bitcoin/bitcoin/blob/master/doc/init.md#configuration).

Let me know if I'm confused :) or if you see improvements or a better way to do this.  I tried to minimize the changes for this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 - [Specter Desktop](#specter-desktop)
   - [DISCLAIMER](#disclaimer)
-  - [Video Walkthrough](#video-walkthrough)
+  - [Documentation and Video Walkthrough](#documentation-and-video-walkthrough)
   - [Why?](#why)
   - [Help wanted: Do you like Specter?](#help-wanted-do-you-like-specter)
   - [How to run](#how-to-run)
@@ -27,7 +27,6 @@
     The Crypto Anarchist Manifesto - Timothy C. May - Sun, 22 Nov 92 12:11:24 PST
 
 [![Build Status](https://api.cirrus-ci.com/github/cryptoadvance/specter-desktop.svg)](https://cirrus-ci.com/github/cryptoadvance/specter-desktop)
-[![Support the project](https://img.shields.io/badge/btcpay-support%20project-orange.svg)](https://donate.specter.solutions/apps/3k77BAT6zshCGNd3i7gw9WKwXQy1/pos)
 
 ## DISCLAIMER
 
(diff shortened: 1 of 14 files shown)


moneymanolis:
@Kexkey what do you mean with "bitcoin.conf file with remote Bitcoin node configurations"? Specter can only read the bitcoin.conf if this file is on the same machine as Specter is running on. 

In any case, have a look at this PR in the big UI revamp: https://github.com/cryptoadvance/specter-desktop/pull/1991/commits/1db98cf18cbc1ebdf8d777ce1ae6ffa918457343

k9ert:
I've added a test and sounds reasonable to me. @moneymanolis can we merge as the change does imho not interfere with https://github.com/cryptoadvance/specter-desktop/commit/1db98cf18cbc1ebdf8d777ce1ae6ffa918457343 ?!

Kexkey:
Oh sorry I should have added tests, thanks @k9ert for that.

@moneymanolis what I meant was "...when using a local bitcoin.conf file to auto-setup Specter to use a remote bitcoin node..."

I checked the commit you linked above but I'm not familiar enough with the code to tell you if the revamp fixes what this PR does.

Thanks for taking care of this PR!  I updated Specter to v1.14.2 for the next release of Cyphernode.  If this PR goes through to the next Specter release in time, it will be part of the final release.

You guys are doing great work!  Thank you!

k9ert:
Really happy to hear that specter is doing well in cyphernode. 
I'm confident to do a release hopefully tomorrow. Pretty sure we'll get this in here as well.
As you have an electrum on your stack, you might want to experiment with the new electrum connection which we implemented via the "spectrum-extension". It does work well for me and it's much more convenient: No more rescanning hassle. Much faster performance. Importing dozens of devices/wallets in minutes.
We created some hints here: https://github.com/cryptoadvance/specter-desktop/releases/tag/v1.14.1
Yes, it's alpha formally but i think in practice it's more beta than alpha. With the UI revamp, it'll be production ready.

k9ert (approved):

PR #2048, opened 2023-01-11
Chore: Refactor device_types and disable change to hotwallet for Spectrum_Node

k9ert:
This PR got started with preventing that the type of a device is changed to a hotwallet, if spectrum is the current node. There was already a method called `is_device_supported` in the node_implementation which checks whether a node support a device (or device class) but in this case we don't have a reference in [src/cryptoadvance/specter/server_endpoints/devices.py](https://github.com/cryptoadvance/specter-desktop/pull/2048/files/8e997d2943e35218d91e7782d66bb81296aba902#diff-d5d5418c260cd6e3f4dbcfa8272ff39754e54b3ac3b649fee180885dd05a2194) we don't have a reference to either of the two but to the string which is defined in `src/cryptoadvance/specter/devices/device_types.py`.

So we need a way to get a Device class via that string and that has been implemented via a `@classmethod` in `device.py`. However, that annoying `device_types.py` simply makes things more complex. Why isn't that string not defined in the Device class itself? So this PR refactors that as well.


src/cryptoadvance/specter/server_endpoints/devices.py
```diff
@@ -490,7 +495,12 @@ def device(device_alias):
                 )
         elif action == "settype":
             device_type = request.form["device_type"]
-            device.set_type(device_type)
+            if app.specter.node_manager.active_node.is_device_supported(
+                Device.get_device_class_by_device_type_string(device_type)
+            ):
+                device.set_type(device_type)
+            else:
+                flash(f"Type of Device {device_type} is not supported by that node.")
```
k9ert, line 503: This is the central piece of this PR. Just mentioning here.

src/cryptoadvance/specter/device.py
```diff
@@ -59,6 +61,13 @@ def blueprint(cls):
             return f"{module_array[2]}_endpoint.static"
         return "static"
 
+    @classmethod
+    def get_device_class_by_device_type_string(cls, device_type: str):
```
k9ert, line 65: Return type hint?!

src/cryptoadvance/specter/util/reflection.py
```diff
@@ -222,25 +223,33 @@ def get_subclasses_for_clazz(clazz, package_dirs: List[str] = None):
...
+def get_subclasses(cls):
+    """Naive implementation of searching for subclasses. This will only return classes which has been
+    imported. If you also want to find classes which are not imported, you need to provide package-directories
+    or well known places as for specific classes as implemented in get_subclasses_for_clazz
+    Returns all subclasses of a specific call including sususub...classes
+    """
+    subclasses = []
+    logger.fatal(f"Searching {cls}")
+    for subclass in cls.__subclasses__():
+        logger.fatal(subclass)
```
k9ert, line 251: fatal ?!

src/cryptoadvance/specterext/spectrum/templates/spectrum/settings.jinja
```diff
@@ -48,8 +48,13 @@
...
             </div>
             <div class="row" style="margin-top: 25px">
-                <button type="submit" class="btn">{{ _("Save") }}</button>
+                <button type="submit" class="btn" name="action" value="save">{{ _("Save") }}</button>
             </div>
+            {% if node_is_running %}
+                <div class="row" style="margin-top: 25px">
+                    <button type="submit" class="btn danger centered" name="action" value="delete">{{ _("Delete Node") }}</button>
+                </div>
+            {% endif %}
```
k9ert, line 57: What's that for a change?!

moneymanolis:
Changed the flash message to an error and the text a bit, but mainly disabled setting Bitcoin Core (or Elements Core) as new type in the UI with a Spectrum connection:

https://user-images.githubusercontent.com/70536101/214140671-74f8be31-6ed0-4a46-a485-119d199978ca.mov

moneymanolis (approved):

PR #2055, opened 2023-01-13
testing bridge_rpc errorhandling

k9ert:
https://github.com/cryptoadvance/spectrum/issues/24 is complaining that the error-handling is not good enough. I think in the meantime it is good enough and this test proves it.

https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specterext/spectrum/bridge_rpc.py#L109-L112


moneymanolis (approved):
LGTM

PR #2066, opened 2023-01-18
Fixes/wallet_rpc_fix

k9ert:
* We wnat to have a better grip on why wallet_manager.update() has been called. So there is a new option param which will get logged
* Also we might want to prepare for being able to have more than one node for a specific network. For that, a wallet need to readjust it's rpc-instance as soon as the wallet_manager changed its rpc-instance (and the rpc is used).

### DIFF
--- a/src/cryptoadvance/specter/wallet.py
+++ b/src/cryptoadvance/specter/wallet.py
@@ -130,9 +130,7 @@ class Wallet:
 
         self.device_manager = device_manager
         self.manager = manager
-        self.rpc = self.manager.rpc.wallet(
-            os.path.join(self.manager.rpc_path, self.alias)
-        )
+        self.manager_rpc = manager.rpc
         self._devices = devices
         # check that all of the devices exist
         if None in self.devices:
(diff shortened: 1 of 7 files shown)


src/cryptoadvance/specter/wallet.py
```diff
@@ -176,6 +174,22 @@ def __init__(
         ):
             self.save_to_file()
 
+    @property
+    def rpc(self):
+        """caches an rpc-instance as long as the rpc from the wallet_manager did not change
```
moneymanolis, line 179:
```suggestion
        """Cache RPC instance. Reuse if manager's RPC instance hasn't changed. Create new rpc instance otherwise."""

File: src/cryptoadvance/specter/wallet.py [hunk 2]
```diff
@@ -176,6 +174,22 @@ def __init__(
...
+        This instance is also used by a lot of objects created by the wallet TxList/TxItem/...
```
moneymanolis, line 180:
```suggestion
        This RPC instance is also used by objects created by the wallet, such as TxList or TxItem
```

moneymanolis:
Just added a doc string in https://github.com/cryptoadvance/specter-desktop/pull/2066/commits/9417407dd8bdd8eb0b01b4f8f35d49aff89bd724 to make it easier to follow.

PR #2082, opened 2023-01-26
Fixed FR translation

oxhak:


k9ert:
Thank you. I think there are lots of other missing translations. Opening the french version looks like half of it is in falling back to english. Would you be interested to add more?

oxhak:
Yes I will, you can merge this one in the meantime and I'll let you know when I've completed what's missing for a new commit.

PR #2102, opened 2023-01-30
Added specter installation step

PaarthAgarwal:
Closes https://github.com/cryptoadvance/specterext-stacktrack/issues/1.


k9ert:
Sorry for the late reply. I don't think that makes so much sense. In line 18 of the same file we already have an installation of specter even with an `--upgrade`. It's true that the depency to specter is not trivial and there are more than one way to fix it but i've created a whole section to address it:
https://docs.specter.solutions/desktop/extensions/intro/#virtualenv-management

So, i'll close this as not necessary.

PR #2105, opened 2023-01-31
Features/extension_graphql

k9ert:
A GraphQL Extension.
image
```
{
  bookmarks {
    title
    desc    
    
  }
}
```
```
{
  users {
    username
    
  }
}
```


k9ert:
Won't happen in master. Replaced by #2196

PR #2106, opened 2023-01-31
Features/middleware

k9ert:

So this PR does those things:
- Implementing a new `return_style` called "middleware" and the old-style will be called `collect`:
  - `collect` will return a dict where the extension_id will be used as key and the returnvalue as value.
  - We have some documentation [here](https://github.com/cryptoadvance/specter-desktop/pull/2106/files#diff-4c264fd49854b2544ada9087b45610508c627ade4440754acd6825514a592100)


k9ert:
Closing this in favor of https://github.com/cryptoadvance/specter-desktop/pull/2186
This should go into uiux-revamp and not into master.

Issue #2123, opened 2023-02-01
Import wallet screen not properly styled


moneymanolis:
![grafik](https://user-images.githubusercontent.com/70536101/216089144-8a5f4d95-9847-48c0-a298-d482aaeb3bd4.png)

indra7777:
sir, can i try to align "import wallet screen"
This is the sample of ux design


<img width="1156" alt="Screenshot 2023-02-01 at 10 08 46 PM" src="https://user-images.githubusercontent.com/98700452/216105875-cc42a254-7df2-45be-a4a6-dc4f52f0997b.png">

PaarthAgarwal:
I can take this on but I want to know what do you mean by not properly styled? Do you want to fix spacing and alignment or you want other changes?

moneymanolis:
Best to compare with other screens, especially the send screen.

indra7777:
Sir , I mean to contribute to open source
If you want sir, i can change the ui
Please help me how to do and how to contribute

indra7777:
If it is not styled my ui not properly,
Then I will try to change ui better
Please help me...

moneymanolis:
It has to fit into the existing styling.

PR #2130, opened 2023-02-01
Import wallet screen improvement

PaarthAgarwal:
Addresses #2123. 
I've started out by replacing the tailwind button with classic button as it goes more with the theme. Before moving further ahead I think it'll be better to discuss what we want first.
Import wallet screen after this commit:
Screenshot from 2023-02-02 00-24-25

Send screen:
Screenshot from 2023-02-02 00-25-09

cc: @moneymanolis @otkstdio 
On comparing both the screens what changes would you like to be made to the import wallet screen?


moneymanolis:
This was fixed in https://github.com/cryptoadvance/specter-desktop/pull/2157 - closing it now.

PR #2134, opened 2023-02-01
Restore "my address functionality"

moneymanolis:
Instead of colours, which weren't so self-explanatory, I switched to changing the address label dynamically @otkstdio :

https://user-images.githubusercontent.com/70536101/216165837-6d2e6354-faf8-4a5b-9edd-f22148947d29.mov

@relativisticelectron OK for you as original inventor of this functionality?


k9ert (approved):

src/cryptoadvance/specter/templates/includes/recipient-box.html
```diff
@@ -488,7 +494,12 @@
 		markRecipient(is_address_mine) {
```
moneymanolis, line 494: @otkstdio this is the relevant part that changes stuff in the UI if the address belongs to the wallet of the sender

Issue #2143, opened 2023-02-02
v2.0.0-pre6: Save Backup PDF doesn’t work. 


moritzwietersheim:
- When Wallet was added, the download of the Backup pdf doesn't work. Not clickable. Doesn't react.

![image](https://user-images.githubusercontent.com/47259243/216223632-f4e4b182-8ed3-4975-9147-aa22ac20e733.png)

Shivoham102:
Hey, I'm new here and want to solve this issue.

Shivoham102:
I think the problem is that there is no Roboto fonts in the fonts folder
![image](https://user-images.githubusercontent.com/66953222/216354483-2e467461-c911-4024-a043-ae262d2599c2.png)

After adding the fonts to the folder this is the only error that's thrown
![image](https://user-images.githubusercontent.com/66953222/216354808-d8324180-34b9-467f-94bc-e795a92f26ae.png)

PR #2149, opened 2023-02-02
Fixed Save Backup PDF Button after creating wallet

Shivoham102:
As mentioned in the issue [2143](https://github.com/cryptoadvance/specter-desktop/issues/2143), the save pdf button was not working after creating a wallet.
I have made two changes-
1. Added the 'RobotoMono-Regular.ttf' and 'RobotoMono-Bold.ttf'  files from the Google fonts website.
2. Changed the index of  `childNodes[]` while accessing the QR Code div to access the image correctly.


Shivoham102:
@k9ert  
Please see lines 13 and 14 in the `wallet_pdf.jinja` file which are these-
```
doc.addFont("{{ url_for('static', filename='fonts/RobotoMono-Regular.ttf') }}", "Roboto", "normal");
doc.addFont("{{ url_for('static', filename='fonts/RobotoMono-Bold.ttf') }}", "Roboto", "bold");
```
These two lines attempt to access the RobotoMono font files.

Also, as for `childNodes[6].children[1].children[1].src` I have only changed the index but I'll try to use the ID now :)

moneymanolis:
@Shivoham102 it is fine / necessary to reintroduce the fonts, they were deleted during the UIUX revamp.
> Also, as for childNodes[6].children[1].children[1].src I have only changed the index but I'll try to use the ID now :)
If you see bad code, don't replicate it.

moneymanolis (changes requested):
Either make it work with an id or we have to wait until @otkstdio is done with the styling in `qr-code.html` since this influences the correct child node.

moneymanolis:
> Could you please point me to it?

Have a look at the component `qr-code.html`

moneymanolis:
Ideally, also add a little Cypress test to test the functionality and make sure with that next time it will be discovered by the the front end tests.

k9ert:
The Uncle-Jim-PDF (just below that button) is broken as well:
```
Uncaught TypeError: document.getElementById(...).shadowRoot.childNodes[3].children[1] is undefined
    <anonymous> http://localhost:25441/wallets/wallet/bluesimple_4/settings/clearcache:3366
5 clearcache:3366:35
    <anonymous> http://localhost:25441/wallets/wallet/bluesimple_4/settings/clearcache:3366
```
Maybe you can wrap the code in try/except so that the user does get a feedback when he clicks. In that case, you can call `showError` (search for an example).
I also created some tests in cypress which are testing the pdf and other download artifacts.
It should currently fail as the Uncle Jim is broken.

cypress/support/commands.js
```diff
@@ -181,6 +182,8 @@ Cypress.Commands.add("addWallet", (walletName, walletType, funded, nodeType, key
       }
       cy.get('[data-cy="create-wallet-btn"]').click()
       cy.get('[data-cy="new-wallet-added-headline"]')
+      cy.get('#pdf-wallet-download').click()
```
moneymanolis, line 185: I don't like that we check this here in the general command! Slows down the tests, makes them super vulnerable. Why not test this once in a dedicated test?

moneymanolis:
@k9ert can you move the Cypress test to a dedicated test, otherwise it will super hard to debug flakiness  with this feature, especially since it is poorly implemented right now. You can just reuse the Ghost wallet in this test. If it is in a separated tests we know directly that the pdf stuff is causing issues and not sth. with the Ghost wallet (basic functionality).

Shivoham102:
> Either make it work with an id or we have to wait until @otkstdio is done with the styling in qr-code.html since this influences the correct child node.

I couldn't figure out how how to give an id to the image created by `QRCode()`, but this works:
`let qrCode = document.getElementById('export-wallet-qr-code').shadowRoot.querySelectorAll("img")[1].src` since there are only two images.
Is that fine?

k9ert:
> Is that fine?

Just saw your comment now.  I removed the download from the command and while doing that, i needed a solution as well. Digging a bit in the qr-code html and i came up with this solution
```
let qrCode = document.getElementById('export-wallet-qr-code').el.children[1].src;
```
I don't think it's possible to improve either your or my version significantly without digging into the qr-elements code.

k9ert:
Oh, test fails but cypress green. That's also a new experience. 
<details>
<summary>For the completeness:</summary>

```
caplog = <_pytest.logging.LogCaptureFixture object at 0x7f9237cb5390>

    def test_checker(caplog):
        callback_mock = Mock()
        caplog.set_level(logging.DEBUG)
        checker = Checker(lambda: callback_mock(), period=0.01, desc="test")
        checker.start()
        time.sleep(
            0.1
        )  # If the above assumptions are failing, you might want to increase this
        assert "Checker test started" in caplog.text
        assert "This message won't show again until stopped and started." in caplog.text
        checker.stop()
        time.sleep(0.01)
>       assert "Checker test stopped" in caplog.text
E       AssertionError: assert 'Checker test stopped' in '[    INFO] Checker test Checking every 0.01 seconds now cryptoadvance.specter.util.checker (checker.py:86)\n[   DEBUG...(checker.py:64)\n[    INFO] Checker test started with period 0.01 cryptoadvance.specter.util.checker (checker.py:34)\n'
E        +  where '[    INFO] Checker test Checking every 0.01 seconds now cryptoadvance.specter.util.checker (checker.py:86)\n[   DEBUG...(checker.py:64)\n[    INFO] Checker test started with period 0.01 cryptoadvance.specter.util.checker (checker.py:34)\n' = <_pytest.logging.LogCaptureFixture object at 0x7f9237cb5390>.text

tests/test_checker.py:20: AssertionError
```

</details>
I've reran the test.

cypress/integration/spec_ghost_machine.js
```diff
@@ -6,11 +6,38 @@ describe('Ghost machine', () => {
...
                 if ($body.find('[data-cy="no-core-connection"]').length) {
                     cy.connect()
-                    cy.addWallet('Ghost wallet', 'segwit', 'funded', 'btc', 'singlesig', 'DIY ghost')
+                    cy.addWallet('Ghost wallet', 'segwit', false, false, 'btc', 'singlesig', 'DIY ghost')
                 }
                 else {
-                    cy.addWallet('Ghost wallet', 'segwit', 'funded', 'btc', 'singlesig', 'DIY ghost')
+                    cy.addWallet('Ghost wallet', 'segwit', false, false, 'btc', 'singlesig', 'DIY ghost')
                 }
+                cy.get('#pdf-wallet-download').click()
```
moneymanolis, line 14: Can we please not test this here and below? Tests should be separate.

moneymanolis (changes requested):
Just one comment about the Cypress tests.

moneymanolis:
`let qrCode = document.getElementById('export-wallet-qr-code').el.children[1].src`

... is fine for me, I couldn't find sth. better from a quick look either.

k9ert (approved):

PR #2186, opened 2023-02-08
Uiux/features/middleware

k9ert:

So this PR does those things:
- Implementing a new `return_style` called "middleware" and the old-style will be called `collect`:
  - `collect` will return a dict where the extension_id will be used as key and the returnvalue as value.
  - We have some documentation [here](https://github.com/cryptoadvance/specter-desktop/pull/2106/files#diff-4c264fd49854b2544ada9087b45610508c627ade4440754acd6825514a592100)


src/cryptoadvance/specter/managers/service_manager/service_manager.py
```diff
@@ -187,13 +191,36 @@ def inject_stuff():
...
+            "callbacks", callbacks.Callback, ext
+        )
+        for callback_class in classes:
+            logger.debug(f"  Loaded Callback {callback_class}")
+
+    @classmethod
+    def extract_thing_classes_from_extension(
+        cls, things: str, thing_class: type, ext
+    ) -> List[type]:
+        if hasattr(ext.__class__, things):
```
k9ert, line 218: Should have documentation

src/cryptoadvance/specter/managers/service_manager/callback_executor.py
```diff
@@ -0,0 +1,142 @@
...
+                no_incoming_edges.append(child)
+
+    if len(output) != len(class_map):
+        raise ValueError("Graph contains a cycle.")
+
+    return output
+
+
+def set_dependency_level_recursive(ext, level=0):
+    for dep in getattr(ext, "depends", []):
```
k9ert, line 140: Documentation please

src/cryptoadvance/specter/managers/service_manager/callback_executor.py
```diff
@@ -0,0 +1,142 @@
...
+        )
+        self._services_sorted = exts_sorted
+        return self._services_sorted
+
+    @property
+    def all_callbacks(self):
+        return get_subclasses(Callback)
+
+    def check_callback(self, callback, *args, **kwargs):
+        if type(callback) != type:
```
k9ert, line 86: Document!

src/cryptoadvance/specter/managers/service_manager.py
```diff
@@ -0,0 +1,437 @@
+import json
+import logging
+import os
+from importlib import import_module
+from inspect import isclass
+from pathlib import Path
+from pkgutil import iter_modules
+import sys
+from typing import Dict, List
```
k9ert, line 9: wrong!

src/cryptoadvance/specter/managers/service_manager/__init__.py
```diff
@@ -1 +1 @@
-from .service_manager import *
+from .service_manager import ExtensionManager
```
k9ert, line 1: is that really necessary?

k9ert: yes, commented.

src/cryptoadvance/specter/managers/service_manager.py
k9ert: deleted!

PR #2187, opened 2023-02-08
Chore: Better test of device name validity

moneymanolis:
Tests should (ideally) be idempotent (Bitcoin Core devices need a Bitcoin Core connection), especially if it is easy to do and they should be precise (directly checking for the patternMismatch - yes or no).

@Aditi-Singh16 please have a look and review the PR directly if you want.


Aditi-Singh16:
Oh okay, Sorry I didn't thought of it that way
However, here it works only when patternMismatch is set to be.false in both the cases
image

Aditi-Singh16:
```
cy.get('#device_name').invoke('prop', 'validity').its('valid').should('be.false')
cy.get("#device_name").clear().type("valid")
cy.get('#device_name').invoke('prop', 'validity').its('valid').should('be.true')
```

This one works

moneymanolis:
@Aditi-Singh16 I am not following completely, how did you get the error above. Locally and in the CI the tests are green ...

moneymanolis:
> ```
> cy.get('#device_name').invoke('prop', 'validity').its('valid').should('be.false')
> cy.get("#device_name").clear().type("valid")
> cy.get('#device_name').invoke('prop', 'validity').its('valid').should('be.true')
> ```
> 
> This one works

That also works! But, it would be a bit less specific.

k9ert (approved):

PR #2196, opened 2023-02-11
A GraphQL Extension

k9ert:
A GraphQL Extension.
image
```
{
  bookmarks {
    title
    desc    
    
  }
}
```
```
{
  users {
    username
    
  }
}
```


moneymanolis:
Closing this, since I don't see anyone picking this up again any time soon.

PR #2219, opened 2023-02-16
UIUX: Making Splashscreen work in electron Appimage

k9ert:
* The splash-screen tried to import resources which are not there in a packaged electron Appimage


moneymanolis (changes requested):
Please, let me win with my Spinner Solution :-)

PR #2223, opened 2023-02-16
Fix settings "Save" button

otkstdio:
This fixes the issue in preferences where the save button fails.


moneymanolis (approved):

k9ert:
Can you elaborate what the issue was and how it has been fixed?

moneymanolis:
Good summary von GPT:

The main process is listening for the "showMessageBoxSync" event using ipcMain.handle, and the renderer process can trigger this event by calling ipcRenderer.invoke('showMessageBoxSync', message, buttons). When the event is triggered, the callback function registered with ipcMain.handle is called, and the code in the main process can interact with the operating system to display the message box.

moneymanolis:
The implementation was fundamentally wrong I think @k9ert:
`dialog.showMessageBoxSync` was used before in `settings.html` which should be called from the main process and not from the renderer process. This is the issue that is fixed with this PR. 

GPT:
"dialog.showMessageBoxSync() is a method provided by the dialog module in the main process of an Electron application, and it cannot be called directly from a renderer process."

moneymanolis:
Regarding restructuring JS code, that's a good idea, but I would do this in separate PR and do this all in once.

k9ert (approved):

moneymanolis:
From the Electron documenation (supporting @k9ert point about separating the JS files):

grafik

PR #2234, opened 2023-02-20
More consistent tor-setup

k9ert:


k9ert:
* There was never an option to not use tor but at the same time it not has been properly setup from the start. So now we have three possible values for `tor_type`: `none`, `builtin` or `custom`. 
* `builtin` is now no longer chooseable in the setup screen if you use a pip-installation
* A bit better error-handling which should prevent that an issue like this blocks proper startup (from the chat):
```
  File "/home/specter/.local/lib/python3.9/site-packages/stem/connection.py", line 1107, in _read_cookie
    raise UnreadableCookieFile(exc_msg, cookie_path, is_safecookie)
stem.connection.UnreadableCookieFile: Authentication failed: unable to read '/run/tor/control.authcookie' ([Errno 13] Permission denied: '/run/tor/control.authcookie')
```

docs/architecture.md
```diff
@@ -41,4 +41,13 @@ The documentation here is incomplete but will better over time.
...
+
+## About Tor
+
+Tor in general means two things: 
+* Using the Tor network for outgoing connection to confuse outgoing traffic
+* Use Tor to create a tor-hidden-service which often enough is also used as a kind of poo-man's firewall and to avoid the details of a port-forwarding setup
+
+Tor support is possible either via a custom installation or via the built in. The type of the installation is stored in `specter.tor_type` which is redirecting to the `config_manager` which by default returns `builtin`. Another valid value is `custom` and initially we have `none`.
```
moneymanolis, line 52:
```suggestion
Tor support is possible either via a custom or via the built-in installation. The type of installation is stored in `specter.tor_type` which is redirecting to the `config_manager` which by default returns `builtin`. Another valid value is `custom` and initially we have `none`.
```

src/cryptoadvance/specter/server_endpoints/settings.py
```diff
@@ -237,6 +237,7 @@ def tor():
     only_tor = app.specter.only_tor
     tor_control_port = app.specter.tor_control_port
     tor_type = app.specter.tor_type
+    tor_builtin_possible = getattr(sys, "frozen", False) and hasattr(sys, "_MEIPASS")
```
moneymanolis, line 240: Comment would be nice! Sth. like
```suggestion
    # This is true if the Python interpreter has been bundled with the application into a single executable, so basically it is true for the apps but not for pip-installations
    tor_builtin_possible = getattr(sys, "frozen", False) and hasattr(sys, "_MEIPASS")
```

moneymanolis:
Shouldn't there be check whether the tor_controller attr exists like so:
```
    @property
    def tor_controller(self):
        if hasattr(self, "_tor_controller") and self._tor_controller:
            return self._tor_controller
        self.update_tor_controller()
        if hasattr(self, "_tor_controller") and self._tor_controller:
            return self._tor_controller
        raise SpecterError(
            "Failed to connect to the Tor daemon. Make sure ControlPort is properly configured."
        )
```

Without it, I get this when closing Specter:

```
...

 if app.specter.tor_controller is not None:
  File "/Users/heinzi/coding/repos/specter-desktop/src/cryptoadvance/specter/specter.py", line 453, in tor_controller
    if self._tor_controller:
AttributeError: 'Specter' object has no attribute '_tor_controller'. Did you mean: 'tor_control_port'?
```

moneymanolis:
Slight change in UI in https://github.com/cryptoadvance/specter-desktop/pull/2234/commits/dafe308f023d7b5626e86774c8180205d76e20c8:
grafik

moneymanolis (changes requested):
Some small suggestions and one question.

k9ert:
> Without it, I get this when closing Specter:

Absolutely. Changed it with [8a9b355](https://github.com/cryptoadvance/specter-desktop/pull/2234/commits/8a9b35501537e2ea3f47e1fea2359edc7358a6a1)

moneymanolis (approved):

PR #2243, opened 2023-02-21
Several UIUX fixes

moneymanolis:
This PR includes
- Make Spectrum progress bar dynamic
- "Tick checkboxes" for Spectrum, in particular without Whitepaper download option
- Run the numbers removed for Spectrum
- Toast message that Spectrum and Core are syncing only once per session


otkstdio (approved):
All works for me! Good to go

Issue #2248, opened 2023-02-23
v2.0-pre27: Log says: No rescan of the blockchain with Spectrum.


moritzwietersheim:
```
2023-02-23T16:06:38.355Z [info] : stderr-SPECTERD: INFO in spectrum: Script 3ddb49d is not synced None != 6692b22dc311ad45797e07cd638a7b5ec620f5f27c95a1c086bcbf0996d6b17f

2023-02-23T16:06:38.552Z [info] : stderr-SPECTERD: INFO in txlist: Cleared the Cache for /Users/moritzwietersheim/.specter/wallets/main/ms__kot_txs.csv (and rawdir)

2023-02-23T16:06:38.805Z [info] : stderr-SPECTERD: INFO in txlist: Cleared the Cache for /Users/moritzwietersheim/.specter/wallets/main/ms__kot_txs.csv (and rawdir)

2023-02-23T16:06:38.816Z [info] : stderr---- Logging error ---

2023-02-23T16:06:38.818Z [info] : stderr-Traceback (most recent call last):
  File "cryptoadvance/specter/util/wallet_importer.py", line 273, in rescan_as_needed
  File "cryptoadvance/specterext/spectrum/bridge_rpc.py", line 59, in rescanblockchain
cryptoadvance.specter.specter_error.SpecterError: No rescan of the blockchain with Spectrum.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "logging/__init__.py", line 1100, in emit
  File "logging/__init__.py", line 943, in format
  File "logging/__init__.py", line 678, in format
  File "logging/__init__.py", line 368, in getMessage
TypeError: not all arguments converted during string formatting
Call stack:
  File "threading.py", line 966, in _bootstrap
  File "threading.py", line 1009, in _bootstrap_inner
  File "threading.py", line 946, in run
  File "socketserver.py", line 683, in process_request_thread
  File "socketserver.py", line 360, in finish_request
  File "socketserver.py", line 747, in __init__
  File "werkzeug/serving.py", line 361, in handle
  File "http/server.py", line 425, in handle
  File "http/server.py", line 413, in handle_one_request
  File "werkzeug/serving.py", line 333, in run_wsgi
  File "werkzeug/serving.py", line 320, in execute
  File "flask/app.py", line 2095, in __call__
  File "werkzeug/middleware/proxy_fix.py", line 187, in __call__
  File "flask/app.py", line 2077, in wsgi_app
[...]

PR #2254, opened 2023-02-24
Chore: uiux/chore/show version

k9ert:
Show the version in the about page and on the error-page


k9ert:
addresses #2248 as well. At least the part which is not properly formatting the error.

moneymanolis (approved):

PR #2257, opened 2023-02-24
Feature: Progress bar for download of Specter binary in Electron

moneymanolis:
The download of the Specter binary (specterd) can take several minutes. To improve the UX of this lengthy process, this PR implements a progress bar of the download. 

### Architectural / UI choices:
- [This library](https://github.com/AndersonMamede/electron-progressbar) is used for the progress bar to have more options as the standard progress bars in Electron. 
- The progress bar instance is a child window of the main window. So this applies (from the documentation): "If informed, the progress bar window will always show on top of the parent window and will block it so user can't interact with it until the progress bar is completed/aborted and closed."

This makes a lot of sense for our use case, since the user cannot do anything else before specterd is downloaded.

This choice led to the need to implement a Force Quit menu item in the dock for MacOS, since the default Quit item cannot be changed or replaced and it doesn't trigger the "before-quit" event if a child window is still open on MacOS. The `Force Quit during download` item gives the (MacOS) user the opportunity to also quit the app from the dock. The item disappears again if the download is finished. 

Note: Main.js should be made much more modular and the download code should probably be seperated out to a dedicated file. This should be done in a different PR, though, since this feature should ideally be available asap.

#### Startup
https://user-images.githubusercontent.com/70536101/221206572-9ac62a91-4d83-4b8a-909b-d176b0a357d4.mov

#### Download finished
https://user-images.githubusercontent.com/70536101/221206886-15843a5f-fe48-41ae-86a0-08dec13283cd.mov

#### Dock
https://user-images.githubusercontent.com/70536101/221206925-79206379-6375-4885-89d9-51a2afc06798.mov

#### Logging of the download progress
Screenshot 2023-02-24 at 13 52 39


pyinstaller/electron/main.js
```diff
@@ -70,6 +64,11 @@ if (isMac && isDev) {
   app.dock.setIcon(dockIcon)
 }
 
+let appSettings = getAppSettings()
+let dimensions = { widIth: 1500, height: 1000 };
```
otkstdio, line 68: Typo on width here @moneymanolis

moneymanolis: You can use the "Add a suggestion" functionality for this little stuff.

otkstdio, line 68:
```suggestion
let dimensions = { width: 1500, height: 1000 };
```

k9ert (approved):

Issue #2260, opened 2023-02-24
Exception thrown when selecting device settings without a node connection


otkstdio:
An exception is thrown when attempting to add a device or change an existing device if you have no node connection.

To reproduce, remove your node / delete spectrum connection so it says "Disconnected" (also the state for a new user), click "Add device" and click on any signing device.

<img width="1329" alt="shot" src="https://user-images.githubusercontent.com/85369328/221285928-736a53ca-0dfc-4f85-a1b2-4e67212ae8c5.png">

alluringambiguous:
Hi, I am looking at the issue, I reproduced it.
I traced the error and found that there is no check written in the `/new_device_keys/<device_type>/` route. 
I can think of 2 possible solutions,
1. One could be to pop up error when ever user clicks on a device (when the node is not connected).
2.  Other could be to not allow user to add device by showing an warning (similar to how it is handled in case of wallets)

![66ca1e7e-6dff-4cff-b754-73a81eb97408](https://user-images.githubusercontent.com/67975841/221363702-ac5947c9-f806-4242-8183-879e64459b00.jpg)



Please let me know your feedback. :) 
@moneymanolis

PR #2266, opened 2023-02-27
Bugfix: Exception thrown when trying to add a device without node connection

moneymanolis:
Fixes https://github.com/cryptoadvance/specter-desktop/issues/2260

`NonExistingNode` gets a `network_parameter `property.

### DIFF
--- a/src/cryptoadvance/specter/node.py
+++ b/src/cryptoadvance/specter/node.py
@@ -65,6 +65,12 @@ class NonExistingNode(PersistentObject):
     def rpc(self):
         return None
 
+    @property
+    def network_parameters(self):
+        if self.is_running:
+            return get_network(self.chain)
+        return get_network("main")
+
     def check_blockheight(self):
         """check_blockheight is a method which is probably deprecated.
         It should return True if there are new blocks available since check_info has been called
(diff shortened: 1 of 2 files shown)


src/cryptoadvance/specter/node.py
```diff
@@ -65,6 +65,10 @@ def is_running(self):
     def rpc(self):
         return None
 
+    @property
+    def network_parameters(self):
+        return get_network("main")
```
k9ert, line 70:
```suggestion
    def network_parameters(self):
        ''' Is needed for creating devices/xpubs as they ask which network to use.
        So this is a convenient thing especially for prod
        '''
        return get_network("main")
```

src/cryptoadvance/specter/node.py
```diff
@@ -215,6 +215,12 @@ def is_running(self):
         else:
             return True
 
+    @property
+    def network_parameters(self):
+        if self.is_running:
```
k9ert, line 220:
```suggestion
    def network_parameters(self):
        '''
        helpful if the cache is not filled. Not good for regtest but let's optimize for prod.
        '''
        if self.is_running:
```

k9ert (approved):

Issue #2267, opened 2023-02-28
2.0.0-pre32: Specter doesn't shut down after 1st install and needs force quit


moritzwietersheim:
- 2.0.0-pre32: Specter doesn't shut down after 1st install and needs to be force quit.
- With the next second start the quitting problem disappears and Specter can be shut down without problems.
- Please see videos and logs below.

---
Core Node setup: 
https://user-images.githubusercontent.com/47259243/221754896-ce1eaaec-047e-46bc-88ea-a79095b9b5ec.mov

[23027 specterApp copy CoreNode .log](https://github.com/cryptoadvance/specter-desktop/files/10846606/23027.specterApp.copy.CoreNode.log)

Log: 
2023-02-28T03:54:08.979Z [info] : stderr-SPECTERD: INFO in controller: GET /nodes/sync_status/                      (200) took    3 ms

2023-02-28T03:54:08.998Z [info] : stderr-SPECTERD: INFO in controller: POST /wallets/wallet/singlesig_oven_2/txlist  (200) took   24 ms

2023-02-28T03:54:09.021Z [info] : stderr-SPECTERD: INFO in controller: GET /wallets/wallet/singlesig_oven_2/pending_psbt_list (200) took    0 ms

2023-02-28T03:54:51.360Z [info] : Quitting ...
2023-02-28T03:54:51.383Z [info] : stdout-[{'unit': 'sat', 'amount': 20000, 'btc_amount': 0.0002, 'recipient_id': 0, 'label': '', 'address': 'bc1qn50yv7f4j2l28244qzywkpyh389x3vnksu09df'}]
[{'unit': 'sat', 'amount': 20000, 'btc_amount': 0.0002, 'recipient_id': 0, 'label': 'test Jade T1', 'address': 'bc1qn50yv7f4j2l28244qzywkpyh389x3vnksu09df'}]
[{'unit': 'sat', 'amount': 10000, 'btc_amount': 0.0001, 'recipient_id': 0, 'label': 'test LNS', 'address': 'bc1q2lhamzv28y9xl0q6vpguzzw8m0wqmmtskt39hwnl0ywdlmrne56shrauxa'}]
[{'unit': 'sat', 'amount': 10000, 'btc_amount': 0.0001, 'recipient_id': 0, 'label': 'test LNS', 'address': 'bc1q2lhamzv28y9xl0q6vpguzzw8m0wqmmtskt39hwnl0ywdlmrne56shrauxa'}]

[...]

PR #2268, opened 2023-02-28
Bugfix: Specter app doesn't close after downloading

moneymanolis:
You can only destroy the progress bar if it hadn't been closed before which is done by default if the download is completed.

Fixes https://github.com/cryptoadvance/specter-desktop/issues/2267


pyinstaller/electron/version-data-dev.json
```diff
@@ -0,0 +1,5 @@
+{
+  "version": "v2.0.0-pre32",
+  "sha256": "aa049abf3e75199bad26fbded08ee5911ad48e325b42c43ec195136bd0736785"
+}
+
```
k9ert, line 5: Is it a good idea to check that in? Maybe it should be git-ignored?

PR #2271, opened 2023-02-28
Chore: Add flag to run elements in Cypress test script

moneymanolis:
- Elementsd is not run unless `--with-elements` flag is set
- `isElementsRunning` check in Cypress added to run Liquid tests only if elementsd is running.

Usage:  `./utils/test-cypress.sh --with-elements run`


moneymanolis:
--with-elements is not respected yet when given ...

k9ert (approved):

PR #2273, opened 2023-03-01
Bugfix/feature: More robust wallet recreation and support tool "Wallet management" for Spectrum

k9ert:
Currently, a specter Wallet is, technically, two folded: A Wallet on specter in form of a json-file and the backend stores the wallet, either core or spectrum.

There could be mismatches and it's difficult to figure out what the situation is. With spectrum, that responsibility is on our side. So this PR:
* Breaks with the "alias" for a wallet. A wallet will get the proper mapped alias and no magic number is attached anymore
* In order to be able to do that, we now support deletion of wallets on the spectrum side.
* this PR contains a tiny spectrum admin-backend tool in order to view and delete wallets on the spectrum side.


moneymanolis:
If we remove the unique alias, don't we need a check whether the name already exists - similar to how it is done with the node names?

moneymanolis:
Most concise way to see what `alias() `is doing now is to look at the enriched documentation in `helpers.py ` (see https://github.com/cryptoadvance/specter-desktop/pull/2273/commits/a76eedc7dcec46ed0c2aee1b5b4bf4c289ff34bc) and to have a look at this new pytest (see https://github.com/cryptoadvance/specter-desktop/pull/2273/commits/eff9c7b6c7f507af19672da4905ea50173ff7a73)

moneymanolis:
This title was added in the front end:

grafik

moneymanolis:
We need to find a better way to integrate the new (cool!) wallet management functionality in Spectrum. A "Manage" button in screenshot 1 in the connections screen seems misplaced (since it suggests the ability to manage the connection further even though it is about managing the wallets). 

grafik

Clicking on Manage goes to:

grafik

moneymanolis:
Introduction of Support Tools in Settings with this PR:

grafik

moneymanolis (approved):

PR #2297, opened 2023-03-09
Chore: refactoring HWI

k9ert:
This PR should change the architecture of how HWI is used. It should no longer be used as a library with dependencies but as an executable. Steps to that goal are:
* separating the HWI functionality into its own core-extension


k9ert:
superseded by #2299

PR #2299, opened 2023-03-09
Chore: Rearchitect HWI

k9ert:
# Design

This PR should change the architecture of how HWI is used. It should no longer be used as a library with dependencies but as an executable. For that this PR does separate the HWI functionality into its own core-extension and refactor the HWIBridge to have now this structure:
[](https://mermaid-js.github.io/mermaid-live-editor/edit#pako:eNqlkc9qwzAMxl_F-NSy9AXCLvsHK2ynHQYjYBRbSQWxHWyZtnR596Vxk7JedpgvsqRPv89GJ6m9QVlK3UGMzwRtAFs5MZ6HOnIAza-f28dApkVx_73ZiHNKDsIxF__SvlE9C7N0MrqlnHJPCIWHCaQOfaqjaoK3SneEjlc5FAK09snxOo8MN9TF71_IX9A95ac-TeqFe3f97G5Pc7VFVhYiY1ANuRZDH2h0Ws993fmIq6uTLKTFYIHMuISJXUneocVKluPVYAOp40pWbhilqTfA-GKIfZBlA13EQkJi_3F0WpYcEs6iyy4XVQ_uy_trjhPk_bL9cxh-AMrUsRI)

Common code is in AbstractHWIBridge. The old implementation goes into the `Lib`, the new one in the `Binary`. The HWIlib implementation has the idea of a `client` which is used as a parameter in that notable function `_extract_xpubs_from_client(client, account)`. In order to reuse that method, i've created a `HwiBinaryClient` which behaves and has the methods of a hwilib client including a dummy `close()` method.

## Testing

The testing has improved seriously. As a tester, you need to attach the HWW and does a:
```
pytest tests/test_hwi_rpc_trezor.py -m manual -s -vvv
```
The `m manual` is needed as those tests will be omitted by default (for CI). The tests will unlock the trezor if needed (tester has to enter pin) and togglepassphrase (tester has to confirm).
All tests are in sequence run against both implementations with the very same asserts. That was quite helpful with implementing a second impl.

# Usage
There is a new Config-key called `HWI_RPC_IMPL` which is either `lib' or `bin`. By default in the BaseConfig it is :
```
HWI_RPC_IMPL = os.getenv("HWI_RPC_IMPL", "bin")
```

So anyone can change it by env-var any time. We're leveraging that in the `pyinstaller/main.js` where we're setting `lib` as env-var which means that inside specterd/electron, still the "old" lib implementation should be relevant.


bitcoinlizard:
I wanted to make you aware of one thing related to HWI and this refactor that you may or may not know about.  Specter Desktop doesn't work with recent versions of HWI starting with version 2.3.0.  Perhaps it would be good to bring Specter Desktop in line with recent versions of HWI?  This is the error when running Specter Desktop 2.0.2 with HWI 2.3.0 or newer:

```
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/usr/lib/python3.11/site-packages/cryptoadvance/specter/__main__.py", line 2, in <module>
    from .cli import entry_point
  File "/usr/lib/python3.11/site-packages/cryptoadvance/specter/cli/__init__.py", line 6, in <module>
    from ..server import setup_logging
  File "/usr/lib/python3.11/site-packages/cryptoadvance/specter/server.py", line 23, in <module>
    from cryptoadvance.specter.hwi_rpc import HWIBridge
  File "/usr/lib/python3.11/site-packages/cryptoadvance/specter/hwi_rpc.py", line 42, in <module>
    class BitBox02NoiseConfig(bitbox02.util.BitBoxAppNoiseConfig):
                              ^^^^^^^^^^^^^
AttributeError: module 'bitbox02' has no attribute 'util'
```

One change of note with HWI 2.3.0 and newer which may be relevant as bitbox02 is mentioned in the error message.  The HWI developers have made bitbox02 like the other hardware wallets.  It is now modular rather than a part of the base code and is no longer a required dependency.

k9ert:
No longer necessary.

bitcoinlizard:
Running hwi version 3.0.0 and Specter Desktop 2.0.5 I still receive the same error as I noted above.  It appears to me that Specter Desktop will only run with hwi versions 2.2.1 and older.

Issue #1957, opened 2022-11-04
Extension-generation needs adjustements 


k9ert:
The extension-generation process is currently assuming you need encrypted userdata. That should get adjusted like this:
- Add an option to the click annotations like `--encrypted-userdata` here: https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/cli/cli_ext.py#L39 (including adding it as a param to the method signature)
- Add a section explaining what it means and that the extension should decide between encrypted/unencrypted here: https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/cli/cli_ext.py#L111-L112
- Add the variable to the `ExtGen` instantiation and there ...
- The service and the controller would needs some adjustments. They are rendered here: https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/cli/cli_ext.py#L150-L160 however, the code for the templates is living here:
  - https://github.com/cryptoadvance/specterext-dummy/blob/master/src/dummyorg/specterext/dummy/service.py#L12
  - https://github.com/cryptoadvance/specterext-dummy/blob/master/src/dummyorg/specterext/dummy/controller.py#L53
  
Mind the different templating syntax. Those are "metatemplates" as they are creating templates which themself are later templates.

You can test it via using `--tmpl-fs-source` and checkout the dummy-project somewhere. Also i've created a small section about https://docs.specter.solutions/desktop/extensions/intro/#virtualenv-management

k9ert:
One other thing here would be that the templates are becoming more dependent on the version of specter controlling the generation-process. It would be cool if we could get another cli-switch which would specify the version of specter, simply assuming that the dummy-repo has tags with that version (and maybe falling back to master if it doesn't).
This is already established in parts: https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/services/extension_gen.py#L41 but need to be parametrized and maybe the fallback-solution need to be implemented.

Aditi-Singh16:
Hi @k9ert, I have a confusion... 
You said that "The extension-generation process is currently assuming you need encrypted userdata."
could you please explain what you meant by this?

Like what data is encrypted and why encrypted data?
Also when I created the extension, I didn't realize this... could you point out where is the indication of this assumption?

k9ert:
[...]

PR #2300, opened 2023-03-10
Chore: Extension generation improvements

Aditi-Singh16:
Fixes for #1957


Aditi-Singh16:
@k9ert , could you please take a look?

.cirrus.yml
```diff
@@ -109,7 +109,7 @@ extension_smoketest_task:
     - mkdir tmp && cd tmp
     - mkdir testextension && cd testextension
     - pwd
-    - python3 -m cryptoadvance.specter ext gen --ext-id cicdtest --org cryptoadvance --no-isolated-client --devicename cicddevice
+    - python3 -m cryptoadvance.specter ext gen --ext-id cicdtest --org cryptoadvance --no-isolated-client --no-encrypted-userdata --devicename cicddevice --version v2.0.0
```
k9ert, line 112: I think the version should be omitted here.
```suggestion
    - python3 -m cryptoadvance.specter ext gen --ext-id cicdtest --org cryptoadvance --no-isolated-client --no-encrypted-userdata --devicename cicddevice
```
We want to know whether the current branch this is currently being tested, works on the master of `specterext-dummy`. That might be not optimal in some cases but the best we can get for most of the cases where we have chages in a PR which (hopefully) still match the state in master of `specterext-dummy`.

src/cryptoadvance/specter/util/version.py
```diff
@@ -59,6 +59,37 @@ def loop(self, dt=3600):
             )
             time.sleep(dt)
 
+    def _check_if_version_is_available(self, custom_version):
```
k9ert, line 62: This check is checking on pypi. But the relevant check is the tag/branch in the specter-ext repo:
https://github.com/cryptoadvance/specterext-dummy/tags

This is what needs to be used when someone is referring to a tag. So can you change that to check whether the tag/branch is existing? You can also put that check as a class-method in the `GithubUrlLoader`, It doesn't need to be in the util package.

src/cryptoadvance/specter/services/extension_gen.py
```diff
@@ -46,13 +48,20 @@ def __init__(
         self.base_path = base_path
         self.org = ext_org
         self.id = ext_id
+        self.encrypted_userdata = encrypted_userdata
+        self.tag = tag
```
k9ert, line 52: You introduce here `self.tag` which is a better name than `self.branch`. Can you remove the `self.branch` then and adjust the usage? E.g. the `GithubUrlLoader` currently has a parameter called "branch" but it should be called `tag` (and initialized with your `self.tag`.

k9ert (changes requested):

k9ert:
In https://github.com/cryptoadvance/specter-desktop/issues/1957 i also talk about necessary changes in the template-directory:

> The service and the controller would needs some adjustments. They are rendered here: https://github.com/cryptoadvance/specter-desktop/blob/master/src/cryptoadvance/specter/cli/cli_ext.py#L150-L160 however, the code for the templates is living here:
> 
>     https://github.com/cryptoadvance/specterext-dummy/blob/master/src/dummyorg/specterext/dummy/service.py#L12
>     https://github.com/cryptoadvance/specterext-dummy/blob/master/src/dummyorg/specterext/dummy/controller.py#L53

That would need to be done as well. Please create PR there.

src/cryptoadvance/specter/util/version.py
Aditi-Singh16: Okay will do that

src/cryptoadvance/specter/cli/cli_ext.py
```diff
@@ -37,6 +38,12 @@ def ext():
 @ext.command()
 @click.option("--org", "org", default=None, help="Use a specific organsiation")
 @click.option("--ext-id", "ext_id", default=None, help="Use a specific extension id")
+@click.option(
+    "--encrypted-userdata/--no-encrypted-userdata",
+    default=None,
+    help="Whether the user data should be encryted or not",
+)
+@click.option("--version", "tag", default=None, help="Use a specific specter version")
```
Aditi-Singh16, line 46: So should I also remove the option for the user to specify the version while setting up an extension?

.cirrus.yml
Aditi-Singh16: the smoke test fails if I remove this

Aditi-Singh16: Okay, I read the issue again and I think, we need a version since templates are becoming more dependent on the version of Specter controlling the generation process and if no version is specified, it should fall back to master.

If the user chooses to add a specific version, it will render extension accordingly, else the tag(branch) will be master and version will be the latest one.

moneymanolis:
Closing due to inactivity.

Issue #2279, opened 2023-03-03
v2.0.0-pre33: Tor Settings seem not to stick and seems to impact price feature


moritzwietersheim:
- I went through the Tor functionality on a Electrum setup. 
- The Tor settings seem to jump back to "disabled" and not stay on "Built in"
- Only stays on "Built in" if "Force all external calls over Tor" is selected.

- Also the Price feature does strange things on Tor. 
- Doesnt work on Bitstamp, but on Coindesk, but jumps back to Bitstamp.

- Logs attached below.
- If higher resolution video is needed, pls let me know.

https://user-images.githubusercontent.com/47259243/222628383-edcca59b-5266-44cb-b77d-622b766cd5a1.mov

[230302 1101 specterApp copy.log](https://github.com/cryptoadvance/specter-desktop/files/10878057/230302.1101.specterApp.copy.log)

k9ert:
yes, bitstamp is known to not work with core. We need to put a warning in there and then also switch it off as price-provider.
https://github.com/cryptoadvance/specter-desktop/issues/1554

Issue #1554, opened 2022-02-03
Price not updating: Failed to fetch data...


BTCBellyButton:
No matter who I select as a price provider it stopped working (even before the latest 1.8.1 release).

![Screenshot from 2022-02-03 16-03-14](https://user-images.githubusercontent.com/68573637/152437003-b4153228-1985-43cd-a9ac-f3d4b06bdd15.png)

k9ert:
I just tested it via specterd on Linux and it works there. What platform are you running on?

BTCBellyButton:
@k9ert Linux Ubuntu 20.04.

BTCBellyButton:
![Screenshot from 2022-02-04 14-18-39](https://user-images.githubusercontent.com/68573637/152597607-bff4cf16-6ad0-49b5-96b0-aab53d76daa7.png)

BTCBellyButton:
[2022-02-04 14:41:44,889] INFO in checker: Checker price started with period 600
[2022-02-04 14:41:47,525] WARNING in price_providers: Failed to get price data from: spotbit_kraken. Exception: SOCKSHTTPConnectionPool(host='h6zwwkcivy2hjys6xpinlnz2f74dsmvltzsd4xb42vinhlcaoe7fdeqd.onion', port=80): Max retries exceeded with url: /now/usd/kraken (Caused by NewConnectionError('<urllib3.contrib.socks.SOCKSConnection object at 0x7ff5a460b760>: Failed to establish a new connection: 0x01: General SOCKS server failure'))
[2022-02-04 14:41:47,526] INFO in checker: Checker executed within 2.637 seconds. This message won't show again until stopped and started.
[2022-02-04 14:41:49,045] WARNING in price_providers: Failed to get price data from: spotbit_kraken. Exception: SOCKSHTTPConnectionPool(host='h6zwwkcivy2hjys6xpinlnz2f74dsmvltzsd4xb42vinhlcaoe7fdeqd.onion', port=80): Max retries exceeded with url: /now/usd/kraken (Caused by NewConnectionError('<urllib3.contrib.socks.SOCKSConnection object at 0x7ff5a460b250>: Failed to establish a new connection: 0x01: General SOCKS server failure'))
[2022-02-04 14:41:49,045] INFO in _internal: 192.168.1.22 - - [04/Feb/2022 14:41:49] "POST /price/update/ HTTP/1.1" 200 -

moneymanolis:
Looks like a Tor issue.

[...]

PR #2304, opened 2023-03-14
Bugfix/UIUX: Fix Tor configuration issues and improve UX of built-in Tor

moneymanolis:
- Fixes https://github.com/cryptoadvance/specter-desktop/issues/2279 - Radio box of the built-in configuration is correctly checked now (temporary usage of localStorage for that)
- Fixes `is_running()` method of the TorDaemonController (didn't work reliably)
- Fixes `stop_tor_daemon()` method of the same class (didn't work at all). The Tor process was never shut down when this function was called. The Tor process only shut down when Specter and thus the whole python process (as the parent process) were shut down.
- Additional functionality: Remove the built-in Tor setup if the user has it and is checking the "Disabled" radio button (second screencast)
- Unnecessary template code removed in `setup_page.jinja` which was only necessary for setting up an internal Bitcoin node
- Styling / design

### Complete Flow 
https://user-images.githubusercontent.com/70536101/225105291-091b6b77-7b5b-458e-9e3c-de35947b2e83.mov

### Checking the Disabled Button if having a built-in Tor setup
https://user-images.githubusercontent.com/70536101/225105583-0010158e-1566-42ea-9c16-0333fede7375.mov


src/cryptoadvance/specter/server_endpoints/settings.py
```diff
@@ -387,6 +397,15 @@ def tor():
     )
 
 
+@settings_endpoint.route("/update_tor_type", methods=["POST"])
+@login_required
+def update_tor_type():
+    tor_type = request.form["torType"]
+    app.specter.update_tor_type(tor_type, current_user)
+    response = {"torTypeChanged": True, "error": None}
```
k9ert, line 405: I don't like `"error": None`. Did we introduced that as a pattern at some time? I hope not. Would prefer to simply not have that key in case on not having an error.

src/cryptoadvance/specter/static/output.css
```diff
@@ -1,5 +1,5 @@
 /*
-! tailwindcss v3.2.7 | MIT License | https://tailwindcss.com
+! tailwindcss v3.2.4 | MIT License | https://tailwindcss.com
```
k9ert, line 2: Did we do a rollback of that version for some reason?

src/cryptoadvance/specter/templates/settings/tor_settings.jinja
```diff
@@ -15,37 +15,40 @@
...
 			<span class="sr-only">Info</span>
 			<div>
-				{{ _("This anonymizes connections to block explorers, price providers and Electrum servers if Tor is activated here.") }}
+				{{ _("This anonymizes connections to block explorers and price providers (if you choose the Tor options for them) if Tor is activated here. The connection to Electrum servers is automatically via Tor if you have Tor enabled.") }}
```
k9ert, line 18: I'm surprised as i was not aware that this description is indeed correct with the current status-quo:
https://github.com/cryptoadvance/specter-desktop/blame/master/src/cryptoadvance/specter/util/price_providers.py#L128-L130

It doesn't make that much sense, though:
* Tor is a prerequisite to use spotbit-onion-services but you can also use to and access clearnet price-services
* There might be issues with bitstamp as it's not accepting requests from the tor-network. I thought that this is currently an issue (see #1554) but it shouldn't with the code i posted above.

I would suggest to leave it as is for now. So this comment was just for clarification.

src/cryptoadvance/specter/tor_daemon.py
```diff
@@ -65,13 +65,24 @@ def get_hashed_password(self, password):
...
-        # if not self.tor_daemon_proc and self.is_port_open():
-        #    raise SpecterError(
-        #        "Port 9050 is open but tor_daemon_proc is not existing. Probably another Tor-Daemon is running?!"
-        #    )
-        return self.tor_daemon_proc and self.tor_daemon_proc.poll() is None
+        """Checks whether the Tor process is still running.
+        Note: poll() from the subprocess module does not work reliably. For example,
+        if the process has been terminated but its exit code has not yet been collected, it would still indicate
+        that the process is still running.
+        """
```
k9ert, line 72: I'm not that convinced about these arguments. Are there open issues we can refer to? Or StackOverflow discussions?
What is meant by: "if the process has been terminated but its exit code has not yet been collected, it would still indicate".
The [API-docs](https://docs.python.org/3/library/subprocess.html#subprocess.Popen) read like:
"Popen.poll() : Check if child process has terminated. Set and return [returncode](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.returncode) attribute. Otherwise, returns None."

src/cryptoadvance/specter/tor_daemon.py
```diff
@@ -65,13 +65,24 @@ def get_hashed_password(self, password):
...
+        if self.tor_daemon_proc is None:
+            return False
+        try:
+            # Note: This pid here is usually not the same as the pid given by the OS to the Tor process
+            process = psutil.Process(self.tor_daemon_proc.pid)
+            logger.debug(f"Is the built-in Tor daemon running? {process.is_running()}")
+            if process.is_running():
+                return True
```
k9ert, line 80:
```suggestion
            return process.is_running():
```

src/cryptoadvance/specter/tor_daemon.py
```diff
@@ -84,15 +95,15 @@ def is_port_open(self):
             return False
 
     def stop_tor_daemon(self):
-        timeout = 50  # in secs
+        # It is possible for the Tor process to terminate unexpectedly without updating self.tor_daemon_proc
+        if not self.is_running():
+            return
         if self.tor_daemon_proc:
+            # This double approach ensures that the Tor child process is terminated on all levels (within Specter and on the OS level)
```
k9ert, line 102: I don't understand those two levels. What means "within Specter"? And what's wrong with the below code?

moneymanolis: Yeah, granted, could very well just be on MacOS. I only tested this on MacOS. I will include that in the comment.

src/cryptoadvance/specter/tor_daemon.py
moneymanolis: Well, the open issue was, that it didn't work reliably for me during testing and the above works like a charm.
"If the process has been terminated but its exit code has not yet been collected, it would still indicate ... " means that the poll method would indicate in this instance that the process is still running when in reality it is not (which was the case during testing).

src/cryptoadvance/specter/server_endpoints/settings.py
moneymanolis: It is not needed, will remove it.

src/cryptoadvance/specter/tor_daemon.py
moneymanolis: Sure, why not.

src/cryptoadvance/specter/tor_daemon.py
moneymanolis: Basically, this is "within Specter": `self.tor_daemon_proc.terminate()` and this on the OS level: `subprocess.run(["pkill", "-f", cmdline_args])`

src/cryptoadvance/specter/templates/settings/tor_settings.jinja
moneymanolis: Copy changed in https://github.com/cryptoadvance/specter-desktop/pull/2304/commits/66a513683ac5f71a537ce68c2f876f08578315da

moneymanolis:
Most notable updates / additions:

- Only save the custom Tor setup if it is actually working:
https://user-images.githubusercontent.com/70536101/225658457-aacc26a0-3ff1-4076-b6fe-78ee78d663fd.mov

- Make clear to the user that leaving the control port blank is fine and make the design of the floating label nice:
https://user-images.githubusercontent.com/70536101/225657535-3260ad93-db5b-45f0-bf38-e8ef3e9dabf5.mov

- Allow the option Enforce Tor only if the Tor setup is not "disabled": 
https://user-images.githubusercontent.com/70536101/225658403-2048ac83-0b0d-4cb4-81cd-a6e00bb656f7.mov

k9ert (approved):

PR #2308, opened 2023-03-15
Bugfix: Keyerror in case of frozen utxos

### LINKED ISSUE #2253: Multiple outputs in a single transaction are smashed into one UTXO
- I found a bit of an exotic bug when a singlesig wallet receives to multiple addresses in one transaction. 
- These UTXOs are not displayed separately in the UI, but as only one large UTXO. And I could not send these UTXOs.
- This happened bc i send funds between my singlesig and my multisig back & forth, and i tried to do 2 batched transactions from multisig to multiple addresses of the singlesig wallet.
- Tx Multiple UTXOs 01:
 https://mempool.space/tx/30230e3cc5aa6508f80bec50aca95d5424b864ea4730a910025e912303490683#flow=&vout=0 
- Tx Multiple UTXOs 02: https://mempool.space/tx/295596865396117a22fa979bec08285b7ba2128a6fad77f469590c3302c45889

Screenshot 2023-02-23 at 11 41 07 PM

Screenshot 2023-02-23 at 11 41 47 PM

k9ert:
As the stupid listlockunspent call does not by far contains as much data for the utxos than the listunspent call, we need to make an extra-call to get that data.

Also the check whether a utxo was frozen missed the differenziation with vouts. 

fixes #2253

### DIFF
--- a/src/cryptoadvance/specter/wallet.py
+++ b/src/cryptoadvance/specter/wallet.py
@@ -848,16 +848,11 @@ class Wallet:
             #       {'txid': '7cdc6c668b665c47de5823caca41c194f2d70815420c564aa4fa5786d4c0693f', 'vout': 0}
             # ]
 
-            # an easy searchable list:
-            locked_utxo_list = [f"{tx['txid']}:{tx['vout']}" for tx in locked_utxo]
-            utxo = self.rpc.listunspent(0)
-            utxo.extend(
-                locked_utxo
-            )  # It's a bit of a unequal list as the locked_utxo does not contain much
+            # an easy searchable dict for locked_utxos
+            # having the txid:vout as key
+            locked_utxo_dict = {f"{tx['txid']}:{tx['vout']}": {} for tx in locked_utxo}
 
-            txlist = self._transactions.get_transactions()
-            txlist_dict = {_tx["txid"]: _tx for _tx in txlist}
-
+            utxos = self.rpc.listunspent(0)
             # Example of utxo
             # [
             #     {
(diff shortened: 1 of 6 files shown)


src/cryptoadvance/specter/wallet.py
```diff
@@ -915,7 +919,14 @@ def check_utxo(self):
...
                                 and utxo_item["vout"] == utxo_vout
                             ):
-                                tx_copy["amount"] = utxo_item["amount"]
+                                if utxo_item.get("amount"):
+                                    amount = utxo_item.get("amount")
+                                else:
+                                    # in the case of locked amounts, the stupid listlockunspent call does not contain reasonable utxo-data
+                                    amount = self.rpc.gettransaction(tx_copy["txid"])[
+                                        "details"
+                                    ][utxo_item["vout"]]["amount"]
```
k9ert, line 928: The vout can not be used as index as vout is a key in the elements of the details array. See:
https://developer.bitcoin.org/reference/rpc/gettransaction.html?highlight=gettransaction

src/cryptoadvance/specter/wallet.py
```diff
@@ -915,7 +919,14 @@ def check_utxo(self):
...
+                                tx_copy["amount"] = amount
```
k9ert, line 929: We also need to set the address of this "utxo" as the address which is already set is coming from the tx-context and might not be the right one.

src/cryptoadvance/specter/wallet.py
k9ert: It most probably can be used as an index. This will hopefully soon be clarified here: https://bitcoin.stackexchange.com/questions/117597/order-of-the-details-array-in-the-output-of-the-gettransaction-wallet-api-call

moneymanolis:
#### Given those UTXOs
grafik

### Address view
#### Master
grafik

#### This PR
grafik

### Frozen UTXO in a TX with multiple outputs
#### Master
Key Error "amount"

#### This PR
grafik

moneymanolis (approved):
Great job!

PR #2309, opened 2023-03-17
Swan plugin: New design, improved UX and bug fixes

moneymanolis:
#### Most important points
- Overall new Specter v2 look for the Swan plugin (see second comment for visuals)
- UTXO view removed on plugin withdrawals overview. I don't know why we had the view there in the first place (and it didn't filter by service ids). All the transactions of the withdrawals are by definition UTXOs since they come from Swan withdrawals.
- Allowed hostnames reduced to localhost to match the current state on the Swan backend. We will adapt this as soon as the backend allows it. Special feedback for Umbrel users with a hint on how to make the plugin work by running Specter locally and connecting to Umbrel electrs.
- Better UX in settings overview (e.g. mentioning the email confirmation of the wallet on the Swan side ...)
- Info about the current working solution (if the user is already signed in, linking of the account works, if not the oauth2 flow breaks when accepting the scope of the Specter client)

#### Misc
- Dev helper styled
- Associate with service styled
- Remarks on how to use USB signer with a remote Specter changed and styled


moneymanolis:
#### Start
swan-plugin-start

#### Working solution info
Screenshot 2023-03-17 at 20 33 52

#### No auto-withdrawal plan yet
grafik

#### With auto-withdrawal plan yet
Screenshot 2023-03-17 at 20 43 03

#### Withdrawals overview
swan-plugin-tx-overview

#### Special feedback for Umbrel
Screenshot 2023-03-17 at 20 32 19

kdmukai:
grammar fix!

"It looks like your using Specter on Umbrel" -> "It looks like you're using Specter on Umbrel."

moneymanolis:
> TODO: How to best handle the situation if there is no API connection when removing the plugin and calling `remove_swan_integration()`. This currently swallows any API exceptions, so that the user thinks that Swan will no longer use the addresses from Specter when in reality it would.

Solved in https://github.com/cryptoadvance/specter-desktop/pull/2309/commits/7c50994fc4eeff6b4b4790c1cd857496245cd3e9

#### Visual
https://user-images.githubusercontent.com/70536101/226426007-44f1f61f-8a52-4efe-a788-7b68f4a20271.mov

k9ert (approved):

PR #2347, opened 2023-06-26
Chore: Use prettier for Electron app 

moneymanolis:
Also in this PR: 
- Change copy for multisig address verification
- Add package-lock to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ pyinstaller/version.txt
 pyinstaller/electron/version-data.json
 pyinstaller/electron/downloadloc.js
 pyinstaller/electron/node_modules
+pyinstaller/electron/package-lock.json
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -23,6 +23,7 @@ pyinstaller/version.txt
 pyinstaller/electron/version-data.json
 pyinstaller/electron/downloadloc.js
 pyinstaller/electron/node_modules
+pyinstaller/electron/package-lock.json
```
k9ert, line 26: Shouldn't that file not get checked in?

k9ert (approved):

.gitignore
moneymanolis: Yes, that's why I'am adding it to `.gitignore` here ...

PR #2348, opened 2023-06-27
Feature: Enable import of a multisig wallet that uses a multi-descriptor

moneymanolis:
- Besides copy changes to reflect the additional functionality, this PR mainly introduces a `keep_multi` parameter to ensure that an imported multisig wallet with a `multi` descriptor is not changing to a `sortedmulti` descriptor as is currently the case.
- New multisig wallets created with Specter still use `sortedmulti`


moneymanolis:
Closing since we went for https://github.com/cryptoadvance/specter-desktop/pull/2349

PR #2354, opened 2023-07-14
Remove whitespace from wallet importer to fix importing via copy-paste from pdf

otkstudio:
When importing your wallet by copy and pasting the descriptor from the Specter backup pdf it fails as copying from a pdf leaves whitespaces in the string. This is rejected by specter because of the whitespace. This PR strips any whitespaces from the import string before importing.

### DIFF
--- a/src/cryptoadvance/specter/util/wallet_importer.py
+++ b/src/cryptoadvance/specter/util/wallet_importer.py
@@ -38,7 +38,7 @@ class WalletImporter:
         if device_manager is None:
             device_manager = specter.device_manager
         try:
-            self.wallet_data = json.loads(wallet_json)
+            self.wallet_data = json.loads("".join(wallet_json.split()))
             (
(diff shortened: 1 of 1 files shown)


src/cryptoadvance/specter/util/wallet_importer.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, wallet_json, specter, device_manager=None):
         if device_manager is None:
             device_manager = specter.device_manager
         try:
-            self.wallet_data = json.loads(wallet_json)
+            self.wallet_data = json.loads("".join(wallet_json.split()))
```
k9ert, line 41: Maybe we can do it a bit more concise:
```suggestion
            self.wallet_data = json.loads(wallet_json.replace(" ", ""))
```

k9ert:
Closing this as it's fixed in #2355

Issue #2356, opened 2023-07-20
Mac signature no longer accepted


k9ert:
Starting the signed binary like this:
```
(.env) ➜  specter-desktop git:(kn/mac_signing) ✗ ./pyinstaller/electron/dist/mac-arm64/Specter.app/Contents/MacOS/Specter
[1]    45152 killed     ./pyinstaller/electron/dist/mac-arm64/Specter.app/Contents/MacOS/Specter
(.env) ➜  specter-desktop git:(kn/mac_signing) ✗
```
So the executable will get killed immediately and you'll see something like this in the logs:
```
Disallowing solutions.specter.desktop because no eligible provisioning profiles found
```

It seems that we need a provisioning profile. However, Apple is not mentioning that at it's [documentation](https://developer.apple.com/developer-id/). There, only signing and notarizing is mentioned.

## validating signatures
Is the signature of that binary even valid? A prerequisite for notarizing is that you sign your software with your developer key. 
<details>
<summary>The signature can be checked like this:</summary>

```
➜  specter-desktop git:(kn/mac_signing) ✗ pkgutil --check-signature ./pyinstaller/electron/dist/mac-arm64/Specter.app
Package "Specter":
   Status: signed by a certificate trusted by macOS
   Certificate Chain:
    1. Developer ID Application: Kim Neunert (FWV59JHV83)
       Expires: 2026-09-11 11:59:39 +0000
       SHA256 Fingerprint:
           06 C7 63 8C 92 5B DD 60 79 8C B1 B0 30 8D B7 98 2C 99 8E F3 33 87 
           A0 BD 03 1B 35 C5 3D 53 3C 3D
       ------------------------------------------------------------------------
    2. Developer ID Certification Authority
       Expires: 2027-02-01 22:12:15 +0000
       SHA256 Fingerprint:
           7A FC 9D 01 A6 2F 03 A2 DE 96 37 93 6D 4A FE 68 09 0D 2D E1 8D 03 
           F2 9C 88 CF B0 B1 BA 63 58 7F
       ------------------------------------------------------------------------
    3. Apple Root CA
       Expires: 2035-02-09 21:40:36 +0000
       SHA256 Fingerprint:
           B0 B1 73 0E CB C7 FF 45 05 14 2C 49 F1 29 5E 6E DA 6B CA ED 7E 2C 
           68 C5 BE 91 B5 A1 10 01 F0 24

➜  specter-desktop git:(kn/mac_signing) ✗ 
```
[...]

PR #2358, opened 2023-07-21
Remove Specter URI to avoid provisioning profiles (Apple)

k9ert:
This addition in the entitlements-file caused #2356 . So let's remove it for now maybe adding it at a later stage.


moneymanolis (approved):
and ... (all the work) is gone.

k9ert:
Let's wait for the new key-infrastructure and we can reestablish this.

Issue #2343, opened 2023-06-21
Specter does not start if Jade is plugged in and the serial port is still open


moneymanolis:
Below is the stack trace. The error disappeared after running in the browser:
```
ports = await navigator.serial.getPorts()
await ports[0].close()
```

We'd need to implement a check for open ports and close them if there are any in Python. Sth. like this:

```
import serial.tools.list_ports as list_ports
ports = list(list_ports.comports())
if len(ports) > 0:
    port = ports[0].device
    ser = serial.Serial(port)
    ser.close()
```

Stack trace:

```
[2023-06-21 17:57:05,491] DEBUG in jade: Failed to connect to Jade simulator at tcp:127.0.0.1:2222
[2023-06-21 17:57:05,491] DEBUG in jade: [Errno 61] Connection refused
Traceback (most recent call last):
  File "/Users/manolis/.pyenv/versions/.specter/lib/python3.10/site-packages/serial/serialposix.py", line 322, in open
    self.fd = os.open(self.portstr, os.O_RDWR | os.O_NOCTTY | os.O_NONBLOCK)
OSError: [Errno 16] Resource busy: '/dev/cu.usbserial-024346F8'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/manolis/.pyenv/versions/3.10.0/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Users/manolis/.pyenv/versions/3.10.0/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/Users/manolis/coding/repos/specter-desktop/src/cryptoadvance/specter/__main__.py", line 7, in <module>
    entry_point()
  File "/Users/manolis/.pyenv/versions/.specter/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/Users/manolis/.pyenv/versions/.specter/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/Users/manolis/.pyenv/versions/.specter/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/manolis/.pyenv/versions/.specter/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
[...]

PR #2371, opened 2023-08-26
remove enumerate in HWIBridge init

stepansnigirev:
Removes initial enumeration of the devices before Specter starts.

Close https://github.com/cryptoadvance/specter-desktop/issues/2343

Requires careful testing on Mac and Windows (tested on Linux, seems to work fine).

Test process for Jade, Ledger and Trezor:
- start Specter with connected but locked hardware wallet
- unlock and go through some HWI functions - add keys, display address, send tx or whatever
- start Specter without any HW connected
- connect and use HWI functions


k9ert (approved):

moneymanolis:
`display address` doesn't work for me anymore with the change (testing with Jade).

PR #2421, opened 2024-03-13
Fix Jade signing issues with Swan Vault

### LINKED ISSUE #2402: Receive script cannot be validated when signing transaction with Jade device
**Describe the bug**
Cannot send a specified amount of BTC from my MultiSig wallet with Jade signer. This issue does not impact transfers where the entire BTC amount is swept (i.e Send Max)

**To Reproduce**
1. Select MultiSig wallet with a Jade signing device controlling one of the key pairs
2. Click on 'Send' and fill in the address field and amount
3. Click 'Create Unsigned Transaction' 
4. Try to sign with a Jade device (you'll need to follow the on-device instructions)

**Screenshots**
Screenshot 2023-11-29 at 3 45 05 PM

 - Browser Chrome
 - Specter Version 2.0.2

moneymanolis:
Fixes #2402

This PR also includes:
- Revert to call enumerate on start-up by default to fix HWI issues on master, changed `enforcehwiinitialisation` to `skiphwiinitialisation`
- Update Jade API to version 2.0.2 and update the Jade hwi client to the latest hwi version
- Avoiding the prompt of the Jade to unlock the device upon Specter start-up
- Providing a UI for registering the multisig (currently only Jade) on the wallets --> setting page if a Jade is part of the multisig, see screenshot:

image

#### Tested
- Jade
- Trezor One
- Trezor Model T


moneymanolis:
@JamieDriver could you check those Jade related commits:
[add "unlock" to jade's enumerate method and its hwi client](https://github.com/cryptoadvance/specter-desktop/pull/2421/commits/375159de948d3f312bce78168f8bd356c2791077)
[add ui to register multisigs for multisig wallets using a jade](https://github.com/cryptoadvance/specter-desktop/pull/2421/commits/619f282113025339adab0f357b6ce5cbd5feb0e5)

requirements.txt
```diff
@@ -26,9 +26,45 @@ bitbox02==6.1.1 \
     # via hwi
 cbor==1.0.0 \
```
JamieDriver, line 27: You might be able to lose `cbor` - it was only present for jade in hwi, and that was upgraded there to use cbor2 (assuming you have latest cut of jade/hwi).  [ unless it is used elsewhere, of course ]

src/cryptoadvance/specter/devices/hwi/jade.py
```diff
@@ -163,12 +164,16 @@ def __init__(
                     "Use JadeAPI.set_[seed|mnemonic] to set simulator wallet"
                 )
         else:
-            if uninitialized and not HAS_NETWORKING:
-                # Wallet not initialised/unlocked nor do we have networking dependencies
-                # User must use 'Recovery Phrase Login' or 'QR Unlock' feature to access wallet
-                raise DeviceNotReadyError(
-                    'Use "Recovery Phrase Login" or "QR PIN Unlock" feature on Jade hw to access wallet'
-                )
+            if uninitialized:
```
JamieDriver, line 168: The below checks seem the wrong way around ... I mean if 'unlock' is false (and you want that to return normally without error), then there's no need to raise an exception if networking is missing.

src/cryptoadvance/specter/devices/hwi/jade.py
```diff
@@ -657,7 +662,10 @@ def can_sign_taproot(self) -> bool:
 
 
 def enumerate(
-    password: Optional[str] = None, expert: bool = False, chain: Chain = Chain.MAIN
+    password: Optional[str] = None,
+    expert: bool = False,
+    chain: Chain = Chain.MAIN,
+    unlock=True,
```
JamieDriver, line 720: Maybe 'unlock=False' is the better option here, and then do something similar/consistent with ledger and trezor if the jade is not unlocked ?  Your call ofc - I am aware that 'unlock=True' will mean existing callers behaviour will be unchanged, so ok maybe that's a more cautious approach.  I'll leave it up to you ofc ;-)

src/cryptoadvance/specter/devices/hwi/jade.py
```diff
@@ -671,9 +679,10 @@ def _get_device_entry(device_model: str, device_path: str) -> Dict[str, Any]:
 
         client = None
         with handle_errors(common_err_msgs["enumerate"], d_data):
-            client = JadeClient(device_path, password, expert, chain, timeout=1)
-            d_data["fingerprint"] = client.get_master_fingerprint().hex()
-
+            client = JadeClient(device_path, password, expert, chain, unlock, timeout=1)
+            # We don't get the fingerprint from a locked Jade
+            if client and unlock:
```
JamieDriver, line 736: Even if 'unlock=False' the jade might be unlocked - ie. if the jade was unlocked already - and you *could* get a fingerprint if you want one.  
I guess the you'd need to 'get_version_info()', get 'JADE_STATE' and test for 'READY' ?

src/cryptoadvance/specter/devices/hwi/jade.py
```diff
@@ -671,9 +679,10 @@ def _get_device_entry(device_model: str, device_path: str) -> Dict[str, Any]:
...
+                d_data["fingerprint"] = client.get_master_fingerprint().hex()
         if client:
             client.close()
```
JamieDriver, line 740: Are you intending to PR this stuff up to HWI ?  [ if not, I might ... or something v similar to make Jade more consistent with ledger/trezor when locked ... ]  But happy to leave to you if you were planning to ... ?

requirements.txt
moneymanolis: Mhh, yes, looks like it is still in there as a transitive dependency from hwi, I am planning / hoping to upgrade to the latest hwi version soonish (to support Trezor 3s) - so will be resolved by this then I guess.

src/cryptoadvance/specter/devices/hwi/jade.py
moneymanolis: I've renamed `unlock` to `unlock_required` in ...

As of now, `unlock_required` is only `False` upon startup. Technically, yes, the Jade could have been unlocked by some other app, but the enumerate call on startup does not actually need the fingerprint. We could also just use `is_startup` as a param for the Jade client / the enumerate function but I thought it'd be better to add something more generic.  Should I change this? 

It also not the end of the world if try to get the fingerprint from the Jade if we can't unlock / auth, looks sth. like this then:

`[{'type': 'jade', 'model': 'jade', 'path': '/dev/cu.usbserial-56D40038341', 'needs_pin_sent': False, 'needs_passphrase_sent': False, 'error': 'Could not open client or get fingerprint information: Unhandled network: ', 'code': -7}]`

src/cryptoadvance/specter/devices/hwi/jade.py
moneymanolis: Happy to leave this up to you :-)

src/cryptoadvance/specter/devices/hwi/jade.py
moneymanolis: I think it is easier to always auth by default and only skip it for certain scenarios, like the startup of Specter ...

src/cryptoadvance/specter/devices/hwi/jade.py
moneymanolis: Makes sense, will change that.

src/cryptoadvance/specter/devices/hwi/jade.py
moneymanolis: Change to `skip_unlocking`

JamieDriver:
All good with me.

stepansnigirev (approved):

Issue #2377, opened 2023-09-07
"You need to select more coins!" when editing a "speed up" transaction


leon-costa:
**Describe the bug**
The "Speed up" button allows to "Fully edit the transaction", which is great when there's no change output (#988), but it fails to create the new transaction. It's probably because the inputs of the original transaction are considered spent, and thus don't appear in the coin selection list.

**To Reproduce**
1. Have an unconfirmed transaction in a wallet
2. Click on 'Speed up'
3. Enter the new fee rate
4. Click on "Advanced" and "Fully edit the transaction"
5. Notice the inputs of the original transaction don't appear in the coin selection table
6. Click on "Create the unsigned transaction"
7. It says "You need to select more coins!"

**Expected behavior**
- Either the new transaction should be created with the new fee rate and the same inputs,
- or the UI should allow the selection of the same inputs as in the replaced transaction

**Desktop:**
 - Where is your node running: local(internal)
 - Where is your specter running: binary
 - If you're running a binary, which OS: Linux
 - Browser: electron
 - Specter Version: tested on 2.0.0, 2.0.1, and 2.0.2pre6

**Additional context**
It seems the code checks if there's enough coins selected by looking at the selected rows whereas there's no row for the spent outputs. The internal `selected-coins` list may contain the right coins though.

PR #2433, opened 2024-04-21
Allow "Speed up" on transactions with a single output

### LINKED ISSUE #988: "Speed up!" RBF feature usable on non-change txs.
While the current "Speed up!" RBF feature is great, it'd be neat to have ability to RBF tx's that do not have any change. 

The user can already do this today with Specter UI, but must manually "hack" around it. Example: 
- Create PSBT for transaction, feerate 10. 
- Save PSBT. 
- Delete the transaction inside Specter. 
- Recreate the exact same transaction (or, at a minimum, with at least 1 of the same inputs), with higher feerate (optionally, do this as many times as one wants) 
- Sign and broadcast the first transaction with the lowest feerate
- Later on, if the tx with lowest feerate hasn't been confirmed and the user wishes to bump the tx, sign & broadcast the identical (or, again, at a minimum reusing 1 of the inputs) tx with a higher feerate

The flow for this could be much easier for a user. Update the "Speed up!" flow to support tx's with non-change by showing the following if no change: 
- If only 1 output: "Text explaining the extra sats for the fees will be taken from the output" 
- If several outputs: "Select which output to take the extra sats for the fees."

Support for adding more/specific inputs to this new, RBF-ing tx seems unnecessary.

leon-costa:
This PR enables the fee bump feature on single-output transactions by reducing the output amount. Without it Specter complains that it can't bump fee in a transaction without change outputs.

This is meant to partially resolve #988. This issue is marked as resolved, but it's not because of #2377 (which seems to be much more complex to solve than what this PR does).

Multiple-output transactions without change are still not supported.

And there's no UI to warn you that the output amount is being reduced. I'm not sure that's necessary.


leon-costa:
Do not hesitate to close this PR if you would rather have a complete fix for the whole issue. It's just that this is the second time I've needed this feature and since I have working code for it, I thought I'd share it in case you or anyone else wants it.

stepansnigirev (approved):

k9ert:
https://github.com/cryptoadvance/specter-desktop/releases/tag/v2.0.3

PR #2435, opened 2024-05-06
Update README.md

KeepYourCoins:
Adding some clarity and consistency in how to provide user authentication details.

Both admin:password and admin:secret are used in curl examples, neither of which actually authenticate. 

Swapping them for user:password shows what the key-pair is meant to represent and then adding a note at the top to say what the default user and password key-value pair is allows for easy authentication.


k9ert (approved):

k9ert:
Thanks, that was definitely a bit confusing.
Deployed: https://docs.specter.solutions/desktop/api/

Issue #397, opened 2020-09-17
warn user if imported key has derivation path different from xpub's depth


stepansnigirev:
inspired by a discussion in a telegram group.
xpub has a depth encoded. if the user enters xpub and derivation path manually he may enter the wrong derivation path. So we should warn the user if he is importing xpub with depth `m` whereas the provided derivation path is longer or shorter than that. We should not warn the user if the derivation path is not provided at all though.

PR #2454, opened 2024-06-02
Warn user if imported key has derivation path different from xpub's depth

roshii:
Fixes https://github.com/cryptoadvance/specter-desktop/issues/397

### DIFF
--- a/src/cryptoadvance/specter/key.py
+++ b/src/cryptoadvance/specter/key.py
@@ -93,8 +93,8 @@ class Key:
                     if path[-1] == "h":
                         path = path[:-1]
                     try:
-                        i = int(path)
-                    except:
+                        int(path)
+                    except ValueError:
                         raise Exception("Incorrect index")
                     derivation_path[0] = "m"
                     derivation = "/".join(derivation_path)
--- a/src/cryptoadvance/specter/key.py
+++ b/src/cryptoadvance/specter/key.py
@@ -1,8 +1,10 @@
+import logging
 from collections import OrderedDict
 from binascii import hexlify
 from embit import base58
 from .util.xpub import get_xpub_fingerprint
 
+logger = logging.getLogger(__name__)
 
 purposes = OrderedDict(
     {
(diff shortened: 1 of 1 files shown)


k9ert:
@stepansnigirev maybe you can have a look whether that PR goes in the right direction.

stepansnigirev:
Makes sense. We should notify about these inconsistencies at least in the logs. Warnings in the UI would be even better.

roshii:
I did try to have this notification sent top UI but my frontend knowledge is pretty limited. Since I will not have enough time to investigate UI modification, I will leave it as is, with a simple warning in the logs.
Hope this helps.

moneymanolis (changes requested):
seems like there is off-by-one error, testing with this: 

`[8c24a510/84h/1h/0h]vpub5Y24kG7ZrCFRkRnHia2sdnt5N7MmsrNry1jMrP8XptMEcZZqkjQA6bc1f52RGiEoJmdy1Vk9Qck9tAL1ohKvuq3oFXe3ADVse6UiTHzuyKx`

image

I've logged `derivation_path` out: 

`['m', '84h', '1h', '0h']`

So this is where the mismatch comes from, it is counting the root whereas the depth of the xpub is counting from the root. @roshii wanna fix it?

Propagating the error to the UI is simple, just raise an Exception instead of the log statement (or in addition):

 ```
        raise Exception(
                f"xpup has a depth of {depth} while derivation path "
                f"indicates the key is {len(derivation_path)} levels deep"
            )
```

This code is taking the exceptions and their strings and propagates them to `src/cryptoadvance/specter/server_endpoints/devices.py`:


```
    @classmethod
    def parse_xpubs(cls, xpubs):
        xpubs = xpubs
        lines = [l.strip() for l in xpubs.split("\n") if len(l) > 0]
        failed = []
        keys = []
        for line in lines:
            try:
                keys.append(Key.parse_xpub(line))
            except Exception as e:
                failed.append(line + "\n" + str(e))
        return keys, failed
        
```

PR #2458, opened 2024-06-17
Removing bleak entirely as it's only available for MacOS

k9ert:
missing bleak caused:
https://gitlab.com/cryptoadvance/specter-desktop/-/jobs/7112728484

I'm not sure why this did not came up earlier.


moneymanolis (approved):

moneymanolis:
haven't tested, though, whether the BLE class is somehow need by the rest of the Jade code. I don't think so.

PR #2466, opened 2024-07-31
Added background bluring when modal is shown

xavierfiechter:
This change reduces distraction from the UI when the modal is shown and increases focus on what's important.


moneymanolis:
superseded by https://github.com/cryptoadvance/specter-desktop/pull/2481 - it was just easier to quickly to a separate PR, since this can be done with one line using tailwind, have a look at the PR, overall please always strive to only use tailwind classes and no custom classes.

PR #2467, opened 2024-08-01
Fix wallet load error on incomplete gettransaction (Failed to load utxos, IndexError: list index out of range)

leon-costa:
The error happens when the wallet has some locked outputs in `listlockunspent` and the corresponding `gettransaction` doesn't contain the details of this particular output.

It may be related to https://github.com/bitcoin/bitcoin/issues/28555. It is at least the same symptom.

The locked outputs were locked by creating a transaction in Specter. This has happened to me multiple times.

The UI says it failed to load the wallet with the following error: Failed to load utxos, IndexError: list index out of range

The raw transaction does contain the missing output so we use that instead.

<details><summary>
Here's an example that generates the error:
</summary>

```
listlockunspent:
[
  {
    "txid": "<some txid>",
    "vout": 0
  }
]

gettransaction <some txid>:
{
  "amount": <redacted>,
  "fee": <redacted>,
  "confirmations": <redacted>,
  "blockhash": "<redacted>",
  "blockheight": <redacted>,
  "blockindex": <redacted>,
  "blocktime": <redacted>,
  "txid": "<some tx id>",
  "wtxid": "<redacted>",
  "walletconflicts": [
  ],
  "time": <redacted>,
  "timereceived": <redacted>,
  "bip125-replaceable": "no",
  "details": [
    {
      "address": "<redacted>",
      "category": "send",
      "amount": <redacted>,
      "label": "<redacted>",
      "vout": 1,
      "fee": <redacted>,
      "abandoned": false
    }
  ],
  "hex": "<some raw tx hex>",
  "lastprocessedblock": {
    "hash": "<redacted>",
    "height": <redacted>
  }
}

decoderawtransaction <some raw tx hex>:
{
  "txid": "<some txid>",
  "hash": "<redacted>",
  "version": 2,
  "size": <redacted>,
  "vsize": <redacted>,
  "weight": <redacted>,
  "locktime": <redacted>,
  "vin": [
    {
      "txid": "<some other txid>",
      "vout": 0,
      "scriptSig": {
        "asm": "",
        "hex": ""
      },
      "txinwitness": [
        "<redacted>",
        "<redacted>"
      ],
      "sequence": <redacted>
    }
  ],
  "vout": [
    {
      "value": <redacted>,
      "n": 0,
      "scriptPubKey": {
        "asm": "<redacted>",
        "desc": "<redacted>",
        "hex": "<redacted>",
        "address": "<redacted>",
        "type": "<redacted>"
      }
    },
    {
      "value": <redacted>,
      "n": 1,
      "scriptPubKey": {
        "asm": "<redacted>",
        "desc": "<redacted>",
        "hex": "<redacted>",
        "address": "<redacted>",
        "type": "<redacted>"
      }
    }
  ]
}
```

</details>

Note: I'm connecting to a Bitcoin Core version v27.1.0.

### DIFF
--- a/src/cryptoadvance/specter/wallet/wallet.py
+++ b/src/cryptoadvance/specter/wallet/wallet.py
@@ -732,17 +732,10 @@ class Wallet(AbstractWallet):
                         # It may be related to https://github.com/bitcoin/bitcoin/issues/28555
                         # In this case we decode the raw transaction to get the data we want.
                         raw_transaction_hex = tx_from_core["hex"]
-                        raw_transaction = self.rpc.decoderawtransaction(
-                            raw_transaction_hex
-                        )
-                        searched_raw_vout = [
-                            _output
-                            for _output in raw_transaction["vout"]
-                            if _output["n"] == utxo_vout
-                        ][0]
-                        tx_copy["amount"] = searched_raw_vout["value"]
-                        script_pubkey = searched_raw_vout["scriptPubKey"]
-                        tx_copy["address"] = script_pubkey["address"]
+                        parsed_transaction = self.TxCls.from_string(raw_transaction_hex)
+                        out = parsed_transaction.vout[utxo_vout]
+                        tx_copy["amount"] = round(out.value * 1e-8, 8)
+                        tx_copy["address"] = out.script_pubkey.address(self.network)
 
                 # Append the copy to the _full_utxo list
                 _full_utxo.append(tx_copy)
--- a/src/cryptoadvance/specter/wallet/wallet.py
+++ b/src/cryptoadvance/specter/wallet/wallet.py
@@ -732,14 +732,17 @@ class Wallet(AbstractWallet):
                         # It may be related to https://github.com/bitcoin/bitcoin/issues/28555
                         # In this case we decode the raw transaction to get the data we want.
                         raw_transaction_hex = tx_from_core["hex"]
-                        raw_transaction = self.rpc.decoderawtransaction(raw_transaction_hex)
+                        raw_transaction = self.rpc.decoderawtransaction(
+                            raw_transaction_hex
+                        )
                         searched_raw_vout = [
                             _output
                             for _output in raw_transaction["vout"]
                             if _output["n"] == utxo_vout
                         ][0]
                         tx_copy["amount"] = searched_raw_vout["value"]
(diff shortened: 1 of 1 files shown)


moneymanolis:
just tested this real quick because I had the same error while testing (again) and one "issue" to mention is that `decoderawtransaction` is not implemented in Spectrum. I haven't looked at the rest of the code yet to be fair, but if it is needed, ideally, this would be added as well to Spectrum. From a quick look, it might be similar to how [getrawtransaction](https://github.com/stepansnigirev/spectrum/blob/5b1fca9a1debf8cbbaf2f1a9b2f4bb02fddd4f0f/src/cryptoadvance/spectrum/spectrum.py#L702) was implemented,, i.e. using an ElectrumX API call.

leon-costa:
ElectrumX doesn't seem to have such an API. [blockchain.transaction.get](https://electrumx.readthedocs.io/en/latest/protocol-methods.html#blockchain-transaction-get) with `verbose=true` seems pretty close, but it takes a transaction hash as an input, not the transaction hex. It could work here because that's what we start with anyway, but I don't know how that could be done with all the intermediaries in your setup.

I see in the spectrum PR you mentioned that embit is used, which [seems to be able to decode raw transactions](https://embit.rocks/#/api/transaction). So it should not be difficult to implement a minimal version of `decoderawtransaction` that only returns the fields used by this PR, i.e. the `vout` array, and inside it only `n`, `value` and `scriptPubKey` / `address`. The first example in the embit page I linked seems to provide an example of all the necessary code.

Actually, specter-desktop also uses embit, so maybe this should be done directly here instead of calling `decoderawtransaction`.

moneymanolis:
> Actually, specter-desktop also uses embit, so maybe this should be done directly here instead of calling decoderawtransaction.

Might be the easiest solution? :-)

leon-costa:
I made the change to use embit in this fix.

oren-z0:
@leon-costa Could it be related to https://github.com/cryptoadvance/specter-desktop/issues/2475 ?
Do you get the "list index out of range" after sending funds from the wallet to itself?

oren-z0:
@leon-costa Please see https://github.com/cryptoadvance/specter-desktop/pull/2476 . I wonder if it fixes your problem too.

moneymanolis:
Closing, superseded by https://github.com/cryptoadvance/specter-desktop/pull/2485 - THANK YOU STILL! Especially for the embit part!

PR #2487, opened 2025-01-03
fix: Modernize CI

AaronDewes:
This

- Updates the GitHub actions workflows with modern versions
- Updates the base image of the Docker container to Debian 12
- Sets up caching for faster Docker builds
- Fix the invalid dependency issue

### DIFF
--- a/.github/workflows/docker-tag.yml
+++ b/.github/workflows/docker-tag.yml
@@ -39,4 +39,6 @@ jobs:
           docker buildx build \
           --platform linux/amd64,linux/arm64 \
           --tag ghcr.io/$IMAGE_NAME:$TAG \
+          --cache-from type=gha \
+          --cache-to type=gha,mode=max \
           --output "type=registry" ./
--- a/.github/workflows/docker-tag.yml
+++ b/.github/workflows/docker-tag.yml
@@ -9,11 +9,11 @@ on:
 jobs:
   build:
     name: Build image
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
 
     steps:
       - name: Checkout project
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
 
       - name: Set env variables
         run: |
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -9,25 +9,29 @@
...
+COPY requirements.txt .
 
 RUN pip3 install --upgrade pip
 RUN pip3 install babel cryptography
-RUN pip3 install .
+RUN pip3 install -r requirements.txt
+
+COPY . .
+
+RUN pip3 install . --no-deps
```
AaronDewes, line 31: This is the main part of this fix, which is a change in the order deps are installed.

It also makes builds faster as a side effect.

Dockerfile
```diff
@@ -9,25 +9,29 @@
...
 ARG DIR=/data/
 
-FROM python:3.10-slim-bullseye AS builder
+FROM python:3.10-bookworm AS builder
 
 ARG VERSION
 ARG REPO
 
-RUN apt update && apt install -y git build-essential libusb-1.0-0-dev libudev-dev libffi-dev libssl-dev rustc cargo libpq-dev
+RUN apt update && apt install -y git libusb-1.0-0-dev libudev-dev libffi-dev libssl-dev rustc cargo libpq-dev
```
AaronDewes, line 17: This is also just a build speed optimization, instead of using the slim image, just use the non-slim image during build, which has most packages we need for building preinstalled.

.github/workflows/docker-tag.yml
```diff
@@ -22,21 +22,23 @@ jobs:
...
         run: |
-                docker buildx build \
-                --platform linux/amd64,linux/arm64 \
-                --tag ghcr.io/$IMAGE_NAME:$TAG \
-                --output "type=registry" ./
+          docker buildx build \
+          --platform linux/amd64,linux/arm64 \
+          --tag ghcr.io/$IMAGE_NAME:$TAG \
+          --cache-from type=gha \
+          --cache-to type=gha,mode=max \
```
AaronDewes, line 43: This enables caching.

AaronDewes (commented):
Short explanations of the main changes

moneymanolis (approved):
Thanks so much for this! Tested locally as well.

PR #2550, opened 2026-02-13
Replace deprecated distutils usage (removed in Python 3.12)

### LINKED ISSUE #2549: Code uses distutils which is removed in Python 3.12+
Two files import from `distutils`, which was deprecated in Python 3.10 and fully removed in Python 3.12:

1. **`src/cryptoadvance/specter/server.py`** (line 4):
   `from distutils.core import setup`
   This import doesn't even seem to be used anywhere in the file. It can just be deleted.

2. **`src/cryptoadvance/specter/util/common.py`** (line 7):
   `from distutils.util import strtobool`
   This is actually used, but it's easy to replace with a small helper function that does the same thing.

If someone tries to run Specter on Python 3.12 or newer, it will crash on startup with a `ModuleNotFoundError: No module named 'distutils'`. Since the project is actively maintained and people will eventually move to newer Python versions, this should probably be fixed.

abhishek-8081:
Fixes #2549

`distutils` was deprecated in Python 3.10 and fully removed in Python 3.12. This PR removes both usages from the codebase:

### Changes

1. **`server.py`**: Removed `from distutils.core import setup` — this import was not used anywhere in the file.

2. **`util/common.py`**: Replaced `from distutils.util import strtobool` with an inline implementation inside the existing `str2bool()` function. The behavior is exactly the same — it accepts the same true values (`y, yes, t, true, on, 1`) and false values (`n, no, f, false, off, 0`) and raises `ValueError` for anything else.

Without this fix, Specter will crash on startup with `ModuleNotFoundError: No module named 'distutils'` if run on Python 3.12+.

### DIFF
--- a/src/cryptoadvance/specter/util/common.py
+++ b/src/cryptoadvance/specter/util/common.py
@@ -4,6 +4,7 @@ from datetime import datetime
 import json
 from flask_babel.speaklater import LazyString
 from typing import Union
+
 logger = logging.getLogger(__name__)
 
 


al-munazzim (approved):
Clean Python 3.12+ compatibility fix. Replaces `distutils.strtobool` (removed in 3.12) with equivalent inline logic in `str2bool()`. The accepted values match the original `strtobool` behavior exactly.

LGTM ✅

k9ert:
@abhishek-8081 Black is failing! Can you run black?!

PR #2620, opened 2026-05-09
Update specterext-timelockrecovery to v0.2.3

oren-z0:
Fixed floating point errors when converting BTC amount from BTC to sats (multiplying by 1e8).

Added descriptors to input of the recovery-transaction and the cancellation-transaction - necessary for some hardware wallets (i.e. Specter DIY) to identify the inputs.

Added descriptors to the self-sending outputs of the recovery-transaction and cancellation-transaction - will be displayed better on some hardware wallets (i.e. don't deduct from total-amount, show address-index and don't make it look like an external destination).

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -35,7 +35,7 @@ specterext-exfund==0.1.7
 specterext-faucet==0.1.2
 cryptoadvance.spectrum==0.7.0
 specterext-stacktrack==0.3.0
-specterext-timelockrecovery==0.2.1
+specterext-timelockrecovery==0.2.3
 
 # workarounds
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -824,9 +824,9 @@ specterext-stacktrack==0.3.0 \
     --hash=sha256:14f96f1f552f57ba017b8bc642f07343edbb1abafe09e03bbaae179d78d7ce23 \
     --hash=sha256:9e2946185730aab377951e83a27d8791a34e0f031e44f15991212b6b85722ca0
     # via -r requirements.in
-specterext-timelockrecovery==0.2.1 \
-    --hash=sha256:2fc2a3eec045377b54926ff41dab9c5f5a6f46074150aec1ff67cb1456fef911 \
-    --hash=sha256:35e181971b5b69ca6706496169c4581f2fa20706c88ba24760e56e153ec51939
+specterext-timelockrecovery==0.2.3 \
+    --hash=sha256:a0b22c4e010061055da4ea6ea4e08bbc38d1f7c2ec58327970caa0ce987cba95 \
+    --hash=sha256:5df3deb6245a22d48f75bbb62fd87dc141a9f60407011ce9fbb0a7bcc40fe4cc
     # via -r requirements.in
 sqlalchemy==1.4.52 \
     --hash=sha256:1296f2cdd6db09b98ceb3c93025f0da4835303b8ac46c15c2136e27ee4d18d94 \


al-munazzim:
Review (security + bug-risk):

I checked this bump from specterext-timelockrecovery 0.2.1 -> 0.2.3 and the upstream delta (v0.2.1...v0.2.3).

What changed upstream:
- float->sat conversion switched to Math.round(...) to avoid precision drift
- PSBT input/output descriptor metadata is now filled explicitly (improves HW wallet signing/UX)
- bech32 addresses normalized to lowercase

Security assessment:
- No obvious new attack surface in this PR itself (only dependency pin update in requirements.in/requirements.txt).
- Hash-pinned wheel/sdist entries are updated, which is good supply-chain hygiene.
- The upstream changes are mostly transaction-construction correctness and metadata handling, not privilege/auth/network surface changes.

Potential bug/regression risks to watch:
1) Rounding behavior changed (floor -> round) in multiple BTC->sat conversions; verify no +1 sat edge effects in split/leftover math.
2) Descriptor-fill now raises on failure (SpecterError), so recovery/cancellation flows may fail hard where they previously progressed with less metadata.
3) Auto-lowercasing bech32 is correct for bech32, but ensure no unintended mutation for non-bech32 recipient formats.

Suggested checks before merge:
- Add/confirm test cases around BTC decimal edge values (e.g. 0.00098487, very small dust-adjacent values).
- Exercise TimelockRecovery end-to-end with Specter DIY and at least one non-DIY signer.

Verdict: looks directionally good and addresses a real correctness bug; I’d merge after the above regression checks are green.

al-munazzim:
Thanks for the CI signal — I dug into the failing runs in detail.\n\nThe 3 red checks (, , ) are **not caused by the extension version bump itself**. All three fail early in the shared bootstrap step  () with:\n\n- \n- \n\nSo the jobs fail before Python deps/tests for this PR’s code changes are exercised.\n\nI’ll prepare a separate PR to harden/fix this CI verification path. Once that is merged, we should be able to rerun this PR and get the checks green here as well.

al-munazzim:
Correction (my previous comment had formatting issues):

Thanks for the CI signal — I dug into the failing runs in detail.

The 3 red checks (`test`, `extension-smoketest`, `cypress`) are **not caused by the extension version bump itself**. All three fail early in the shared bootstrap step `Install elementsd` (`tests/install_noded.sh --elements binary`) with:

- `ERROR: upstream GPG verification failed for elements SHA256SUMS.asc`
- `NO_PUBKEY 2F2A88D7F8D68E87`

So the jobs fail before Python deps/tests for this PR’s code changes are exercised.

I’ll prepare a separate PR to harden/fix this CI verification path. Once that is merged, we should be able to rerun this PR and get the checks green here as well.

al-munazzim:
Follow-up is up: #2621\n\nIt hardens This script will result in having bitcoind or elementsd binaries, either by binary download or via compilation
Do one of these:
$ ./install_node.sh --bitcoin binary
$ ./install_node.sh --bitcoin compile
$ ./install_node.sh --elements binary
$ ./install_node.sh --elements compile
For more context, see https://github.com/cryptoadvance/specter-desktop/blob/master/docs/development.md#how-to-run-the-tests key import for elements signature verification (imports from both key sources before VALIDSIG check), which should address the  failure path we saw.\n\nAfter #2621 merges, we can rerun this PR and checks should be able to go green.

al-munazzim:
Correction (previous comment had shell-formatting issues):

Follow-up is up: #2621

It hardens `tests/install_noded.sh` key import for elements signature verification (imports from both key sources before `VALIDSIG` check), which should address the `Install elementsd` failure path we saw.

After #2621 merges, we can rerun this PR and checks should be able to go green.

oren-z0:
Rebased over @al-munazzim last commit

k9ert (approved):

PR #2633, opened 2026-06-03
fix: restore Ledger udev hidraw fallback

### LINKED ISSUE #2622: Specter 2.1.8 new udev rules for Ledger do not work.
**Describe the bug**
With the Specter v2.1.8 download came new udev rules for Ledger.
When using these rules my Ledger Nano S was not recognized (My Ledger firmware was updated last month, so pretty up to date).
I was able to select my Ledger device, but was not able to sign a transaction with it.

**To Reproduce**
Use the new udev rules for Ledger that come with Specter v2.1.8
Try to sign a transaction with the Ledger.
The signing fails (could not retreive key).

**Desktop**
 - Running Bitcoin Core locally on Linux Mint.
 - Specter runs locally as an AppImage on Linux Mint
 - Specter Version 2.1.8

**Additional context**
When I revert back to the old udev rules for Ledger, everything works fine under Specter v2.1.8
I downloaded these from https://github.com/LedgerHQ/udev-rules/blob/master/20-hw1.rules

The Ledger udev rules from Specter v2.1.7 work fine. The Coldcard and Trezor udev rules from both v2.1.7 and v2.1.8 work fine as well.

[New-20-hw1.rules-withSpecterV2.1.8.txt](https://github.com/user-attachments/files/27604315/New-20-hw1.rules-withSpecterV2.1.8.txt)
[Old-20-hw1.rules-fromLedgerGithub.txt](https://github.com/user-attachments/files/27604314/Old-20-hw1.rules-fromLedgerGithub.txt)

al-munazzim:
## Summary
- Restore Ledger's current upstream `20-hw1.rules` layout for vendor `2c97`
- Remove the product-id-restricted Ledger rules so newer Ledger firmware/products keep matching
- Restore the `hidraw*` fallback with `MODE="0666"` for hidraw-based Ledger communication

## Test Plan
- Verified `udev/20-hw1.rules` contains the vendor-only `2c97` rule and the `hidraw*` fallback
- Compared the updated rules against `LedgerHQ/udev-rules` `master` `20-hw1.rules` (only trailing newline differs)

Fixes #2622


k9ert:
@al-munazzim Did you had a look at the rules attached to this comment?
https://github.com/cryptoadvance/specter-desktop/issues/2622#issuecomment-4440623981

Can you please go in more detail why thos changes make sense?

al-munazzim:
I looked at the attached rules from https://github.com/cryptoadvance/specter-desktop/issues/2622#issuecomment-4440623981 in detail.

The relevant Ledger part of that comment is basically arguing for two compatibility properties:

1. Match Ledger by vendor id `2c97` instead of enumerating product ids.
   - That is useful because Ledger keeps adding product/app/interface product ids.
   - Product-id matching can become stale even when the vendor id remains stable.
   - The reporter's failure mode fits this: a device/path not covered by the narrow product-id list will not receive desktop access permissions.

2. Keep a `hidraw*` fallback in addition to the `SUBSYSTEMS=="usb"`/`uaccess` rule.
   - Some stacks access Ledger over hidraw rather than libusb.
   - `TAG+="uaccess"` on the USB device alone is not always enough for those hidraw consumers.
   - The fallback `KERNEL=="hidraw*", ATTRS{idVendor}=="2c97", MODE="0666"` restores the older compatibility path.

What I intentionally did *not* take from the attachment:

- I did not copy the broader `GROUP="plugdev"`/`MODE="0666"` permissions onto the USB subsystem rule. That would be more permissive than the current Ledger upstream-style rule and may be unnecessary on modern udev/logind setups.
- I did not change Trezor or Coldcard rules, because the issue and reproduction point at Ledger and the existing Specter file only bundles this Ledger/HW.1 rule file.

So this PR is a narrower version of the attached-rule idea: it keeps the compatibility-relevant Ledger changes — vendor-only `2c97` and hidraw fallback — while avoiding a broader permission change for all supported hardware.

I also compared this shape against Ledger's upstream rule structure before opening the PR. The current PR restores the upstream-style Ledger rule shape rather than inventing a Specter-specific rule set.

Cyberwiz9000:
That looks good to me.

k9ert (approved):

PR #2634, opened 2026-06-04
Accept descriptor QR exports from Specter DIY

### LINKED ISSUE #2625: Can't scan Wallet Descriptor from Specter DIY
**Describe the bug**
Specter DIY exports the Wallet Descriptor in as a QR code containing a json that Specter Desktop can't scan.
I'm not sure which side does not follow the official standard, but even if it's Specter DIY, it would be nice if Specter Desktop could read it anyways.

**To Reproduce**
After loading a seed to Specter DIY go to:

Wallets -> Default -> Settings -> Export wallet descriptor -> Show QR code.

The QR will contain something like:
```json
{"descriptor": "wpkh([9616b1f5/84h/0h/0h]xpub6DV...d4", "label": "Default"}
```

On Specter Desktop go to:
Add device -> Specter-DIY -> Scan QR code
"Unknown key format"

al-munazzim:
Fixes #2625.

## Summary
- Parse JSON QR payloads that contain a Bitcoin output descriptor, like Specter DIY's wallet descriptor export.
- Extract origin-prefixed extended public keys from descriptor bodies and feed them into the existing xpub import paths.
- Keep existing Cobo/Coldcard JSON handling unchanged and support both current and legacy device import templates.

## Verification
- Ran a Node smoke test for descriptor extraction with singlesig, nested singlesig, and sortedmulti descriptor examples.

### DIFF
--- a/src/cryptoadvance/specter/templates/device/components/add_keys.js
+++ b/src/cryptoadvance/specter/templates/device/components/add_keys.js
@@ -23,3 +23,15 @@ function setDeviceType(type) {
 		showNotification(`Device type changed to ${name}`);
 	}
 }
+
+function getXpubsFromDescriptor(descriptor) {
+	let data = "";
+	descriptor = descriptor.split("#")[0];
+	let matches = descriptor.matchAll(/\[([^\]]+)\]([A-Za-z0-9]+)/g);
+	for (let match of matches) {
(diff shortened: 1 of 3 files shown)


k9ert:
underlying issue closed. Let's not fix what's not broken. Closing.

Issue #2335, opened 2023-05-17
Error: Failed to load utxos, IndexError: list index out of range ✕


runjam1:
When I click on a wallet to open it I get this error message:

Failed to load utxos, IndexError: list index out of range ✕

CryptoCase:
I am having this same issue. This did not happen on the previous Spectre versions, after updating I am now unable to access my wallets due to this issue even after rescanning and resyncing with the blockchain.

xblok:
Hi, I posted about this issue on Telegram and just adding some information here...

The error that I'm getting is `[2023-07-03 21:21:26,679] ERROR in wallet: list index out of range
Traceback (most recent call last):
  File "..../specter-desktop/src/cryptoadvance/specter/wallet/wallet.py", line 727, in check_utxo
    searched_vout = [
                    ^
IndexError: list index out of range
[2023-07-03 21:21:26,680] ERROR in wallet_manager: Failed to load utxos, IndexError: list index out of range
Traceback (most recent call last):
  File "..../specter-desktop/src/cryptoadvance/specter/wallet/wallet.py", line 727, in check_utxo
    searched_vout = [
                    ^
IndexError: list index out of range`

Looking at the source code, the error comes from `wallet.py`, lines 718-725.

Running `bitcoin-cli listlockunspent` on my node returns the following:

`[
  {
    "txid": "ID1",
    "vout": 0
  }
]`

Details of that transaction returned by the `bitcoin-cli gettransaction ID1` command have a details object where `vout == 1`.

`tx_from_core` holds the ID1 transaction with vout == 1 but the `utxo_vout `that it filters by is == 0, so the `searched_vout` won’t be found.

Some more context on my error: it's a 2-3 multisig wallet. I dug out local Specter logs and found some more details that could be relevant. Shortly before the multisig walled stopped loading with the above error, I see a warning in the log: `WARNING in wallet: Request error for method lockunspent(True, [{'txid': 'ID2', 'vout': 1}]): Invalid parameter, expected unspent output`.

ID2 refers to a transaction that I sent before ID1 was created from my single sig wallet to the multisig wallet and which confirmed successfully. I also remember selecting the UTXO that I got from that transaction, and there is a chance that it was selected when I created a PSBT that would result in ID1 transaction. Could that have led to some kind of mix-up?

[...]

PR #2666, opened 2026-07-28
fix: retry wallet import with widened range on shrink-refusal (#2604)

### LINKED ISSUE #2604: wallet_importer: "new range must include current range" on wallet re-create after partial import
Split from #2335 (see @alvistar and @ironrai comments there).

## Symptom

Re-creating a wallet in Specter fails with:

```
ERROR in wallet_importer: new range must include current range = [0,1008] and new range must include current range = [0,1000]
  File ".../cryptoadvance/specter/util/wallet_importer.py", line 235, in create_wallet
    self.wallet = wallet_manager.create_wallet(...)
  File ".../cryptoadvance/specter/managers/wallet_manager.py", line 353, in create_wallet
    w = self.WalletClass.create(...)
  File ".../cryptoadvance/specter/wallet/wallet.py", line 362, in create
    raise SpecterError(all_issues)
SpecterError: new range must include current range = [0,1001] and new range must include current range = [0,1000]
```

Reported on v2.0.2 (macOS and Linux). Not verified on current v2.1.7 — please confirm if still reproducible.

## Repro (from @ironrai in #2335)

1. Open a wallet, start a transaction, close the dialog without signing.
2. Hit the `check_utxo` IndexError from #2335 and can no longer open the wallet.
3. Try to re-import the wallet under the same name → fails with the error above. Workaround was deleting the wallet file in Bitcoin Core and re-importing under a different name.

## Root cause (best guess)

Wallet creation path at \`wallet.py:353-362\`:

\`\`\`python
if use_descriptors:
    res = wallet_rpc.importdescriptors(args)
else:
    res = wallet_rpc.importmulti(args, {\"rescan\": False})

if not all([r[\"success\"] for r in res]):
    all_issues = \" and \".join(...)
    raise SpecterError(all_issues)
\`\`\`

On the legacy (non-descriptor) path, args carry \`\"range\": [0, cls.GAP_LIMIT]\`. If the Bitcoin Core wallet file already contains the same descriptor at a wider range (e.g. \`[0, 1008]\` after \`keypoolrefill\` extended it), Core rejects \`importmulti\` with a new range of \`[0, 1000]\` because the new range does not *include* the current one. Core's rule: range can only grow on re-import, never shrink.

The error only surfaced because the earlier #2335 crash left the Core-side wallet in a state that Specter then couldn't cleanly recover from — but the shrink-refusal is an independent Specter bug.

## Suggested fix directions

- Query current descriptor range before re-import and pass \`max(current_range_end, GAP_LIMIT)\` rather than a fixed \`GAP_LIMIT\`.
- Or: prefer the descriptor path (\`importdescriptors\` with \`active=True\`) for all Core versions that support it — the descriptor path doesn't carry the \`range\` constraint.
- Or: on shrink-refusal, retry with the wider range extracted from the error, or explicitly re-unload/re-load the wallet file.

ack-chai:
Fixes #2604.

Bitcoin Core's legacy `importmulti` refuses to shrink an already-imported descriptor's keypool range (`new range must include current range`). This can surface when re-creating a wallet under a name Core already has a wider range recorded for (e.g. after a `keypoolrefill`).

This PR parses the range Core reports in its error and retries the import once with a range wide enough to include it, rather than failing outright. The descriptor-wallet path (`importdescriptors`) is unaffected since it doesn't carry this range constraint.

Note: per the issue thread, this wasn't confirmed reproducible on v2.1.7+ (only v2.0.2) — this fixes the underlying shrink-refusal bug regardless, since it's a real gap in the legacy import path independent of the #2335 crash that originally surfaced it.

Added unit tests in `tests/test_wallet_create_range.py` (fully mocked RPC, no regtest node needed) covering: successful retry on shrink-refusal, correct failure when retry also fails, and no pointless retry on unrelated errors.


al-munazzim (approved):
Looks good to me.

I reviewed the legacy-wallet `importmulti` path against #2604. The retry is narrow: it only triggers when Core returns the specific shrink-refusal message (`new range must include current range = [...]`), computes a range covering the widest reported current range across receive/change imports, retries once, and still falls through to the existing `SpecterError` handling if the retry fails or the error is unrelated.

Verification I ran locally on PR head `14918313`:

```text
python -m pytest tests/test_wallet_create_range.py -q
# 3 passed, 1 warning in 0.04s

python -m py_compile src/cryptoadvance/specter/wallet/wallet.py tests/test_wallet_create_range.py
# OK

python -m black --check src/cryptoadvance/specter/wallet/wallet.py tests/test_wallet_create_range.py
# 2 files would be left unchanged
```

CI also completed green at review time: black, Build & smoke-test specterd (Linux), test, cypress, and extension-smoketest all passed.

k9ert (approved):

k9ert:
Thanks for your PR!

PR #2676, opened 2026-08-06
Warn on ColdCard Mk3 seed-entropy advisory at device setup

nrobi144:
## Summary

Shows a security warning on the ColdCard device-setup page, linking Coinkite's advisory.

In July 2026 Coinkite [disclosed](https://blog.coinkite.com/coldcard-mk3-seed-generation-warning/) a firmware build error that reduced generated-seed entropy on **ColdCard Mk3** (firmware 4.0.1-5.0.3) and pre-fix **Mk4/Q**. Specter doesn't generate the ColdCard seed, so this isn't a Specter flaw — but users who add a ColdCard via Specter may hold funds on an affected seed.

## Change

Single-template change in `new_device_keys.jinja`: a warning banner inserted right before the existing `#coldcard-instructions` block, reusing the established `device_class.device_type != 'coldcard'` guard, the existing `.warning` CSS class, and the `{{ _() }}` i18n convention. It renders only for ColdCard (device_type "coldcard" has no subclasses, so Passport/Keystone/Cobo are unaffected). No Python/route/model changes.

Refs #2675.

### DIFF
--- a/src/cryptoadvance/specter/static/output.css
+++ b/src/cryptoadvance/specter/static/output.css
@@ -1100,6 +1100,10 @@ input[type="number"]::-webkit-outer-spin-button,
   z-index: 50;
 }
 
+.z-20 {
+  z-index: 20;
+}
+
 .z-30 {
   z-index: 30;
 }
--- a/src/cryptoadvance/specter/templates/device/new_device/new_device_keys.jinja
+++ b/src/cryptoadvance/specter/templates/device/new_device/new_device_keys.jinja
@@ -30,10 +30,14 @@
                 </div>
             {% endif %}
             
-            <div id="coldcard-warning" class="warning {% if device_class.device_type != 'coldcard' %}hidden{% endif %}">
-                {{ _("Security notice:") }}
-                {{ _("Coinkite disclosed a firmware flaw that reduced generated-seed entropy on ColdCard Mk2/Mk3 (firmware 4.0.1-4.1.9) and Mk4/Mk5/Q before their fixed firmware. If your seed was generated on-device with fewer than 50 independent dice rolls, update firmware, generate a new seed and move your funds.") }}
-                <a href="https://blog.coinkite.com/coldcard-mk3-seed-generation-warning/" target="_blank" rel="noopener noreferrer">{{ _("Read Coinkite's advisory") }}</a>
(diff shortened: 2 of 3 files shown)


al-munazzim (changes requested):
Thanks for jumping on this quickly. The placement and low-scope template-only approach look fine, and I confirmed the updated template still parses.

I think we should adjust the warning text before merging, because the firmware/model details here no longer match Coinkite's current advisory:

- The text says `ColdCard Mk3 (firmware 4.0.1-5.0.3)`, but the current Coinkite advisory says Mk2/Mk3 are affected on `4.0.1 through 4.1.9 inclusive`, with fixed Mk2/Mk3 firmware `4.2.0 or later`.
- The text says `pre-fix Mk4/Q`, but the advisory also lists Mk5, and gives model-specific fixed releases: Mk4/Mk5 standard `5.6.0+`, Q standard `1.5.0Q+`, Mk4/Mk5 Edge `6.6.0X+`, Q Edge `6.6.0QX+`.

Can you update the banner to avoid the stale/incorrect `4.0.1-5.0.3` range and include Mk5? A concise version could be something like:

> Coinkite disclosed a firmware flaw affecting seeds generated on Mk2/Mk3 firmware 4.0.1-4.1.9, and on Mk4/Mk5/Q before the fixed standard/Edge firmware releases. If your seed was generated on-device without at least 50 independent dice rolls, update firmware, generate a new seed, and move your funds.

The advisory link can carry the exact per-model fixed versions. Once the warning text is corrected, this looks good to me.

nrobi144:
You're right, thanks for the careful check — I verified against Coinkite's advisory and the firmware/model details were off. Corrected the banner:

> Coinkite disclosed a firmware flaw affecting seeds generated on ColdCard Mk2/Mk3 (firmware 4.0.1-4.1.9) and on Mk4/Mk5/Q before their fixed firmware. If your seed was generated on-device with fewer than 50 independent dice rolls, update firmware, generate a new seed and move your funds.

Dropped the stale `4.0.1-5.0.3` range, added Mk5, and left the exact per-model fixed versions to the advisory link. Pushed in the latest commit.

al-munazzim (approved):
Thanks for updating this. The stale firmware range is removed, Mk5 is included, and the exact fixed-version details are left to Coinkite's advisory link, so this addresses my review concern.

k9ert (approved):

k9ert:
Thanks for the PR. Very much appreciated!

PR #2687, opened 2026-08-08
fix(swan): add timeout to remote HTTP requests to prevent hang

bunlongheng:
The Swan integration makes several outbound HTTP requests - token POST, authenticated GETs, POST/PATCH/PUT/DELETE via the generic request helper, and GitHub template fetches in the extension generator - with no timeout set. A slow or unresponsive remote endpoint will hang the wallet process indefinitely.

Added a 30-second timeout to each requests call in the Swan OAuth client and extension generator so the process can fail fast and recover instead of blocking forever.

### DIFF
--- a/src/cryptoadvance/specter/services/extension_gen.py
+++ b/src/cryptoadvance/specter/services/extension_gen.py
@@ -152,7 +152,7 @@ class ExtGen:
             shutil.copy(sourcepath, targetpath)
             print(f"    --> Created {targetpath} (copied)")
         else:
-            r = requests.get(self.env.loader.url_for_template(sourcepath))
+            r = requests.get(self.env.loader.url_for_template(sourcepath), timeout=30)
             open(targetpath, "wb").write(r.content)
             print(f"    --> Created {targetpath} (via Github)")
 
@@ -233,7 +233,7 @@ class GithubUrlLoader(BaseLoader):
     def get_source(self, environment, template):
         url = self.url_for_template(template)
         for attempt in range(3):
-            r = requests.get(url)
+            r = requests.get(url, timeout=30)
             if r.status_code == 200:
                 return r.text, url, None
             if r.status_code == 429 and attempt < 2:
--- a/src/cryptoadvance/specterext/swan/client.py
+++ b/src/cryptoadvance/specterext/swan/client.py
@@ -137,13 +137,25 @@ class SwanClient:
             ).decode()
             auth_header["Authorization"] = f"Basic {auth_hash}"
 
-        response = requests.post(
-            f"{self.api_url}/oidc/token",
-            data=payload,
-            headers=auth_header,
-            timeout=30,
-        )
-        resp = json.loads(response.text)
+        try:
+            response = requests.post(
+                f"{self.api_url}/oidc/token",
+                data=payload,
+                headers=auth_header,
+                timeout=30,
+            )
+        except requests.exceptions.RequestException as e:
+            logger.exception(e)
+            raise SwanApiException(
+                f"Could not reach the Swan API ({self.api_url}/oidc/token): {e}"
+            ) from e
+        try:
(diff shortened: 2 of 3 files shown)


al-munazzim (changes requested):
Thanks for adding bounded waits to these remote HTTP calls — the timeout arguments themselves are being passed as expected.

I found one blocker in the Swan path: `authenticated_request()` currently catches all exceptions and then unconditionally logs `response.status_code` / `response.text`. With the newly-added timeout, `requests.get()` / `requests.request()` can raise before `response` is assigned, so the original `requests.exceptions.Timeout` is masked by `UnboundLocalError`. That also means callers such as the Swan settings route, which catch `SwanApiException` to show the user-friendly "Error communicating with Swan API" path, will instead see an unexpected exception.

Focused reproduction on this PR head (`6e8a762c`):

```text
requests.get kwargs: {'headers': {'User-Agent': 'Specter Desktop', 'Authorization': 'Bearer token'}, 'timeout': 30}
raised: UnboundLocalError cannot access local variable 'response' where it is not associated with a value
context: Timeout simulated timeout
```

Please handle `requests.exceptions.RequestException`/`Timeout` explicitly (or initialize/check `response` before logging it) and raise `SwanApiException` with a useful message so the existing UI error handling continues to work. A focused regression test for timeout behavior in `SwanClient.authenticated_request()` would also be helpful.

Verification run locally:

```text
python3 -m py_compile src/cryptoadvance/specterext/swan/client.py src/cryptoadvance/specter/services/extension_gen.py
# OK
```

CI at review time: black and extension-smoketest were green; test, cypress, and specterd smoke were still in progress.

k9ert:
Good catch, thanks — fixed in 92a0ef3 (branch rebased onto current master, so the old hashes are gone).

`authenticated_request()`:
- `requests.exceptions.RequestException` (timeout, connection error) is now caught explicitly and translated into `SwanApiException`, with the original kept as `__cause__`. The request context (endpoint/method/payload) is logged instead of the not-yet-assigned `response`.
- status-code check and `response.json()` moved out of the `try` — no more blanket `except Exception`. Invalid json also becomes a `SwanApiException`.
- an unsupported `method` now raises instead of falling through to an unbound `response`.

Also fixed the same class of problem one level up: `_get_access_token()` does the `requests.post` to `/oidc/token` *before* `authenticated_request()`'s try block, so a timeout there leaked a raw `requests.exceptions.Timeout` past the `except SwanApiException` handlers as well.

So the settings route / oauth callback keep showing the "Error communicating with Swan API" path.

Regression tests in `tests/test_specterext_swan_client.py`: GET timeout, POST (`requests.request`) connect-timeout, token-endpoint timeout, and non-200 status. They use the `app_no_node` fixture, since these pure-client tests don't need the regtest bitcoind that `app` pulls in.

al-munazzim (approved):
Re-reviewed 92a0ef3e61800b5352e7309fd21eeceed7cfd79a. The timeout path no longer masks the original request failure with UnboundLocalError: request exceptions from both authenticated_request() and the /oidc/token request are caught explicitly, chained into SwanApiException, and the status/json handling remains outside the network exception path. Unsupported methods now fail explicitly as well.

Verification run locally on this head:

```text
python3 -m py_compile src/cryptoadvance/specterext/swan/client.py src/cryptoadvance/specter/services/extension_gen.py tests/test_specterext_swan_client.py
# OK

Focused Flask/app-context smoke with mocked requests.get Timeout and token-endpoint requests.post Timeout
# both raised SwanApiException with the original Timeout preserved as __cause__
```

The full pytest target could not be run in this disposable environment without installing the project test stack because tests/conftest.py imports app dependencies not present globally, but the focused reproduction for the blocker is fixed. Looks good to me.

k9ert (approved):

k9ert:
Thanks for your PR!

PR #2688, opened 2026-08-08
fix: secure Docker bind default

k9ert:
## Summary

- stop the production image from binding to all interfaces by default
- honor the existing HOST configuration when --host is omitted while keeping explicit CLI precedence
- warn when an explicitly external normal server bind runs without authentication
- validate malformed configured hosts and preserve the warning in the file log

## Usage

External Docker access now requires an explicit opt-in:

    docker run -e HOST=0.0.0.0 -p 25441:25441 ghcr.io/cryptoadvance/specter-desktop

If authentication is disabled, Specter emits a prominent exposure warning.

## Verification

- 16 focused CLI server tests pass
- Black check passes
- git diff --check passes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,4 +64,4 @@ COPY --from=builder /usr/local/bin /usr/local/bin
 # Expose ports
 EXPOSE 25441 25442 25443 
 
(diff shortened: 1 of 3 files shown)


k9ert:
This is not worth fixing as an exposure via 0.0.0.0 on the Dockerfile does not expose it on the Docker host.

PR #2698, opened 2026-08-12
Fix Coldcard USB warning flow

### LINKED ISSUE #2624: Coldcard signing issues with a multisig transaction.
**Describe the bug**
When signing in a 2 of 3 multisig with a Coldcard device, Specter shows several popup windows incorrectly.

**To Reproduce**
1. Create a 2 of 3 multisig transaction that includes a coldcard device.
2. Under the Transaction Details with Sign Transaction window, choose the Coldcard device.
3. Choose sign via USB (first time).
If I start with the Coldcard device, I get the question "Do you really want to use your Coldcard over USB?".
Image

If I start with the Trezor device first and second with the Coldcard device I get both the question and the USB choice.
Image

4. I click "Yes" on the question.
Specter goes back to the Transaction Details with Sign Transaction window with the device options.
5. Choose the Coldcard device again in the Device list, then you get the following;
Image

6. Choose sign via USB (second time), then you get the following:
Image

7. Choose sign via USB (third time),
8. Plug in the Coldcard device, input pin code, choose ready to sign.
9. The transaction comes through to the Coldcard device, I can sign and thus the signing is now completed correctly.

**Expected behavior**
Signing should work correctly like it works with Trezor and Ledger. I should get only one popup asking to connect the Coldcard device, then be able to choose the Coldcard device and finish the signing.

Preferably the question "Do you really want to use your Coldcard over USB?" is removed. Or have the option in Specter to disable this question.
The question is at least one of the issues messing up the workflow.

I don't think the question adds to the overall security, including for new users, if it makes the workflow unreliable.
Also, I imagine people using Specter in a multisig setup using Coldcard as one of the signing devices understand the risk of using USB.

My node is Bitcoin Core 30.2 and 29.3 running locally on Linux Mint.
Specter runs as an imageApp on the same Linux Mint.
Specter Version 2.1.8

**Additional context**
I don't have these issues with Ledger nor Trezor devices
I remember having similar issues in previous versions of Specter with Coldcard, but not as bad as now, afaik.

Another issues that happens is that I get the detecting Coldcard device window popup when I am already done signing with the Coldcard. I can then click away and proceed with doing the transaction with no further issues.
Also a similar issue is that after signing with the Coldcard, and then choosing another device (i.e. Trezor), the "detecting coldcard usb" windows pops up, then I have to click away and choose the Trezor device again, and then it works as expected.

Finally when I connect the Coldcard before choosing it in the Specter Sign Transaction window, a "detecting plug in your device" window pops up but nothing happens. I then have to click away, and choose the Coldcard again. I then get the question "Do you really want to use your Coldcard over USB?" again, but the signing completes without answering the question.

With a single sig the issues are less pronounced, been a while since I did these with a Coldcard though.

al-munazzim:
## Summary
- make the Coldcard USB warning an awaited step before HWI enumeration starts
- resume the same pending enumerate/sign action after the user accepts the warning
- resolve the pending flow cleanly when the warning is cancelled or the Coinkite link is chosen

## Why
Fixes #2624. The old flow appended the Coldcard USB warning, then immediately continued into the normal HWI detection overlay. Clicking “Yes” only set `coldcardUsb = 1` and cancelled the overlay, so users had to choose Coldcard/sign-via-USB again and could get repeated or stale detecting popups.

## Verification
```bash
python3 - <<'PY'
from pathlib import Path
p=Path('src/cryptoadvance/specter/templates/includes/hwi/hwi.jinja')
text=p.read_text()
assert 'appendChild(document.getElementById("coldcard_warning"))' not in text
assert "showPageOverlay('coldcard_warning')" in text
assert 'async function confirmColdcardUsbNotice()' in text
assert 'let acceptedColdcardUsb = await confirmColdcardUsbNotice();' in text
print('static assertions passed')
PY

git diff --check
node --check /tmp/hwi_jinja_extracted.js
```

Manual hardware verification is still needed with a Coldcard multisig setup:
- Coldcard first signer
- another signer first, then Coldcard
- Coldcard plugged in before selecting USB
- Coldcard plugged in after the warning/detecting prompt


k9ert:
@cyberwiz9000 would you be able to test this?

Cyberwiz9000:
Hey, thanks for the quick work again.

I have not hardware-tested the PR yet, I would have to setup a testing environment first.

From reviewing the code, the normal transaction-signing path reported in [#2624](https://github.com/cryptoadvance/specter-desktop/issues/2624) appears to be fixed:

The Coldcard warning is now awaited before HWI detection starts.
Clicking Yes resumes the same signing attempt.
Escape, backdrop cancellation, or the Coinkite option stops cleanly without enumerating.
The warning should no longer reappear when subsequently signing with Trezor or Ledger.

However, before I test it, I have found two remaining issue, could you please have look at these?

### **1. Mixed-device flow, functional issue**

The Coldcard check in [hwi.jinja](https://github.com/cryptoadvance/specter-desktop/blob/4bcbb2b49327c7ba3cbd6d95f54bf34234696615/src/cryptoadvance/specter/templates/includes/hwi/hwi.jinja#L104-L112) relies on:
deviceTypes == 'coldcard'
Transaction signing works because ['coldcard'] is loosely coerced to "coldcard". A mixed list such as ['coldcard', 'trezor'] does not match.
In a mixed-device flow such as address verification, the user can subsequently select Coldcard. [selectDevice()](https://github.com/cryptoadvance/specter-desktop/blob/4bcbb2b49327c7ba3cbd6d95f54bf34234696615/src/cryptoadvance/specter/templates/includes/hwi/components/select_device.jinja#L54-L57) displays the progress overlay before rescanning the selected Coldcard.
The warning can then be appended behind the progress overlay and remain invisible because [showPageOverlay()](https://github.com/cryptoadvance/specter-desktop/blob/4bcbb2b49327c7ba3cbd6d95f54bf34234696615/src/cryptoadvance/specter/templates/includes/overlay/overlay.html#L21-L47) displays only the first child.
If the operation is cancelled, enumerate() returns null, after which rescanDevices.length can throw a TypeError.

**Proposed fix: 3 small changes, with Illustrative snippets showing the intention:**

**Make the initial Coldcard-only check explicit**

```js

const coldcardOnly =
    deviceTypes === 'coldcard' ||
    (
        Array.isArray(deviceTypes) &&
        deviceTypes.length === 1 &&
        deviceTypes[0] === 'coldcard'
    );

if (coldcardOnly && coldcardUsb === 0) {
    const accepted = await confirmColdcardUsbNotice();
    if (!accepted) return null;
}

```

This handles both 'coldcard' and ['coldcard'] without relying on loose coercion.

**Handle Coldcard selection in mixed-device flows**

After the existing hidePageOverlay() call, request consent before unlocking the selected device or showing the progress overlay:

```js

if (
    selectedDevice.type === 'coldcard' &&
    coldcardUsb === 0 &&
    !(await confirmColdcardUsbNotice())
) {
    return null;
}

// Existing unlock logic follows.
showHWIProgress(/* existing arguments */);

```

This ensures the warning is displayed before the progress overlay when Coldcard is selected from a mixed list.

**Guard the rescan result against null**

```js

const rescanDevices = await enumerate(
    selectedDevice.type,
    selectedDevice.path,
    passphrase
);

if (!rescanDevices) {
    if (overlayIsActive()) hidePageOverlay();
    return null;
}

if (rescanDevices.length === 1) {
    selectedDevice = rescanDevices[0];
}

```

This preserves the existing rescan behaviour while preventing a cancelled or failed enumeration from causing a null.length error.

Together, these three limited changes prevent the hidden stacked warning and the subsequent TypeError without requiring a larger restructuring of the HWI flow.

### **2. Concurrent invocation, smaller robustness issue**

The PR stores one global coldcardNoticeResolver. Two rapid USB actions can overwrite the first resolver, leaving the first operation pending indefinitely.

**Proposed fix: Reject a duplicate confirmation while one is already pending**

**Illustrative snippet showing the intention:**

```js

async function confirmColdcardUsbNotice() {
    if (coldcardUsb === 1) return true;

    // Do not overwrite the resolver for the warning already shown.
    if (coldcardNoticeResolver !== null) return false;

    coldcardNoticeAccepted = false;

    return await new Promise(resolve => {
        coldcardNoticeResolver = resolve;
        showPageOverlay('coldcard_warning');
    });
}

```
The duplicate action then exits through the existing false/null path, while the original operation remains connected to the visible warning.

### **Final remarks and questions**

Could you address these two points before I perform the hardware test? In the meantime, I’ll prepare an isolated test environment.

Once the changes are ready, could you provide a Linux x86_64 test build for the exact updated PR commit, preferably specterd or an AppImage, together with the commit hash and SHA-256 checksum?

Alternatively, please let me know the recommended way to build or obtain it.

_Code review and proposals were assisted by ChatGPT Codex 5.6 Sol and Maple Kimi K2.6_

Cyberwiz9000:
Would this PR (#2698) be suitable for inclusion in the next v2.2.0 prerelease? Either in its current form or after addressing the two additional points I raised? That would give me an official build to use for hardware testing.

al-munazzim:
Thanks for the careful review — both points made sense.

I pushed an update in `06b15f3de83cc03143b319d525d78857275df309` that addresses them:

- The initial Coldcard-only check is now explicit for both `'coldcard'` and `['coldcard']`, without relying on loose array/string coercion.
- Selecting a Coldcard from a mixed-device list now asks for the USB warning immediately after the selector overlay is closed and before the progress overlay/rescan path starts, so the warning should not get stacked invisibly behind the progress overlay.
- The selected-device rescan now guards `null` before checking `.length`, so cancellation/failed enumeration exits cleanly.
- `confirmColdcardUsbNotice()` now rejects a duplicate invocation while a warning is already pending, instead of overwriting the existing resolver.

Verification:

- Local `git diff --check`: passed.
- Local JS syntax smoke check of the modified template `<script>` blocks with `node --check`: passed.
- GitHub CI for `06b15f3de83cc03143b319d525d78857275df309`: `black`, `Build & smoke-test specterd (Linux)`, `test`, `cypress`, and `extension-smoketest` are all passing.

On the test build / prerelease question: I think the safest path for hardware testing is an official v2.2.0 prerelease that includes this PR, since that gives you a reproducible artifact from the project release workflow. I cannot decide prerelease inclusion myself, but from the current state this PR looks suitable for inclusion after maintainer review. The PR CI does smoke-build `specterd` on Linux, but that workflow does not currently upload a downloadable binary artifact. If you want to build the exact current commit locally, use commit `06b15f3de83cc03143b319d525d78857275df309` with the same Python 3.10/PyInstaller path used by CI:

```bash
git fetch https://github.com/al-munazzim/specter-desktop.git fix-coldcard-usb-warning-flow
git checkout 06b15f3de83cc03143b319d525d78857275df309
python3.10 -m venv .buildenv
source .buildenv/bin/activate
pip install -r requirements.txt --require-hashes
pip install -e ".[test]"
pip install -r pyinstaller/requirements.txt
echo "0.0.0-test" > pyinstaller/version.txt
cd pyinstaller
pyinstaller specterd.spec
./dist/specterd --help
```

If maintainers prefer, I can also add a temporary artifact upload to the PR build workflow or provide a dedicated test-build workflow in a follow-up commit.

PR #2702, opened 2026-08-21
Specter-DIY: import xpubs without the 3s timeout, one confirmation per key set, wrong-network handling

schnuartz-ai:
## Background

Newer Specter-DIY firmware (see cryptoadvance/specter-diy#381) asks the
user to confirm every `xpub` export on the trusted display — a deliberate
privacy property (any process that can reach the serial port could
otherwise enumerate wallet keys silently).

Problems importing such a device into Specter:

1. **Timeout.** `SpecterClient.get_pubkey_at_path()` passed
   `timeout=self.TIMEOUT` (3 s). A human confirming on-device easily
   needs longer, so the import failed with `DeviceBusyError: Timeout` and
   a closed serial port.
2. **One confirmation per key, some for the wrong network.**
   `new_device_keys.jinja` ("Get via USB") fetches one xpub per table
   row, each its own request / serial connection / on-device prompt — 4
   taps for a single-sig + multisig account. `_extract_xpubs_from_client`
   does the same for 8 (4 mainnet + 4 testnet), prompting for a network
   the device isn't even on.

## What this PR does

- **`get_pubkey_at_path()` passes no timeout** — it waits indefinitely,
  like `sign_tx()` already does. `get_master_fingerprint()` keeps its
  short bounded timeout.

- **Batch authorization via the firmware's `xpubauth` protocol**:
  - `SpecterClient.begin_xpub_authorization(paths)` /
    `end_xpub_authorization()` wrap `xpubauth begin <scope>` /
    `xpubauth end`: one scoped request authorizes a whole set of paths
    with a single on-device confirmation, after which the individual
    `xpub` reads return without a prompt.
  - New `HWIBridge` RPC methods of the same name (paths parsed and
    canonically re-serialised at the RPC boundary), plus `hwi.js` /
    `hwi.jinja` wrappers, so the **actual "Get via USB" flow** authorizes
    all pending table rows up front, then loops as before — each
    `getXpub()` returns near-instantly. A **Cancel** on the device
    aborts the whole import instead of falling through to per-key
    prompts.
  - `_extract_xpubs_from_client()` gets the same treatment for its 8-key
    fetch, and **skips the network the device isn't on** (a rejected
    scope for one network's standard paths tells us the device is on the
    other).

- **Clear message when the device is on the wrong network.**
  `get_pubkey_at_path()` recognizes `error: network mismatch: device is
  on <network>` and raises `SpecterDIYNetworkMismatchError` (a
  `BadArgumentError` subclass). The "switch to X" network is derived from
  the requested path's coin type (Mainnet / Liquid / the shared
  testnet-ish group), so it's right even for Liquid. `extract_xpub()`
  re-raises it instead of swallowing it; `handleHWIError()` unwraps the
  JSON-RPC `"Internal error:"` prefix for just that message.

- **Graceful fallback everywhere.** Clients without
  `begin_xpub_authorization` (every non-DIY device) and old DIY firmware
  without `xpubauth` keep the existing per-request flow, unchanged.

## Security hardening (separate commit)

From a code review, independent of the `xpubauth` flow:

- **`handleHWIError()` HTML-escapes the error string** before
  `showError()` (which renders via `innerHTML`). An HWI error can carry
  text straight off a connected wallet (the firmware echoes `error:
  <...>` through, HWI does the same for other devices). Static-message
  `handleHWIError()` call sites were switched to `showError()` so it only
  ever gets a caught error.
- **`_network_mismatch_error()` never reflects the device string** - only
  a whitelisted label ("an unknown network" otherwise). The hint
  requires a hardened purpose/coin type, and for an unrecognized coin
  type it says "switch to the network that derivation is for" rather than
  guessing from `client.chain`.
- **`extract_xpub()` returns `None` only on cancellation** - the
  network-mismatch error and every real failure now propagate to the UI
  instead of being logged and swallowed to an ambiguous `None`.
- The USB transport read loop sleeps 10ms on an empty read instead of
  spinning during the confirmation.

## Test plan

- [x] `black --check` (v22.3.0) on the changed files
- [x] `pytest tests/test_specter_diy_hwi.py tests/test_hwi_rpc.py` — 20
  passed, 2 skipped (timeout contract; `xpubauth` client wrapper
  authorize / old-firmware / rejected-scope / cancel / empty; RPC-level
  path normalization; the network-skip logic; `SpecterDIYNetworkMismatchError`
  translation for mainnet / testnet / liquid; cancellation propagates and
  the client is closed).
- [x] Real Specter-DIY hardware, through the actual "Get via USB" flow:
  one ~4 s `xpubauth begin` confirmation, four fast unprompted `xpub`
  reads, `xpubauth end`; device on the wrong network → refusal screen +
  a "switch the device to X" toast in Specter.

## Companion

Pairs with cryptoadvance/specter-diy#381 (the on-device confirmation, the
`xpubauth` scope protocol, and the individual-request network guard).

### DIFF
--- a/src/cryptoadvance/specter/devices/hwi/specter_diy.py
+++ b/src/cryptoadvance/specter/devices/hwi/specter_diy.py
@@ -23,6 +23,44 @@ from binascii import a2b_base64, b2a_base64
 py_enumerate = enumerate
 logger = logging.getLogger(__name__)
 
+# Prefix of the machine-parseable error the firmware sends back (as
+# "error: <this><network>") when it refuses a plain `xpub <path>` request
+# because the path's implied coin type doesn't match the network currently
+# active on the device. Keep in sync with specter-diy's
+# apps/xpubs/xpubs.py.
+_NETWORK_MISMATCH_PREFIX = "network mismatch: device is on "
+
+# Human-readable labels for the network identifiers the firmware reports,
+# matching the names shown on the device's own screen (embit's
+# NETWORKS[...]["name"]).
+_NETWORK_DISPLAY_NAMES = {
+    "main": "Mainnet",
+    "test": "Testnet",
+    "regtest": "Regtest",
+    "signet": "Signet",
+    "liquidv1": "Liquid",
+    "liquidtestnet": "Liquid Testnet",
+    "elementsregtest": "Liquid Regtest",
+}
+
+
+class SpecterDIYNetworkMismatchError(BadArgumentError):
+    """
+    The device refused an xpub request because the requested derivation's
+    coin type doesn't match the network currently active on the device
+    (e.g. asking for a mainnet-style key while the device is set to
+    Testnet). Raised instead of a generic BadArgumentError so callers can
+    tell the user exactly which network to switch the device to, the same
+    thing the device's own screen already told the person holding it.
+
+    :ivar device_network: the raw network identifier the device reported
+        (e.g. "test", "regtest") - not a display name.
+    """
+
+    def __init__(self, device_network: str, message: str) -> None:
+        super().__init__(message)
+        self.device_network = device_network
+
 
 class SpecterClient(HardwareWalletClient):
     """Create a client for a HID device that has already been opened.
@@ -92,7 +130,10 @@ class SpecterClient(HardwareWalletClient):
         """
         # xpub now requires on-device user confirmation and may take
         # arbitrarily long - wait indefinitely, same as sign_tx()
-        xpub = self.query("xpub %s" % bip32_path)
+        try:
+            xpub = self.query("xpub %s" % bip32_path)
+        except BadArgumentError as e:
+            raise self._network_mismatch_error(e) from e
         hd = ExtendedKey.deserialize(xpub)
         # Specter returns xpub with a prefix
         # for a network currently selected on the device
@@ -100,6 +141,36 @@ class SpecterClient(HardwareWalletClient):
             b"\x04\x88\xb2\x1e" if self.chain == Chain.MAIN else b"\x04\x35\x87\xcf"
         )
         return hd
+
+    def _network_mismatch_error(self, error: BadArgumentError) -> Exception:
+        """
+        Recognize the firmware's "network mismatch: device is on <network>"
+        response and turn it into a SpecterDIYNetworkMismatchError with a
+        message naming both networks involved. Any other BadArgumentError
+        is returned unchanged.
+        """
+        msg = str(error)
+        if not msg.startswith(_NETWORK_MISMATCH_PREFIX):
+            return error
+        device_network = msg[len(_NETWORK_MISMATCH_PREFIX) :].strip()
+        device_label = _NETWORK_DISPLAY_NAMES.get(
+            device_network, device_network.capitalize()
+        )
+        # We only know which *side* (main vs. everything test-like) of the
+        # network split we asked for - same ambiguity the device itself
+        # has, since testnet/signet/regtest/liquidtestnet all share BIP44
+        # coin type 1'. Naming the specific requested network would be a
+        # guess; naming the device's actual network (verbatim from the
+        # device) is not.
+        requested_label = (
+            "Mainnet" if self.chain == Chain.MAIN else "Testnet/Signet/Regtest"
+        )
+        return SpecterDIYNetworkMismatchError(
+            device_network,
+            "This Specter-DIY is currently set to %s, so it can't share a "
+            "%s key from here. Switch the device to %s and try again."
+            % (device_label, requested_label, requested_label),
+        )
 
     def begin_xpub_authorization(self, paths) -> bool:
         """
--- a/src/cryptoadvance/specter/devices/hwi/specter_diy.py
+++ b/src/cryptoadvance/specter/devices/hwi/specter_diy.py
@@ -43,6 +43,44 @@ _NETWORK_DISPLAY_NAMES = {
     "elementsregtest": "Liquid Regtest",
 }
 
+# Standard BIP44-style purposes that carry an implied coin type (and hence
+# an implied network). Keep in sync with specter-diy's apps/xpubs/scope.py.
+_STANDARD_PURPOSES = {44, 48, 49, 84, 86, 87}
+
+# By-name switch-to hint for a requested path's coin type, mirroring
+# specter-diy's network_hint_for_path(). Coin type 1' is shared by
+# testnet/signet/regtest/liquidtestnet/elementsregtest, so it can only be
+# named as the group.
+_COIN_TYPE_NETWORK_HINTS = {
+    0: "Mainnet",
+    1: "Testnet, Signet or Regtest",
+    1776: "Liquid",
+}
+
+
+def _requested_network_label(bip32_path):
+    """
+    The network a standard-purpose derivation path implies, by name, from
+    its coin type - or None if the path isn't standard-purpose (and so
+    carries no network meaning). Same logic the firmware applies.
+    """
+    parts = [p for p in str(bip32_path).strip().split("/") if p and p.lower() != "m"]
+    if len(parts) < 2:
+        return None
+
+    def _index(component):
+        component = component.strip().rstrip("h'")
+        return int(component)
+
+    try:
+        purpose = _index(parts[0])
+        coin_type = _index(parts[1])
+    except ValueError:
+        return None
+    if purpose not in _STANDARD_PURPOSES:
+        return None
+    return _COIN_TYPE_NETWORK_HINTS.get(coin_type)
+
 
 class SpecterDIYNetworkMismatchError(BadArgumentError):
     """
@@ -133,7 +171,7 @@ class SpecterClient(HardwareWalletClient):
         try:
             xpub = self.query("xpub %s" % bip32_path)
         except BadArgumentError as e:
-            raise self._network_mismatch_error(e) from e
+            raise self._network_mismatch_error(e, bip32_path) from e
         hd = ExtendedKey.deserialize(xpub)
         # Specter returns xpub with a prefix
         # for a network currently selected on the device
@@ -142,7 +180,9 @@ class SpecterClient(HardwareWalletClient):
         )
         return hd
 
-    def _network_mismatch_error(self, error: BadArgumentError) -> Exception:
+    def _network_mismatch_error(
+        self, error: BadArgumentError, bip32_path: str
+    ) -> Exception:
         """
         Recognize the firmware's "network mismatch: device is on <network>"
         response and turn it into a SpecterDIYNetworkMismatchError with a
@@ -156,15 +196,16 @@ class SpecterClient(HardwareWalletClient):
         device_label = _NETWORK_DISPLAY_NAMES.get(
             device_network, device_network.capitalize()
         )
-        # We only know which *side* (main vs. everything test-like) of the
-        # network split we asked for - same ambiguity the device itself
-        # has, since testnet/signet/regtest/liquidtestnet all share BIP44
-        # coin type 1'. Naming the specific requested network would be a
-        # guess; naming the device's actual network (verbatim from the
-        # device) is not.
-        requested_label = (
-            "Mainnet" if self.chain == Chain.MAIN else "Testnet, Signet or Regtest"
-        )
+        # Prefer the network implied by the path we actually asked for
+        # (its coin type) - that's unambiguous for mainnet (0') and Liquid
+        # (1776'), and for the shared 1' coin type it's the same group the
+        # device names on its own screen. Fall back to the client's chain
+        # only if the path isn't standard-purpose.
+        requested_label = _requested_network_label(bip32_path)
+        if requested_label is None:
+            requested_label = (
+                "Mainnet" if self.chain == Chain.MAIN else "Testnet, Signet or Regtest"
+            )
         return SpecterDIYNetworkMismatchError(
             device_network,
             "This Specter-DIY is currently set to %s, so it can't share a "
--- a/src/cryptoadvance/specter/devices/hwi/specter_diy.py
+++ b/src/cryptoadvance/specter/devices/hwi/specter_diy.py
@@ -62,19 +62,23 @@ def _requested_network_label(bip32_path):
     """
     The network a standard-purpose derivation path implies, by name, from
     its coin type - or None if the path isn't standard-purpose (and so
-    carries no network meaning). Same logic the firmware applies.
+    carries no network meaning). Matches the firmware rule: the purpose
+    and coin type must both be *hardened* for the path to carry network
+    semantics.
     """
     parts = [p for p in str(bip32_path).strip().split("/") if p and p.lower() != "m"]
     if len(parts) < 2:
         return None
 
-    def _index(component):
-        component = component.strip().rstrip("h'")
-        return int(component)
+    def _hardened_index(component):
+        component = component.strip()
+        if component[-1:] not in ("h", "H", "'"):
+            raise ValueError("not hardened")
+        return int(component[:-1])
 
     try:
-        purpose = _index(parts[0])
-        coin_type = _index(parts[1])
+        purpose = _hardened_index(parts[0])
+        coin_type = _hardened_index(parts[1])
     except ValueError:
         return None
     if purpose not in _STANDARD_PURPOSES:
@@ -193,25 +197,32 @@ class SpecterClient(HardwareWalletClient):
         if not msg.startswith(_NETWORK_MISMATCH_PREFIX):
             return error
         device_network = msg[len(_NETWORK_MISMATCH_PREFIX) :].strip()
-        device_label = _NETWORK_DISPLAY_NAMES.get(
-            device_network, device_network.capitalize()
+        # The device controls this string; never put it in the message
+        # verbatim (it's rendered, not escaped, in the UI). Only ever show
+        # a known, whitelisted label.
+        device_label = (
+            _NETWORK_DISPLAY_NAMES.get(device_network) or "an unknown network"
         )
-        # Prefer the network implied by the path we actually asked for
-        # (its coin type) - that's unambiguous for mainnet (0') and Liquid
-        # (1776'), and for the shared 1' coin type it's the same group the
-        # device names on its own screen. Fall back to the client's chain
-        # only if the path isn't standard-purpose.
+        # Name the network to switch to from the requested path's own coin
+        # type - unambiguous for mainnet (0') and Liquid (1776'), the
+        # shared group for 1'. If the coin type has no name we recognize,
+        # don't guess from the client's chain (that can produce a "switch
+        # to Mainnet" while the device already is on Mainnet) - say what
+        # the device's own screen says instead.
         requested_label = _requested_network_label(bip32_path)
-        if requested_label is None:
-            requested_label = (
-                "Mainnet" if self.chain == Chain.MAIN else "Testnet, Signet or Regtest"
+        if requested_label is not None:
+            message = (
+                "This Specter-DIY is currently set to %s, so it can't share a "
+                "%s key from here. Switch the device to %s and try again."
+                % (device_label, requested_label, requested_label)
             )
-        return SpecterDIYNetworkMismatchError(
-            device_network,
-            "This Specter-DIY is currently set to %s, so it can't share a "
-            "%s key from here. Switch the device to %s and try again."
-            % (device_label, requested_label, requested_label),
-        )
+        else:
+            message = (
+                "This Specter-DIY is currently set to %s, which doesn't match "
+                "the requested derivation. Switch the device to the network "
+                "that derivation is for and try again." % device_label
+            )
+        return SpecterDIYNetworkMismatchError(device_network, message)
 
     def begin_xpub_authorization(self, paths) -> bool:
         """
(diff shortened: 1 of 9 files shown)


k9ert:
Is this tested? @schnuartz-ai @Schnuartz

al-munazzim:
CI has now exercised the added regression coverage: the PR's current check run is green for black, the main test job, cypress, extension-smoketest, and the Linux specterd build/smoke test. The new tests are mocked/unit-level coverage for the timeout contract: xpub requests call the transport without a timeout, while fingerprint requests still use the bounded `SpecterClient.TIMEOUT` path.

What is still not covered here is a manual end-to-end run with an actual Specter DIY device requiring xpub confirmation. So from the automated side this is tested; hardware verification would still be useful before treating it as fully device-verified.

schnuartz-ai:
### Hardware test — network-mismatch message end-to-end

Tested against a physical Specter-DIY running cryptoadvance/specter-diy#381,
through the real "Get via USB" flow at `/devices/new_device_keys/specter/`
(plus directly via `HWIBridge.jsonrpc()`).

**Device on Testnet, Specter on Mainnet** → "Get via USB" requests the
mainnet account keys, the device refuses the first one, and Specter shows
exactly:

> This Specter-DIY is currently set to Testnet, so it can't share a
> Mainnet key from here. Switch the device to Mainnet and try again.

No `Internal error: …` wrapper, no doubled period, and no generic
"Failed to retrieve device data. Please try again." stacked on top — just
the one actionable sentence.

- The "switch to X" network comes from the **requested path's coin
  type**, so `m/84h/1776h/0h` gives *"…can't share a **Liquid** key…
  Switch the device to **Liquid**…"* — which the bitcoin-only `Chain`
  enum can't represent.
- Driven through `HWIBridge.jsonrpc()` with the device on Mainnet /
  Regtest / Liquid: the matching-coin-type request returns the xpub, the
  two mismatched ones raise `SpecterDIYNetworkMismatchError` naming the
  right target network.

**Happy path (device on Mainnet)** — "Get via USB" for a full account →
one on-device *"Share multiple XPUBs?"* confirmation (Network: Mainnet,
the four standard paths listed), then four fast unprompted `xpub` reads.
Previously four separate confirmations.

`pytest tests/test_specter_diy_hwi.py tests/test_hwi_rpc.py` green;
`black --check` (22.3.0) clean on the changed files.

---

**The message in Specter Desktop** (device on Testnet, "Get via USB"):

Specter Desktop toast: This Specter-DIY is currently set to Testnet, so it can't share a Mainnet key from here. Switch the device to Mainnet and try again.

**Same moment, device side** — the DIY shows the refusal screen while Specter is mid-request:

device refusal screen + Specter Desktop "Extracting public key…"

**Happy path** (device on Mainnet) — one confirmation covers the whole standard set:

Share multiple XPUBs? confirmation on device, Specter Desktop "Requesting access to public keys…"

schnuartz-ai:
### Review addressed — `843ab831`

**🔴 1. Cancel on the batch authorization now aborts the export.**
`beginXpubAuthorization()` was folding *every* failure into the same
`false` that also means "unsupported → use the per-key fallback", so an
explicit Cancel on the "Share multiple XPUBs?" screen fell through to a
stream of individual prompts. Fixed: it re-raises (after showing the
toast); the "Get via USB" handler catches and `return`s. `false` is now
returned *only* for the non-throwing unsupported / rejected-scope cases.
`_extract_xpubs_from_client` already propagated `ActionCanceledError`
within its single RPC — that path was fine; only the multi-RPC
`new_device_keys.jinja` flow had the gap.

**🔴 2. CI on the current head.**
The branch rebases cleanly on `master` (verified locally: `black
--check`, `pytest tests/test_hwi_rpc.py tests/test_specter_diy_hwi.py` —
green). I can't push `master`'s newer `.github/workflows/` changes from
the fork without `workflow` scope, and `update-branch` via the API hits
the same wall, so the branch still shows behind — a maintainer "Update
branch" (or approving the workflow run) gets the full `test` / `black` /
`cypress` / `extension-smoketest` suite onto the real head. The 21 Aug
green-CI comment is from the timeout-only version and doesn't count for
the current diff — agreed.

**🟠 3. The `_extract_xpubs` tests were a false positive.**
Exactly right — `_FakeExtKey.to_string()` returned a non-decodable
placeholder, `convert_xpub_prefix()` threw for every key, the per-key
`except` swallowed it, and the tests asserted only on *which paths* were
requested. Now: a real xpub, plus assertions that all four (or eight)
`[fpr/derivation]xpub` lines actually come back with the slip132 prefix
applied. Also added two tests for the RPC-boundary path validation below.

**🟠 4. Authorization living device-wide between RPCs.**
Documented in the `begin_xpub_authorization` docstring: `begin`/`end` are
separate RPCs, connection reopened per call, so the authorization lives
on the device on its own between them. It stays narrow — RAM only,
network-bound, exact confirmed paths, one-shot each, dropped on lock /
re-init / network change. Residual exposure is a host crash mid-batch
leaving the still-unread paths authorized until one of those events. I've
called out an atomic server-side `begin→reads→end` (`extract_xpub_batch`)
and/or a short firmware TTL as the follow-up that closes it, but left it
out of this PR to keep the surface contained, per your read that it's not
a blocker.

**🟡 5. RPC now validates all paths itself.**
`begin_xpub_authorization` parses every entry with `bip32.parse_path()`
and re-serialises it canonically *before* the list is joined into the
firmware scope string. A `;` or range expression in a single `paths[]`
entry now fails at the boundary instead of reaching the scope grammar.
Tests: `test_begin_xpub_authorization_rejects_a_non_bip32_path`,
`test_begin_xpub_authorization_normalises_paths_before_the_device_sees_them`.

**🟡 6. `handleHWIError()` change is now scoped.**
It only unwraps the `"Internal error: …."` wrapper when the message is
the Specter-DIY network-mismatch sentence (matched on its text). Ledger /
Trezor / Coldcard / etc. error formatting is untouched.

schnuartz-ai:
### Reworked: batch xpub authorization removed

Force-pushed a rebuild. On top of the original timeout fix
(`4e977d83` + `d331ed7c`) there is now a single commit, and the batch
`xpubauth` wiring is gone entirely.

**Gone:** `SpecterClient.begin_xpub_authorization()` /
`end_xpub_authorization()`, the `begin_xpub_authorization` /
`end_xpub_authorization` HWI-bridge RPCs (and their `exposed_rpc`
entries), the `beginXpubAuthorization` / `endXpubAuthorization` JS +
`hwi.jinja` wrappers, the "authorize the whole set up front" step in
`new_device_keys.jinja`, and the scope-probe / skip-the-other-network
logic in `_extract_xpubs_from_client()` — that method is byte-for-byte
back to its pre-PR form (fetches each standard key individually, both
networks).

This resolves the review blockers that were specific to the batch flow:
- the **cancel-doesn't-abort** bug — there is no batch confirmation to
  cancel any more; each `getXpub()` is confirmed on its own and a
  cancelled one stops the import via the existing null/return path.
- the **RPC-boundary path validation** and the **device-wide
  authorization lifetime** concerns — both were about the
  `begin_xpub_authorization` RPC, which no longer exists.
- the **`_extract_xpubs` false-positive test** — those tests are removed
  with the code they covered; a new test asserts
  `_extract_xpubs_from_client()` makes the 8 individual
  `get_pubkey_at_path` calls with no batch step and returns all 8
  converted keys.

**Kept:** the 3s-timeout removal on `get_pubkey_at_path()`
(`get_master_fingerprint` keeps its short timeout); the
`SpecterDIYNetworkMismatchError` translation (network taken from the
requested path's coin type, Liquid included); `extract_xpub()`
re-raising it; the scoped `handleHWIError()` unwrap for that one message.

`black --check` clean; `pytest tests/test_specter_diy_hwi.py
tests/test_hwi_rpc.py` → 10 passed, 2 skipped. The fork PR's `test` /
`black` / `cypress` / `extension-smoketest` checks still need a
maintainer "Approve and run" to execute on the head.

Schnuartz (approved):
Tested on real Hardware (works very well if you tell AI to make some specific Tests. So AI Can connect to the board and send tests. 

Also did some with Specter Desktop fork.

Also reviewed code

schnuartz-ai:
### Review addressed — `7f527125`

**🔴 1. DOM-XSS via `device_network` — fixed.**
`_network_mismatch_error()` no longer puts the device-controlled string
anywhere in the message. `device_label` is now
`_NETWORK_DISPLAY_NAMES.get(device_network) or "an unknown network"` —
whitelist-only. The whole message is now composed exclusively of fixed
literals + whitelisted labels + the path-derived `requested_label`
(itself only ever a value from `_COIN_TYPE_NETWORK_HINTS` or a fixed
`Chain` string), so nothing device-controlled reaches `showError()` /
`innerHTML`. The raw value is still kept on
`SpecterDIYNetworkMismatchError.device_network` for programmatic callers,
just never rendered. Test:
`test_network_mismatch_message_never_echoes_the_device_string` (feeds
``, asserts it's absent from `str(err)` and that
`"an unknown network"` is used).

**🔴 2. Silently-swallowed xpub errors — fixed.**
`HWIBridge.extract_xpub()` now returns `None` **only** for
`ActionCanceledError` (the caller's "user cancelled, stop quietly"
signal). Every other exception — the network mismatch, transport /
deserialize / prefix-conversion errors — propagates as the JSON-RPC
error, so `handleHWIError()` shows it once. `getXpub() == null` in the
"Get via USB" loop now unambiguously means "cancelled", and the quiet
`return` there is correct — no swallowed failures, no double toast.
Tests: `test_extract_xpub_returns_none_when_the_user_cancels`,
`…_propagates_a_network_mismatch`, `…_propagates_an_unexpected_error`.

**🟠 3. Unbounded timeout under the global HWI lock — acknowledged.**
`get_pubkey_at_path()` waits without a timeout, matching `sign_tx`,
`sign_message`, `display_address` and every other interactive
Specter-DIY call. A device that ACKs then stalls holds `hwilock` — but
that's true of all of them, and bounding it properly is a transport-wide
change, not this PR. Noted in a code comment; happy to file a follow-up
for a generous human timeout across the interactive ops.

**🟡 4. Duplicated network semantics — tightened.**
`_requested_network_label()` now requires the purpose *and* coin type to
be hardened (`m/84h/1h/…`, not `m/84/1/…`), matching the firmware's
`purpose < HARDENED → not standard-purpose` rule. A non-hardened path
falls back to the client `chain` instead of claiming a network from an
under-specified path. Test:
`test_get_pubkey_at_path_ignores_a_non_hardened_purpose_for_the_hint`.
Sharing the actual BIP32 parser across the process boundary would be
nicer still but is a bigger refactor.

**CI.** `pytest tests/test_specter_diy_hwi.py tests/test_hwi_rpc.py` →
15 passed, 2 skipped; `black --check src/` clean. The fork PR's
`test` / `black` / `cypress` / `extension-smoketest` still need a
maintainer "Approve and run" on the current head (and the branch is
`BEHIND master` — the push can't touch `master`'s newer
`.github/workflows/` from a fork without `workflow` scope, so a
maintainer "Update branch" is needed for the full run).

schnuartz-ai:
### Re-review addressed — `604f1aa8`

**🔴 2. Generic HWI-error XSS path — closed two ways.**
- `showNotification()` (the sink for `showError` / `handleHWIError`) now
  sets `el.textContent`, not `el.innerHTML`. No caller passes markup
  (checked all ~98); some pass hardware-wallet / HWI error text. This
  closes the class for *every* device and every RPC path
  (`sign_tx`/`display_address`/…), not just xpub.
- Belt and braces on the server side: `extract_xpub()` returns `None`
  only for `ActionCanceledError`; the network-mismatch error (message
  built entirely from a fixed whitelist) propagates; **every other
  failure is logged in full server-side but the UI is handed a fixed
  string** — the raw `error: <...>` the firmware echoes back never
  reaches the JSON-RPC message. New tests:
  `test_extract_xpub_never_forwards_raw_device_error_text_to_the_ui`
  (feeds ``), `…_surfaces_the_network_mismatch_message`,
  `…_returns_none_when_the_user_cancels`.

**🟠 3. Network-mismatch hint for unusual coin types — fixed.**
No more `client.chain` guess. `_requested_network_label()` requires a
hardened purpose *and* coin type (matching the firmware). When the coin
type has no known name (`m/84h/2h/0h`, …) the message is now *"…currently
set to Mainnet, which doesn't match the requested derivation. Switch the
device to the network that derivation is for…"* — same as the device's
own screen, never a contradictory "set to Mainnet → switch to Mainnet".
Tests: `test_get_pubkey_at_path_does_not_guess_a_target_for_an_unnamed_coin_type`
(the `m/84h/2h/0h` case), `…_ignores_a_non_hardened_path_for_the_hint`.

**🟠 4. Legacy "Manual configuration" flow → 4 wrong-network dialogs — fixed.**
`_extract_xpubs_from_client()` now stops requesting a coin-type group as
soon as the device refuses one path with a network mismatch. A
Specter-DIY import via that flow is now **one** refusal dialog + the
matching group, not four refusals. Rewritten from 8 copy-pasted
try/except blocks into a data-driven loop; every non-DIY client fetches
both groups in the same order as before. Tests:
`test_extract_xpubs_from_client_stops_a_group_on_the_first_wrong_network`,
and the existing `…_fetches_each_key_individually` still asserts the full
8-path behavior for a client that never raises a mismatch.

**🟡 5. Timeout / HWI lock — acknowledged, follow-up.**
`get_pubkey_at_path()` waits without a timeout, consistent with
`sign_tx` / `sign_message` / `display_address`. A device that ACKs then
stalls holds `hwilock` and the transport busy-polls — true for all of
them; bounding it belongs in a shared "generous interactive timeout +
non-busy transport wait" change, not this PR. Noted in a code comment; I
can open a follow-up issue.

**🔴 6 / 🟠 process.** Local: `pytest tests/test_specter_diy_hwi.py
tests/test_hwi_rpc.py` → 17 passed, 2 skipped; `black --check src/`
clean. The fork PR still needs a maintainer "Approve and run" for the
full `test` / `black` / `cypress` / `extension-smoketest` on the current
head, and a "Update branch" (the push can't carry `master`'s newer
`.github/workflows/` from a fork without `workflow` scope). Re-review on
the final SHA very welcome — this commit reworked the whole error path.

PR title updated to reflect that it's the timeout + wrong-network
handling, not "one confirmation for the import".

schnuartz-ai:
### Re-review addressed — `86c876eb`

**🔴 1. `showNotification()` `<br>` regression — fixed, `textContent` kept.**
Good catch — I missed `tx-table.html`. Audited every `showError` /
`showNotification` callsite (parsed the argument of each, checked for
`<br>/<b>/<a>/<span>/…`): exactly one carried markup, `tx-table.html`'s
"Failed to fetch wallet data! `<br>` …". That now passes `\n`, and the
message slot in `message-box.html` gets `white-space: pre-line` so a
`\n` renders as a line break for any caller. `textContent` stays — every
device's error strings are inert now, not just Specter-DIY's.
(The server-rendered `<message-box>{{ … | safe }}` flash messages are
untouched — those set the light DOM directly, not via `textContent`, and
`pre-line` doesn't change HTML parsing.)

**🔴 2. Required CI on the current SHA — not something I can trigger.**
Fork PR: `test` / `black` / `cypress` / `extension-smoketest` sit at
`action_required` until a maintainer runs them, and the branch is
`BEHIND master` (the push can't carry `master`'s newer
`.github/workflows/` from a fork without `workflow` scope, so `Update
branch` is also a maintainer action). Local: `pytest
tests/test_specter_diy_hwi.py tests/test_hwi_rpc.py` → 17 passed, 2
skipped; `black --check src/` clean. The APPROVED review predates this
whole error-path rework — a re-review on the final SHA is very welcome.

**🟠 3. Busy-poll under the HWI lock — partly fixed here.**
The USB transport's `read_until` now `time.sleep(0.01)` when a
non-blocking `read(1)` returns nothing, instead of spinning for the
whole confirmation. Verified on hardware: fingerprint + a confirmed
`xpub` still round-trip fine. The remaining "wall-clock-infinite wait
holds `hwilock`" applies equally to `sign_tx` / `sign_message` /
`display_address`; I'll open a follow-up for a shared generous
interactive timeout rather than bounding just this one call.

**🟡 JSON-RPC error code instead of string-matching.**
Agreed it'd be cleaner, and it's a Specter-Desktop-only change (the
firmware string stays as-is for old-client compat). Not doing it in this
PR — the string match is contained to one `.indexOf()` and the
user-facing message is unaffected. Happy to follow up with a
`-32010 / SPECTER_DIY_NETWORK_MISMATCH` code the frontend keys off.

schnuartz-ai:
### Re-review addressed — `7546a371`

**🔴 1. HTML-entity / `textContent` regression — fixed by reverting `textContent`.**
You're right: `HTMLSafeBabel` HTML-escapes every translated string on
purpose (for `innerHTML` rendering), and I verified `{{ _("Can't…") |
tojson }}` doesn't help — the `Markup` value already has `&#39;` baked
in, so `json.dumps` just carries it through.

So `showNotification()` / `message-box.html` / `tx-table.html` are back
to their original state — `innerHTML`, `<br>` works, entities decode, no
regression anywhere.

The XSS concern is now handled at the **one boundary where untrusted
device text actually enters the notification system**: `handleHWIError()`
HTML-escapes its error string before `showError()`. The firmware echoes
`error: <…>` straight through, and HWI does the same for other wallets;
`handleHWIError` is the sink for all of that (getXpub / signTx /
displayAddress / registerMultisig / pin). The three
`handleHWIError("<static string>")` call sites were switched to
`showError()` directly, so `handleHWIError` now only ever receives a
caught error / RPC response — no risk of double-escaping an
already-safe Jinja string.

Verified with node: `error: ` →
`error: &lt;img src=x onerror=alert(1)&gt;` → rendered as inert text; the
Specter-DIY message's `can't` round-trips through
`&#39;` → `'` correctly.

Server side stays as a second layer: `extract_xpub()` hands the UI a
fixed generic string for any unexpected error (test:
`test_extract_xpub_never_forwards_raw_device_error_text_to_the_ui`),
never the raw `error: <…>`.

*Regression test:* the repo has no JS unit harness (only Cypress, needs a
running app), so this is a documented pure-string transform + the
server-side test. Happy to add a Cypress spec if you'd prefer one.

**🔴 2. CI on the final SHA.** Still needs a maintainer "Approve and run"
(fork PR) and "Update branch" (`behind master`; the push can't carry
`master`'s `.github/workflows/` from a fork without `workflow` scope).
Local: `pytest tests/test_specter_diy_hwi.py tests/test_hwi_rpc.py` → 17
passed, 2 skipped; `black --check src/` clean. A re-review on the final
SHA is very welcome — this is the third rework of the error path.

**Not in this PR (agreed):** the dedicated `SPECTER_DIY_NETWORK_MISMATCH`
JSON-RPC code and a shared generous HWI interactive timeout — both
Specter-Desktop-only follow-ups, no firmware change needed.

schnuartz-ai:
### Reverted the `xpubauth` removal — batch authorization is back

An earlier force-push had stripped the scoped/batch `xpubauth`
authorization from this PR. That was a mistake on my side and has been
undone: the branch is back to the state that has

- `xpubauth begin <scope>` / `xpubauth end` (and the `scope.py` grammar /
  `Authorization` / limits),
- one on-device confirmation covering a bounded standard key set (the
  "Share multiple XPUBs?" screen),
- single-use consumption, RAM-only state, cleared on lock / re-init /
  network change / self-exhaustion / replacement `begin`,

plus everything that was kept throughout: the per-request confirmation,
the wrong-network refusal + machine-parseable error, the 3s-timeout
removal, input bounding / depth cap.

The screenshots and older comments referencing "Share multiple XPUBs?"
apply again. Sorry for the churn.

schnuartz-ai:
### Added: security hardening on top of the restored batch flow (`c4aca370`)

The `xpubauth` batch authorization is back (see the previous comment).
On top of it, the parts of the earlier security-review round that are
**independent of the batch feature** have been re-applied:

- `handleHWIError()` HTML-escapes the error string before `showError()` —
  the one place text straight from a connected wallet enters the
  notification system (firmware echoes `error: <...>`, HWI does the same
  for other devices). Static-string `handleHWIError()` callers switched
  to `showError()`.
- `_network_mismatch_error()` never puts the device-reported network
  token in the message (whitelist only); requires a hardened
  purpose/coin type for the hint; for an unnamed coin type it says
  "switch to the network that derivation is for" instead of a
  possibly-contradictory `client.chain` guess.
- `extract_xpub()` returns `None` only on an on-device cancellation;
  other failures propagate to the UI instead of vanishing into a log.
- USB transport read loop sleeps 10ms on an empty read instead of
  busy-spinning through the confirmation.

`pytest tests/test_specter_diy_hwi.py tests/test_hwi_rpc.py` → 25 passed,
2 skipped; `black --check src/` clean. Verified on real Specter-DIY
hardware: comms unaffected, `m/84h/1h/0h` → "…Testnet, Signet or
Regtest…", `m/84h/2h/0h` (unnamed coin type) → "…which doesn't match the
requested derivation…".

Schnuartz:
Tested again on Real Hardware and it worked well

schnuartz-ai:
### On CI / merge state (`c4aca370`)

Nothing is failing — the `test` / `black` / `Test specterd build` checks
show `action_required` because this is a fork PR: they need a maintainer
to click **"Approve and run"**. The old green run was against a much
earlier state and doesn't count for the current head (agreed).

To help make that a one-click decision:

**Merge with current `master` (`9fd713f4`) is clean — zero conflicts.**
I test-merged locally. The only files this PR and `master` both touch are
`hwi.jinja` and `select_device.jinja` (from #2698's Coldcard USB flow):
- `hwi.jinja`: my change is inside `handleHWIError()`; #2698's is the
  `coldcardUsb` / `confirmColdcardUsbNotice()` / `enumerate()` block
  right after it — adjacent, non-overlapping, git merges them cleanly.
- `select_device.jinja`: my change is the one-line
  `handleHWIError(...)` → `showError(...)` at the fingerprint-mismatch
  check; #2698's is the coldcard-consent block and a `rescanDevices ==
  null` guard — different parts of the file.

On the locally merged tree:
- `black --check src/` — clean
- `pytest tests/test_specter_diy_hwi.py tests/test_hwi_rpc.py` — 25
  passed, 2 skipped

So: **"Update branch"** (or a maintainer merge) + **"Approve and run"**
is all that's needed; there's no conflict resolution to do and no local
red.

(The branch can't be updated from my side — pushing `master`'s newer
`.github/workflows/` changes from a fork needs `workflow` scope, and the
`update-branch` API hits the same wall.)

